PR #43, opened 2026-04-27
refactor(server): proper Python package with tool-calling, tests, uv setup

easel:
## Summary

- Replaces ad-hoc `scripts/server*.py` with a proper `dflash` Python package under `src/dflash/server/`
- OpenAI-compatible API with modular parsing, schemas, and tool-calling support
- pytest coverage for parsing and server endpoints
- `pyproject.toml` with `[build-system]`, `dflash-server` entry point, and `uv.lock`
- Removes old `scripts/server.py`, `scripts/server_tools.py`, `scripts/test_server.py`

**Base:** PR #48 (`fix/consumer-blackwell-auto-detect`) — RTX 5090 hardware support

## Test plan
- [ ] `uv sync` — `dflash-server` entry point installs without warnings
- [ ] `pytest dflash/tests/` passes
- [ ] `dflash-server --model <gguf>` starts and serves `/v1/chat/completions`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #47, opened 2026-04-27
feat(server): derive model name from GGUF; default port 1236, ctx 128K

easel:
## Summary

- `/v1/models` now returns the GGUF filename stem (e.g. `Qwen3.5-27B-Q4_K_M`) instead of the hardcoded `luce-dflash`
- Default port changed to `1236`
- Default `--max-ctx` changed to `131072` (128K validated on RTX 5090 with TQ3_0 KV cache)
- Startup doc updated to use `uv run`

## Test plan

- [ ] `uv run scripts/server.py` starts on port 1236
- [ ] `GET /v1/models` returns the GGUF stem as the model id
- [ ] Chat completions work at 128K context


easel:
Closing - content absorbed into PR #43 (feat/setup-results-uv) which now contains all server/uv changes as a single clean commit rebased on top of the 5090 hardware support PR (#48).

PR #48, opened 2026-04-27
fix(dflash): auto-detect GPU arch to prevent sm_120a on consumer Blackwell

easel:
## Problem

On CUDA 13.2+ with consumer Blackwell GPUs (for example, RTX 5090, SM 12.0), using an unset
`CMAKE_CUDA_ARCHITECTURES` or `native` can resolve to `sm_120a` instead of
`sm_120`, which can trigger `CUDA_ERROR_ILLEGAL_INSTRUCTION` at runtime on
consumer hardware.

## Fix (auto-detect only)

- At configure time, if `CMAKE_CUDA_ARCHITECTURES` is unset or `native`, run
  `nvidia-smi --query-gpu=compute_cap --format=csv,noheader`.
- Parse the compute capability (for example `12.0`) and set
  `CMAKE_CUDA_ARCHITECTURES` explicitly (for example `120`).
- Keep the change isolated to `dflash/CMakeLists.txt` so it can be reviewed and
  merged independently from consumer-specific workaround behavior.

## Test plan

- [ ] `cmake -B build -S dflash/` prints `dflash27b: GPU compute_cap 12.0 → CUDA_ARCHITECTURES=120` on Blackwell hardware.
- [ ] `cmake --build build` succeeds without CUDA-arch related compiler/runtime errors on a Blackwell consumer system.

### DIFF
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -3,6 +3,34 @@ set(DFLASH27B_GPU_BACKEND "cuda" CACHE STRING "GPU backend to build: cuda or hip
 set_property(CACHE DFLASH27B_GPU_BACKEND PROPERTY STRINGS cuda hip)
 string(TOLOWER "${DFLASH27B_GPU_BACKEND}" DFLASH27B_GPU_BACKEND)
 if(DFLASH27B_GPU_BACKEND STREQUAL "cuda")
+    # CUDA 12.8+ can resolve "native" on some Blackwell systems to the
+    # accelerator-specific sm_120a target. Consumer Blackwell needs the plain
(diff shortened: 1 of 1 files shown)


davide221:
@easel thanks for the contribution! Is the speed problem still present ?

easel:
> @easel thanks for the contribution! Is the speed problem still present ?

Yes. I think it's related to the workflow -- I'm putting together a small benchmark script to compare.

easel:
This may not be necessary if expectation is to always build multi-arch binary. I ran into it because claude got excited about optimizing and ended up with a slightly incompatible build.

PR #56, opened 2026-04-28
dflash: asymmetric K/V quantization (builds on#54)

dusterbloom:
## Why this PR

PR #54 merged a minimal first cut of asymmetric K/V quantization (env vars `DFLASH27B_KV_K` / `DFLASH27B_KV_V` parsed inline in `qwen35_target_graph.cpp`). This PR replaces that inline parsing with a self-contained `kv_quant` module and adds the surface needed to ship the feature end-to-end.

I started this work in parallel with #54 — when I rebased, I found #54 had landed first. After diffing both implementations I rebased mine on top of `main` and dropped the inline blocks from #54 in favour of `dflash::resolve_kv_types(...)`. The rebase compiles green and the unit tests pass.

## What #54 was missing (and what this PR adds)

- **`atoi(s) != 0` bug** — #54 guards string parsing on `if (std::atoi(s) != 0)`, but `atoi("q8")` returns 0, so `DFLASH27B_KV_K=q8` is silently ignored. This PR drops that guard and uses a real string parser.
- **No build-flag change** — #54 did not enable `GGML_CUDA_FA_ALL_QUANTS`, so any K≠V pair beyond the diagonal {F16,Q4_0,Q8_0,BF16,TQ3_0}² wouldn't have a compiled fattn kernel. This PR forces it ON in `dflash/CMakeLists.txt`, instantiating the wide K×V matrix.
- **No supported-pair validation** — unsupported pairs would silently fall through to a missing kernel. This PR checks against the fattn matrix at allocation and aborts with the full supported list on mismatch.
- **No `kv_v_type` on `TargetCache`** — #54 only stores `kv_k_type`; this PR adds the V counterpart and plumbs it through `build_full_attn_block`.
- **256-stride trigger only checks K** — fattn V-tile alignment can also need it. This PR widens the trigger to either side being TQ3.
- **No CLI surface** — this PR adds llama-cli-style `-ctk`/`--cache-type-k` and `-ctv`/`--cache-type-v` to `test_dflash` + `test_generate`, and `--ctk`/`--ctv` (with auto-TQ3 suppression) to `run.py`, `server.py`, `server_tools.py`.
- **No tests / docs** — adds a `test_kv_quant` unit binary (parse, precedence, matrix, TQ3 backward-compat) and a README section.

Accepted types widened from 4 to 8 (`f16, bf16, q4_0, q4_1, q5_0, q5_1, q8_0, tq3_0`, case-insensitive). Legacy `DFLASH27B_KV_F16/Q4/TQ3` and `--kv-tq3/--kv-q4/--kv-f16` retained as symmetric shorthand; per-axis vars/flags take precedence.

## TQ3_0 semantics under asymmetric

K and V are stored unrotated. The FWHT rotation is purely a Q transform with the inverse applied to the FA output, gated solely on `kv_k_type == TQ3_0`. So asymmetric pairs need no extra FWHT logic — `K=TQ3, V=Q8` and `K=Q8, V=TQ3` both work without changes to the rotation site.

## Test plan

- [x] `test_kv_quant` unit tests pass (T1 parse, T2 precedence, T3 matrix, T4 TQ3 shorthand)
- [x] Smoke on RTX 3090 across nine pairs (`F16/F16`, `Q4_0/Q4_0`, `Q4_0/Q8_0`, `F16/Q4_0`, `F16/Q8_0`, `Q8_0/Q4_0`, `TQ3_0/TQ3_0`, `TQ3_0/Q8_0`, `Q8_0/TQ3_0`) — all coherent
- [x] `DFLASH27B_KV_K=tq3_0 DFLASH27B_KV_V=q5_0` aborts at allocation with full supported-pair listing
- [x] Backward-compat: `DFLASH27B_KV_TQ3=1` and `--kv-tq3` produce identical output to baseline
- [x] Reviewer to sanity-check that dropping the inline blocks from #54 doesn't lose any functionality (`resolve_kv_types` covers their precedence: legacy shorthand last-wins, then per-axis overrides)

### DIFF
--- a/dflash/CMakeLists.txt
+++ b/dflash/CMakeLists.txt
@@ -42,6 +42,7 @@ if(NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/deps/llama.cpp/ggml/CMakeLists.txt")
         "deps/llama.cpp submodule missing. Run: "
         "git submodule update --init --recursive")
 endif()
+set(GGML_CUDA_FA_ALL_QUANTS ON CACHE BOOL "Compile fattn kernels for all quant pairs (needed for asymmetric KV-quant)" FORCE)
 add_subdirectory(deps/llama.cpp/ggml EXCLUDE_FROM_ALL)
 
 # ─── dflash27b static library ──────────────────────────────────────
@@ -53,6 +54,7 @@ add_library(dflash27b STATIC
     src/qwen35_target_graph.cpp
     src/qwen3_dflash_graph.cpp
     src/kv_cache.cpp
+    src/kv_quant.cpp
     src/f16_convert.cu
     src/delta_net_chunked.cpp
 )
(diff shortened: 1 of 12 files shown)


davide221:
Screenshot 2026-04-29 alle 00 16 07

PR #57, opened 2026-04-28
megakernel: WMMA tensor-core rewrite of pf_dn_chunk_phase2 (closes part of #5)

dusterbloom:
## Why this PR

Issue #5 reports prefill performance well below the README's 37,800 pp520 figure on RTX 3090. Multiple users (chfwinter, ateamlunchbox, dknos) saw ~8.5k tok/s. dknos diagnosed cuBLAS MAGMA fallback on their sm_120 setup and proposed an `at::mm_out` swap.

On this RTX 3090 (sm_86) I confirmed via `torch.profiler` that the GEMMs already land cleanly on tensor cores (`ampere_bf16_s16816gemm…`, `cutlass_80_tensorop_*`) — **the MAGMA fallback dknos saw is not the issue here**. Untouched, this machine sits at ~17,800 tok/s pp512 (twice issue #5's 8.5k baseline, but still half the README claim).

The hottest residual on this hardware is `pf_dn_chunk_phase2` — 47.6% of CUDA time, all scalar FP32 inner products in shared memory. This PR rewrites the matmul-shaped pieces of that kernel onto WMMA bf16 tensor cores (m16n16k16, f32 accumulate).

## What's in the PR

A second kernel `pf_dn_chunk_phase2_wmma` lives alongside the scalar reference; runtime env var `MEGAKERNEL_DN_PHASE2_WMMA_RUNTIME=1` flips between them — no rebuild required for A/B testing. Default behavior (`=0` or unset) is the unchanged scalar kernel, so this PR is opt-in for users.

Eight commits, each documented and benchmarked independently:

| | Phase | What | tok/s |
|---|---|---|---:|
| `085fc51` | 1 | build-flag scaffold | 17,847 |
| `946dc87` | 0+1 | instrumentation + 50-prompt parity corpus + scope plan | 17,847 |
| `c2d866a` | 2 | WMMA d-compute (M=8 pad 16, N=32 split as 2 N-tiles, K=128) | 18,273 |
| `e70c218` | 3 | WMMA o_inter qs (same shape as d) | 19,348 |
| `77258df` | 4 | WMMA state update (M=32 split as 2, N=128 split as 8, K=8 pad 16); accumulator preload + per-fragment γ scale | 20,470 |
| `6c719c7` | 5 | `__launch_bounds__(DN_PHASE2_BLOCK, 2)` — smem caps occupancy at 2 blocks/SM, declaring (128, 2) lets the compiler exploit it | 20,970 |
| `2bac32e` | 5b | drop dead `s_w` f32 (only `s_w_bf16` is read in WMMA path), tighten smem allocation so 2 blocks really fit per SM | unchanged |
| `ebe12fd` | 7 | precompute `exp(cs[c])` and `exp(cs[c]-cs[sp])` tables once per chunk — 1408 exps/chunk → 44 (32× redundancy fix) | 20,733 |
| `7d82707` | 8 | paired bf16 conversion (`__floats2bfloat162_rn`) for the per-chunk `s_state_bf16` mirror refresh | 20,642 (best 21,347) |

## Performance — RTX 3090, 512-token prefill, qwen3.6-27b-dflash

```
                            Prefill ms  tok/s   phase2 µs/launch
Scalar (current main)            28.69  17,847       710
WMMA Path B (this PR)            24.42  20,470       482
                                 -14.9% +15-18%      -32%
```

**Phase2 kernel time -32%** (12.79 ms → 8.7 ms across 18 layer launches). Total prefill +15-18% depending on the noise window — variance on the non-phase2 sections widens the spread.

Above the headline number: the README's 37,800 pp520 is not reachable from this branch. Every WMMA path I could hit cleanly is on tensor cores, and the remaining work (smem latency, sync overhead, scalar bookkeeping) doesn't have an obvious lever in the same magnitude. To close the rest would need structural change — persistent kernel, vectorized cp.async (attempted in this branch and reverted, see below), or revisiting whether the figure is reproducible on stock 3090 with locked clocks.

## Correctness

Parity gate built up in the first commit: 50 fixed prompts × 32 generated tokens, captured as `corpus/baseline.json` from current main and re-captured at every phase as `corpus/wmma_pX.json`.

Across all 8 phases the WMMA kernel matches the scalar kernel on **47 / 50 first generated tokens**. The 3 divergent prompts are all numeric-answer (factorial code, square root of 144, speed of light) where the logit margin between adjacent number/format tokens is razor-thin and bf16 truncation flips argmax under quantization noise. Mid-decode divergence in the 47 matching prompts is expected — bf16 state truncation propagates through subsequent decode steps.

Two additional sanity checks done locally:
- `bench_pp_tg.py` correctness section (`The capital of France is`) → predicts the same first token in scalar and WMMA modes.
- 9-pair smoke comparing scalar reference against WMMA on a varied prompt set, all coherent output.

## What didn't work — Phase 6 cp.async (reverted before commit)

I implemented a third kernel `pf_dn_chunk_phase2_wmma_pipe` with double-buffered f32 staging via `__pipeline_memcpy_async`. Goal: overlap chunk N+1 loads with chunk N compute. Result: bit-identical first-token output to the unpipelined kernel, **zero wall-time gain** (25.07 ms vs 25.05 ms).

Two contributing factors:

1. Per-element 4-byte `cp.async` issuance — each thread issued ~27 cp.async calls per chunk. Issue overhead at this granularity ate the overlap gain. Vectorized 16-byte loads via `float4` would help (4× fewer instructions) but require 16-byte aligned smem destinations, and the bank-conflict pad `DK_S = 129 floats` breaks alignment for non-zero rows.

2. Smem grew from ~50 KB (Phase 5 with `(128, 2)`) to ~67 KB. The kernel falls to 1 block/SM, giving back the +3% Phase 5 occupancy gain — net wash.

The working `cp.async` infrastructure is preserved in the branch history for future use if anyone wants to vectorize cp.async with a realigned smem layout.

## Test plan

- [x] `MEGAKERNEL_DN_PHASE2_WMMA_RUNTIME=1 python megakernel/diag_prefill_kernels.py` — kernel mix verdict + tok/s
- [x] `MEGAKERNEL_DN_PHASE2_WMMA_RUNTIME=1 python megakernel/build_corpus.py` then diff vs `corpus/baseline.json` — parity gate
- [x] `python megakernel/bench_pp_tg.py` correctness section under both env values
- [x] Reviewer: `MEGAKERNEL_DN_PHASE2_WMMA_RUNTIME=0` (default) is unchanged from current main — no risk to existing users
- [x] Reviewer: confirm the WMMA path is only ever triggered explicitly via env var (no implicit dispatch)

## Suggested follow-ups (not in this PR)

- Vectorized cp.async with realigned smem layout (`DK_S = 132` and accept slight bank conflict, or reorganize to avoid stride pad entirely).
- Run `ncu` on bare-metal Linux (WSL2 blocks GPU performance counters via `ERR_NVGPUCTRPERM`) to identify the real residual bottleneck. Right now we're picking optimizations from first principles + `torch.profiler` timing only.
- Backport the WMMA pattern into the Blackwell `prefill_megakernel.cu` if the author of that file thinks the persistent-kernel structure is compatible.
- Default the WMMA path on for all users once the 3-prompt argmax-flip correctness gap is investigated (likely fine for production prefill but currently opt-in to be safe).


davide221:
Screenshot 2026-04-29 alle 00 34 25

PR #63, opened 2026-04-30
dflash: add reset_target_cache for fast daemon mode cache reuse

howard0su:
In daemon mode, the old code destroyed and recreated the entire GPU cache (KV + SSM + conv + rollback + target_feat) between requests, costing ~5 seconds of allocation overhead per request. This halved throughput from 46 to 22 tok/s.

Fix: add reset_target_cache() that zeros all state tensors in-place without freeing/reallocating GPU buffers. Daemon mode now calls step_graph_free + reset_target_cache instead of step_graph_destroy + free_target_cache + create_target_cache.

Also make migrate_prefill_cache() a no-op when rollback tensors already exist (daemon second+ request).

Results on RTX 2080 Ti:
  1st request (cold, with migrate): 35 tok/s
  2nd+ requests (warm):             47 tok/s (was 22, now matches standalone)

### DIFF
--- a/dflash/src/internal.h
+++ b/dflash/src/internal.h
@@ -255,6 +255,11 @@ bool create_target_cache(const TargetWeights & w,
 
 void free_target_cache(TargetCache & c);
 
+// Zero all state tensors (KV, SSM, conv, target_feat, rollback) in place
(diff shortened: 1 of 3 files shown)


davide221:
Great work @howard0su! Thanks for the contribution

PR #71, opened 2026-05-01
dflash: add Q8_0 draft quantization support

howard0su:
Add the ability to quantize the DFlash draft model from BF16 to Q8_0, reducing VRAM usage from 3.46 GB to 1.84 GB (53% of original), at the same time, token/s increase about 8% on my configuration (2080ti).

New files:
- scripts/quantize_draft_q8.py: BF16 safetensors → Q8_0 GGUF converter. Projection weights quantized to Q8_0, norms kept as F32.
- src/gguf_draft_loader.cpp: GGUF-based draft loader, mirrors the target GGUF loader pattern. Validates arch metadata and all dimensions.

Modified files:
- src/internal.h: add load_draft_gguf() declaration
- test/test_dflash.cpp: auto-detect .gguf draft path
- CMakeLists.txt: add gguf_draft_loader.cpp to build

Benchmark results (RTX 2080 Ti, HumanEval 10-prompt, n_gen=256):

|  Draft     | Budget  | AL     | tok/s|   Notes|
|  -------   | ------  | -----  | ------  | -----|
|  BF16      |15      |10.68  |111.66|
|  Q8_0      |15      |10.78  |120.64  |+0.9% AL, +8.0% tok/s|
|  Q8_0      |22      |11.32   |95.40|

Q8_0 at budget=15: identical AL, 8% faster (smaller draft = faster ggml_mul_mat). Q8_0 at budget=22: higher AL (11.32 vs 10.68) but verify overhead reduces net tok/s vs budget=15 on this GPU.

Usage:
  python3 scripts/quantize_draft_q8.py models/draft/model.safetensors models/draft/draft-q8_0.gguf
  DFLASH_DRAFT=models/draft/draft-q8_0.gguf python3 scripts/bench_he.py

PR #72, opened 2026-05-01
perf(pflash): add SM75 target-resident TTFT path

weicj:
# perf(pflash): add SM75 target-resident TTFT path

## Summary

This PR adds an opt-in SM75 / RTX 2080 Ti path for PFlash TTFT-oriented use cases.

- Adds FP16 drafter compute support for SM75, including BF16->F16 GGUF load conversion when native BF16 tensor cores are unavailable.
- Tunes the WMMA FlashPrefill fallback for SM75 with padded shared-memory layout and `DFLASH_FP_K_TILE=32`.
- Adds PFlash chunk selection with rare-query lexical rescue, plus `test_pflash_chunk_select`.
- Adds opt-in target-resident PFlash daemon flow:
  - `DFLASH_PFLASH_KEEP_TARGET=1`
  - `DFLASH_PFLASH_SKIP_DRAFT_RELOAD=1`
  - default-off fallback when the DFlash draft remains parked
  - skip `migrate_prefill_cache` in that fallback because rollback tensors are not used
- Documents the SM75 benchmark as an experimental TTFT path, not as full PFlash + DFlash speculative decode.

## Benchmark

Hardware and environment:

- GPU: RTX 2080 Ti 22 GB / SM75
- Driver: 595.58.03
- CUDA toolkit: 12.0
- CMake: 3.28.3
- Power limit: 280 W
- Persistence mode: enabled
- Target: Qwen3.6-27B Q4_K_M via `test_dflash`
- PFlash drafter: Qwen3-0.6B FP16 GGUF
- Prompt: same 16K NIAH qtail prompt
- Methodology: warm daemon request timing after model load; tokenizers treated as preloaded; same hardware, same power limit, same prompt

| Case | Request time | Speedup | Notes |
|------|-------------:|--------:|-------|
| no PFlash | 50.35 s | 1.00x | original 16K prompt |
| current PFlash hook | 26.11 s | 1.93x | parks and reloads target + draft |
| `DFLASH_PFLASH_KEEP_TARGET=1` | 14.19 s | 3.55x | target stays resident, draft reloads |
| `KEEP_TARGET=1` + `SKIP_DRAFT_RELOAD=1` | 4.13 s | 12.21x | TTFT-only fallback; no speculative decode |

## Correctness / Validation

- `cmake --build build-sm75-f16 --target test_dflash test_pflash_chunk_select test_flashprefill_kernels -j$(nproc)` passes.
- `test_pflash_chunk_select` passes.
- `test_flashprefill_kernels` passes on SM75:
  - mean vector max diff `0.00000`
  - sparse forward max diff `0.00043`
  - S=8192 e2e FlashPrefill `13.8 ms / iter`
- `_prefill_hook.py` passes `python -m py_compile`.
- Clean PR-worktree SM75 build passes:
  `cmake --build dflash/build-sm75-pr --target test_dflash test_pflash_chunk_select test_flashprefill_kernels -j24`.
- 16K NIAH quality smoke retained the key and answer on the original prompt and a 5-position synthetic NIAH sweep.

## Caveats

- This is a PFlash TTFT path. It intentionally does not claim new DFlash/DDTree decode results on RTX 2080 Ti.
- With `DFLASH_PFLASH_SKIP_DRAFT_RELOAD=1`, the draft remains parked after compressed prefill. This is a default-off fallback for TTFT / very short output, not a decode-speed path.
- A keep-target + DFlash draft reload countercheck at `max_ctx=17000` hit a 106 MiB CUDA allocation OOM in the draft graph on RTX 2080 Ti.
- The quality validation here is retrieval-style NIAH smoke, not broad Math/GSM/code/chat validation.
- The new residency flags remain default-off.

### DIFF
--- a/dflash/CMakeLists.txt
+++ b/dflash/CMakeLists.txt
@@ -56,6 +56,7 @@ add_library(dflash27b STATIC
     src/qwen35_target_graph.cpp
     src/qwen3_dflash_graph.cpp
     src/qwen3_drafter.cpp
+    src/pflash_chunk_select.cpp
     src/qwen3_0p6b_loader.cpp
     src/qwen3_0p6b_graph.cpp
     src/flashprefill_kernels.cu
@@ -169,7 +170,13 @@ if(DFLASH27B_TESTS)
     if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test/test_flashprefill_kernels.cpp")
         add_executable(test_flashprefill_kernels test/test_flashprefill_kernels.cpp)
         set_target_properties(test_flashprefill_kernels PROPERTIES CUDA_ARCHITECTURES "${_dflash27b_archs}")
+        target_compile_definitions(test_flashprefill_kernels PRIVATE DFLASH27B_MIN_SM=${_dflash27b_min_sm})
         target_link_libraries(test_flashprefill_kernels PRIVATE dflash27b CUDA::cudart)
+    endif()
+    if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test/test_pflash_chunk_select.cpp")
+        add_executable(test_pflash_chunk_select test/test_pflash_chunk_select.cpp)
+        target_include_directories(test_pflash_chunk_select PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src)
+        target_link_libraries(test_pflash_chunk_select PRIVATE dflash27b)
     endif()
     if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test/test_kv_quant.cpp")
         add_executable(test_kv_quant test/test_kv_quant.cpp)
--- a/dflash/docs/SPEC_PREFILL.md
+++ b/dflash/docs/SPEC_PREFILL.md
@@ -51,6 +51,66 @@ DFLASH_FP_PROFILE=1    # log mean/score/select/forward stage timings
 
 See `src/flashprefill.h` for the full list and defaults.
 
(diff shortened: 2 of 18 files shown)


howard0su:
I tried this path as well to convert draft model from BF16 to FP16 in order to leverage 2080Ti's Tensor Core. Based on some experiment, I prefer using Q8_0. check PR #71 

1. 2080 only has 22GB VRAM. Q8_0 will save about 1.35GB which technically match the main experiment platform 3090 24GB.
2. No need to use BF16 on draft model as our main model is already quantized pretty aggressively already.
3. A small perf gain compare to FP16 (8.0% on my machine.)

weicj:
Thanks. I pushed an update in 4645fa8.

Fixes for the three issues identified by cubic:

- `test_flashprefill_kernels`: numerical validation now fails immediately on non-finite reference/output/diff values, so NaN/Inf cannot be hidden by max-diff accumulation.
- `qwen3_0p6b_loader`: BF16 capability is now checked from the active CUDA device at runtime via `cudaGetDeviceProperties`, while `DFLASH27B_DRAFT_FP16=1` still forces the FP16 path. This avoids mixed-arch false negatives on SM80+ devices.
- `test_dflash`: generate now rejects requests while target weights are parked, before entering target-only or speculative decode.

I also agree Q8_0 is a good direction for the 2080 Ti path, especially for the VRAM and perf reasons in #71. I kept this PR scoped to the FP16/BF16->F16 SM75 enablement plus PFlash TTFT/residency path so it does not collide with #71; happy to either rebase onto the Q8_0 draft path after #71 lands or split a follow-up patch for that.

PR #73, opened 2026-05-01
perf(dflash): use ggml_swiglu_split for FFN

howard0su:
FFN: Replace separate silu(gate) * up with ggml_swiglu_split(gate, up). This reduces graph nodes from 3059 to 2995 (-64, one per layer) and enables ggml-cuda's MUL_MAT+MUL_MAT+GLU three-way fusion, which merges the gate matvec, up matvec, and SwiGLU activation into a single kernel launch. Saves 192 kernel launches per forward (128 from triple fusion + 64 from node elimination).

Output is bit-identical to the previous separate silu+mul path.


davide221:
Thanks for the contribution Howard 💪

PR #74, opened 2026-05-01
    perf(dflash): GPU argmax for draft tokens, verify + DDTree accept path

howard0su:
Use ggml_argmax 
1. in the draft graph to extract draft_tok on GPU instead of CPU argmax over 248K vocab. In DDTree mode (K=8), full logits are still transferred for extract_draft_topk (real log-probs needed for tree scoring), but draft_tok extraction no longer requires 16 separate CPU argmax passes. 
2. target_tok extraction. Transfers only N×4 bytes (token IDs) instead of N×248K×4 bytes (logits).

draft_logits: 18.4ms → 13.4ms (-5ms, GPU argmax saves CPU work)
accept:       11.1ms →  0.1ms (-11ms, from previous commit)
        - Eliminates 22.8MB GPU→CPU logits transfer per verify step
        - Posterior argmax computed on GPU as part of the compute graph

### DIFF
--- a/dflash/test/test_dflash.cpp
+++ b/dflash/test/test_dflash.cpp
@@ -657,7 +657,11 @@ static bool build_target_step(
     sg.logits = go.logits;
     sg.delta_captures = std::move(go.delta_captures);
     ggml_set_output(sg.logits);
-    ggml_build_forward_expand(sg.gf, sg.logits);
+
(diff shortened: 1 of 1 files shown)


davide221:
Good speed up!

howard0su:
I realized a problem. this change only applies when we don't have top-p and temperature support. we may have to do differently when top-p and temperature is supported later. @davide221

PR #76, opened 2026-05-01
dflash : fix CUDA toolkit include path for C++ sources

pintaric:
## Summary

  `flashprefill.cpp` and `qwen3_0p6b_graph.cpp` include `<cuda_runtime.h>` directly, so the CUDA toolkit headers must be available when compiling the library.

  This adds `find_package(CUDAToolkit REQUIRED)` and exposes `${CUDAToolkit_INCLUDE_DIRS}` to `dflash27b`.

  ## Test

  - `CCACHE_DISABLE=1 cmake --build build --target test_dflash test_flashprefill_kernels -j`


davide221:
@pintaric thanks for the contribution!

PR #78, opened 2026-05-02
bench(pflash): add dual-GPU PFlash phase-split harness

weicj:
# bench(pflash): add dual-GPU PFlash phase-split harness

## Summary

Adds an opt-in harness for running the PFlash prefill phase as a persistent CUDA process, optionally on a separate GPU from later target prefill/generation work.

The harness is intentionally narrow:

- `pflash_daemon` keeps the Qwen3-0.6B PFlash drafter resident;
- `scripts/phase_split_dual_gpu.py` sends counted token IDs to that daemon;
- `--pflash-gpu` selects the CUDA GPU used for the PFlash phase;
- reports record compressed token/text outputs, PFlash timing, and GPU resource peaks.

PFlash is a prefill-side optimization. This harness measures the PFlash/prefill phase and writes the compressed prompt artifact; it does not measure or modify decode.

## Motivation

PFlash is most useful when a long source prompt can be compressed before target prefill. For long-context experiments, the PFlash phase benefits from being isolated as a resident CUDA worker: the drafter load cost is separated from per-request compression time, and the PFlash GPU role can be measured independently.

This gives maintainers and contributors a small reproducible surface for PFlash phase-split experiments: load the drafter once, compress prompts repeatedly, and inspect timing plus resource use from generated JSON/Markdown reports.

## Before / After

Before this change, PFlash experiments did not have a dedicated phase-split harness. Compression timing could be mixed with drafter startup or later target work, and there was no standard way to pin only the PFlash phase to a selected CUDA GPU while recording its resource use.

After this change, the PFlash drafter can stay resident in `pflash_daemon`, repeated prompts can be compressed through the same path, and each run emits compressed prompt artifacts plus timing/resource reports for the PFlash phase.

Local capacity evidence from the motivating phase-split prototype on two 22GB RTX 2080 Ti GPUs:

| setup | observed passing source context | compressed context | pass condition |
|---|---:|---:|---|
| single-GPU/co-resident PFlash + target | 24,573 tokens | 1,309 tokens | NIAH key/answer retained and target output matched |
| dual-GPU phase split | 196,605 tokens | 9,949 tokens | NIAH key/answer retained and target output matched |
| dual-GPU phase split | 229,365 tokens | 11,573 tokens | NIAH key/answer retained and target output matched |
| dual-GPU phase split | 262,125 tokens | 13,229 tokens | NIAH key/answer retained and target output matched |

The 262K case is about 10.7x the observed single-GPU/co-resident passing source context in this sweep. It is the largest passing point from this local run, not a claimed universal limit. This PR keeps the upstream patch focused on the PFlash-side harness that makes this kind of capacity testing reproducible.

## Why This Is Useful

- PFlash compression can be benchmarked as its own prefill-side phase.
- The drafter load cost is separated from per-request compression time.
- The PFlash phase can be pinned to a specific CUDA GPU with `--pflash-gpu`.
- Reports include source token count, compressed token count, compression ratio, PFlash tok/s, and GPU resource peaks.
- Synthetic NIAH prompts can check whether a known key/answer survives compression.
- Real prompt files or stdin can be compressed through the same daemon path.

## What Changed

- Added `test/pflash_daemon.cpp`.
  - Loads the Qwen3-0.6B PFlash drafter once.
  - Accepts `compress <keep_x1000> <lookahead> <chunk> <pool> <counted_ids.bin>` over stdin.
  - Streams compressed token IDs through the existing int32 sentinel protocol.
- Added `scripts/phase_split_dual_gpu.py`.
  - `bench-niah` builds synthetic prompts and records key/answer retention after compression.
  - `run-prompt` compresses file/stdin prompts.
  - Reports JSON/Markdown summaries and GPU monitor output for the PFlash phase.
- Added the `pflash_daemon` CMake target.
- Documented the harness in `docs/SPEC_PREFILL.md`.

## Validation

- Build: `pflash_daemon` builds locally.
- Script check: `phase_split_dual_gpu.py` passes Python syntax checks.

### DIFF
--- a/dflash/CMakeLists.txt
+++ b/dflash/CMakeLists.txt
@@ -162,6 +162,14 @@ elseif (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
      )
 endif()
 
+# PFlash phase-split harness
+
+if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test/pflash_daemon.cpp")
+    add_executable(pflash_daemon test/pflash_daemon.cpp)
+    target_include_directories(pflash_daemon PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src)
+    target_link_libraries(pflash_daemon PRIVATE dflash27b ggml ggml-cuda)
+endif()
+
 # ─── Tests (numerics vs oracle) ────────────────────────────────────
 
 option(DFLASH27B_TESTS "Build numerics tests" ON)
--- a/dflash/docs/SPEC_PREFILL.md
+++ b/dflash/docs/SPEC_PREFILL.md
@@ -51,6 +51,51 @@ DFLASH_FP_PROFILE=1    # log mean/score/select/forward stage timings
 
 See `src/flashprefill.h` for the full list and defaults.
 
+## Dual-GPU PFlash phase split
+
+PFlash targets the prefill side of long-context requests. The dual-GPU phase
+split harness is an opt-in benchmark/runtime path for measuring the PFlash
+prefill phase as its own resident CUDA process:
+
+- `pflash_daemon` keeps the Qwen3-0.6B PFlash drafter resident.
+- `scripts/phase_split_dual_gpu.py` sends counted token IDs to the daemon.
+- `--pflash-gpu` selects the CUDA GPU used for the PFlash phase.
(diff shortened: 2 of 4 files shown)


davide221:
Thanks for this great contribution @weicj it is the right direction

PR #80, opened 2026-05-02
bench(dflash): add dual-GPU target/draft split harness

weicj:
## Summary

This PR adds an opt-in dual-GPU target/draft split harness for DFlash speculative decoding experiments.

The target model/cache can stay on one CUDA GPU while the DFlash draft graph runs on another CUDA GPU. The default same-device path remains unchanged.

## Why Useful

Today the DFlash test path effectively assumes the target and draft work fit on the same CUDA backend. For larger quantized targets, that can make local experimentation difficult because the target model/cache and draft working buffers compete for the same GPU memory.

DFlash has a useful split boundary: the draft consumes target features and returns draft hidden states, which can then be projected by the target-side `lm_head`. That makes it possible for the benchmark path to place draft work on another GPU while keeping the target model resident on the main GPU.

## What Changed

- Add explicit DFlash CUDA placement controls:
  - `--target-gpu`
  - `--draft-gpu`
  - `DFLASH_TARGET_GPU`
  - `DFLASH_DRAFT_GPU`
- Build the target backend and draft backend on the selected CUDA devices.
- Add an opt-in draft feature mirror path for split execution.
  - The draft GPU keeps an F32 mirror of the target features needed by the DFlash draft graph.
  - Initial and incremental updates are synchronized from the target GPU with peer copies.
  - The mirror is window-bounded rather than a full-context host staging path.
- In split mode, run the DFlash draft graph hidden-only when the target `lm_head` is not local.
- Add a target-side projection graph that applies the target `lm_head` to draft hidden states after they are copied back to the target GPU.
- Extend the HumanEval benchmark helper to pass split-GPU flags and report DFlash decode/acceptance metrics.

## Notes

- The new path is opt-in.
- The default same-device DFlash path is unchanged.
- The benchmark helper now reports the split-GPU DFlash decode/acceptance metrics used below.

## Validation

- Built `test_dflash` with the local DFlash split changes.
- `scripts/bench_he.py` passes Python syntax checking.
- Local Q4 split-GPU smoke test passed on dual RTX 2080 Ti 22GB.

Local HumanEval-style validation on dual RTX 2080 Ti 22GB:

| Workload | Path | Decode | Notes |
| --- | --- | ---: | --- |
| HumanEval-style, 10 prompts, n_gen=128 | Qwen3.5-27B Q4 target + z-lab DFlash draft, split GPU | `51.86 tok/s` | AL `7.09`, accept `44.3%` |


davide221:
Screenshot 2026-05-04 alle 12 44 56
Lovely PR thanks @weicj

PR #83, opened 2026-05-03
fix(server): CORS, content blocks, health endpoint, event loop blocking, daemon crash guard

mohamed-em2m:
- Add CORSMiddleware (allow_origins=["*"]) so Open WebUI on port 3000
  can reach the API on port 8080 without browser CORS errors

- Add _content_to_str() helper and apply it to all ChatMessage/
  AnthropicMessage content fields; fixes 422 crash when clients send
  list[dict] content blocks instead of plain strings

- Add `stop` field to ChatRequest; Open WebUI sends it on every request,
  previously caused Pydantic 422 validation error

- Add GET /health endpoint returning 200/503; Open WebUI polls this on
  startup and shows "disconnected" permanently without it

- Enrich GET /v1/models response with context_length and created fields
  so Open WebUI renders the context bar correctly

- Replace blocking list(_token_stream()) and async-comprehension over
  _astream_tokens() in non-streaming paths with _collect_tokens_sync()
  via run_in_executor; previously stalled the entire event loop during
  generation

- Add _write_cmd() guard that checks daemon_proc.poll() before writing
  to stdin; previously a dead daemon caused a broken pipe hang with no
  useful error, now returns clean 503

- Extract _build_cmd_line() to deduplicate cmd_line construction that
  was copy-pasted across 4 paths and had begun to diverge

### DIFF
--- a/dflash/scripts/server.py
+++ b/dflash/scripts/server.py
@@ -4,16 +4,14 @@ OpenAI-compatible HTTP server on top of test_dflash.
     pip install fastapi uvicorn transformers
     python3 scripts/server.py                 # serves on :8000
 
-    curl http://localhost:8000/v1/chat/completions \\
-        -H 'Content-Type: application/json' \\
+    curl http://localhost:8000/v1/chat/completions \
+        -H 'Content-Type: application/json' \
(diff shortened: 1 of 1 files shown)


davide221:
@mohamed-em2m thanks for the cleaning, it is really useful in these initial phases 💪

PR #97, opened 2026-05-04
fix(dflash): use GGUF-loaded EOS ids instead of hardcoded 248045

anoane:
## Bug

Three sites in `test_dflash.cpp`'s spec-decode commit loops checked `tok == 248045` to terminate generation. Token 248045 is `<|im_start|>` — the chat-START marker, **not** EOS. Whenever the model legitimately emitted that token at the natural turn boundary `<|im_end|>\n<|im_start|>`, the loop terminated prematurely and the output included trailing chat-protocol garbage past the real `<|im_end|>`.

## Reproducer

```
PROMPT: "What is 2 + 2? Just the number."

before fix, gen tail: ... 19  248046  198  248044    248045
                          "4" <|end>  \n  <|eot> <|start>   ← stops here (wrong)

after  fix, gen tail: ... 19  248046  198
                          "4" <|end>  \n               ← stops here on real EOS
            (loader prints: eos_id=248046 eos_chat_id=-1
             from GGUF tokenizer.ggml.{eos,eot}_token_id)
```

## Fix

Read `tokenizer.ggml.eos_token_id` and `tokenizer.ggml.eot_token_id` from GGUF metadata into `TargetWeights::{eos_id,eos_chat_id}` at load time. Check against those via a single `IS_EOS_QWEN36_27B(tok, w)` macro. Missing metadata keys default to `-1`, and the `>= 0` guard ensures the sentinel never matches a real token id.

## Verified

Tested across Qwen3.5-27B and Qwen3.6-27B targets in BF16, Q4_K, Q4_K_M, Q6_K, NVFP4, NVFP4_M, NVFP4_Mv2 — all report `eos_id=248046` (= `<|im_end|>`) and stop cleanly there. Body content (including `<think>...</think>` blocks) is byte-identical to the previous binary up to the bug-fire point on every tested prompt.

## Files (3 files, +34 / −3)

- `dflash/src/internal.h` — `eos_id` / `eos_chat_id` fields on `TargetWeights`
- `dflash/src/gguf_target_loader.cpp` — read u32 metadata keys, cast to int32 with `-1` sentinel for missing keys
- `dflash/test/test_dflash.cpp` — `IS_EOS_QWEN36_27B(tok, w)` macro + 3 call-site replacements

### DIFF
--- a/dflash/src/gguf_target_loader.cpp
+++ b/dflash/src/gguf_target_loader.cpp
@@ -327,6 +327,19 @@ bool load_target_gguf(const std::string & path,
     out.ssm_dt_rank= (int)ssm_dt;
     out.ssm_n_group= (int)ssm_grp;
 
+    // EOS token ids from GGUF tokenizer metadata (stored as UINT32 by the
+    // GGUF spec; we use the u32 helper and cast). UINT32_MAX is the
+    // missing-key sentinel and maps to int32_t -1, which the runtime EOS
+    // check rejects via the `>= 0` guard.
+    {
+        const uint32_t kEosKeyMissing = 0xFFFFFFFFu;
+        const uint32_t raw_eos      = get_u32_or(gctx, "tokenizer.ggml.eos_token_id", kEosKeyMissing);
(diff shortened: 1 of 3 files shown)


davide221:
Thanks @anoane for your first contribution and bug fix

PR #98, opened 2026-05-04
fix(dflash): auto-detect CUDA architecture for Blackwell

easel:
Split out Blackwell auto-detection from workaround/consumer-specific logic so the workaround can be reviewed and merged independently.

What this changes
- Query `nvidia-smi` at configure time when `CMAKE_CUDA_ARCHITECTURES` is unset or `native`.
- Convert the reported compute capability (for example, `12.0`) into `CMAKE_CUDA_ARCHITECTURES`.
- Keep the change isolated to `dflash/CMakeLists.txt`.


easel:
Superseded by PR #48 (auto-detect) and PR #99 (workaround dependency).

PR #99, opened 2026-05-04
fix(dflash): set consumer Blackwell ggml flag when 12x arch selected

easel:
## Problem

Consumer Blackwell GPUs use `sm_12x` but not `sm_120a`, and ggml's
`sm_12X -> sm_12Xa` handling plus FP4 MMA kernels can trigger illegal-instruction
crashes.

## Fix

- Enable `GGML_CUDA_BLACKWELL_CONSUMER` automatically when the resolved
  CUDA arch list contains any `12x` entry.
- Add explicit override flag `DFLASH27B_USE_BLACKWELL_CONSUMER_FIX=ON` to force
  workaround behavior for custom cross-compile or custom arch-list flows.

## Notes

This PR now intentionally does not include the llama.cpp submodule pointer update;
it was dropped because the required dependency is already present in current upstream
submodule history.

## Test plan

- [ ] Configure and build with a consumer Blackwell setup to confirm
  `GGML_CUDA_BLACKWELL_CONSUMER=ON`.
- [ ] Validate runtime no longer trips `CUDA_ERROR_ILLEGAL_INSTRUCTION` in the FP4 MMQ
  path without applying auto-detect workaround.

### DIFF
--- a/dflash/CMakeLists.txt
+++ b/dflash/CMakeLists.txt
@@ -44,6 +44,33 @@ endif()
 # the spec_prefill demo (target_gen path uses standard quant pairs).
 option(DFLASH27B_FA_ALL_QUANTS "Compile ggml-cuda fattn kernels for all KV-quant pairs" ON)
 set(GGML_CUDA_FA_ALL_QUANTS ${DFLASH27B_FA_ALL_QUANTS} CACHE BOOL "" FORCE)
+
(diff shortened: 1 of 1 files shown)


easel:
This may not be necessary if expectation is to always build multi-arch binary. I ran into it because claude got excited about optimizing and ended up with a slightly incompatible build.

davide221:
LGTM, merging

PR #104, opened 2026-05-05
perf: SSM intermediates F16 → Q8_0 (saves 540 MB VRAM)

howard0su:
Change rollback intermediate storage from F16 to Q8_0, reducing the fixed VRAM cost from 1152 MB to 612 MB (48 delta layers × [128,128,48,16] per-step state captures).

Measured Q8_0 quantization error: 41 dB SNR, <1% relative RMSE. Intermediates are only used for rollback after rejected tokens — the small precision loss self-corrects on the next accepted token.

Changes:
- qwen35_target_graph.cpp: ssm_intermediate tensor type F16 → Q8_0
- qwen35_target_graph.cpp: disable tree_persist for Q8_0 (kernel only supports F32/F16 output), fall back to ggml_cpy F32→Q8_0
- f16_convert.cu: add dflash27b_launch_q8_0_to_f32 dequant kernel
- test_dflash.cpp: dispatch Q8_0 or F16 dequant based on tensor type

Output quality: identical fibonacci reasoning. Same tok/s (47 vs 46). Acceptance rate unchanged (26.6%, 4.26 tok/step).


dflash/src/f16_convert.cu
```diff
@@ -47,3 +64,13 @@ extern "C" void dflash27b_launch_bf16_to_f32(const void * src,
     bf16_to_f32_kernel<<<blocks, threads, 0, stream>>>(
         (const __nv_bfloat16 *)src, (float *)dst, n_elems);
 }
+
+extern "C" void dflash27b_launch_q8_0_to_f32(const void * src,
```
howard0su: Not sure if there is ggml version of this.

davide221:
2 compile errors:                    
- test_dflash.cpp:647: 'dflash27b_launch_bf16_to_f32' was not declared                             
- test_dflash.cpp:658: 'dflash27b_launch_bf16_to_f32' was not declared

PR #108, opened 2026-05-06
perf: Replace Q8_0 format for KV with Q4_0 + Rotation, fix window_filled for long context

howard0su:
Three prefill optimizations that eliminate the Turing-specific ubatch scaling workaround and make long-context prefill dramatically faster:

1. FA windowed prefill: use g_fa_window (default 2048) during prefill, capping attention cost at O(ubatch × window) instead of O(ubatch × kv_len). Compute time stays flat at ~1.15s/step across 100K context (was 3.5s→13s).

2. last_token_logits_only: skip lm_head matmul for all but the last token during prefill (only last_tok is needed to seed decode). Saves ~233MB scratch at ubatch=384 and eliminates a large [vocab × ubatch] matmul.

3. gallocr pre-reserve: build max-shape graph before prefill loop so gallocr allocates its CUDA buffer once upfront, preventing hundreds of reallocations as the mask grows.

### DIFF
--- a/dflash/src/internal.h
+++ b/dflash/src/internal.h
@@ -217,6 +217,12 @@ struct TargetCache {
     ggml_type kv_k_type = GGML_TYPE_Q8_0;
     ggml_type kv_v_type = GGML_TYPE_Q8_0;
 
+    // When true, K is FWHT-rotated in the graph before writing to the
+    // standard-type cache (Q4_0/Q8_0/etc), and Q is rotated at attention
+    // time. This gives TurboQuant-style outlier spreading with fast FA
+    // kernels that work on all GPU architectures.
+    bool kv_k_rotated = false;
+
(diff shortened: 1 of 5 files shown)


howard0su:
need additional change in llama.cpp
[0001-turbo_wht-support-strided-input-fused-wht-quantize-k.patch](https://github.com/user-attachments/files/27427242/0001-turbo_wht-support-strided-input-fused-wht-quantize-k.patch)

howard0su:
need double check the quality of the output

davide221:
Screenshot 2026-05-07 alle 11 58 35
Screenshot 2026-05-07 alle 11 58 11

@howard0su

howard0su:
Looks good although I cannot explain why AL is increased.

PR #110, opened 2026-05-06
fix: fix Sparce attention and Qwen loader on windows

aminya:


davide221:
@aminya thanks for the contribution, lgtm

PR #122, opened 2026-05-07
bench(dflash,pflash): add CUDA/HIP mixed backend placement

weicj:
# bench(dflash,pflash): add CUDA/HIP mixed backend placement

## Summary

Add CUDA/HIP mixed backend placement support for PFlash and DFlash bench harness execution.

This PR makes the PFlash/DFlash harness split paths backend-portable instead of CUDA-only or multi-GPU-specific. The same implementation can be built as separate CUDA or HIP binaries, then combined by the harness when CUDA/HIP mixed backend placement is needed:

- single backend / single device, where the split path still runs inside one selected GPU backend;
- single backend / multiple devices, where target execution may use the existing target layer split;
- CUDA/HIP mixed backend, where PFlash phase split or DFlash draft split crosses a host-data/process boundary.

This PR keeps mixed placement in the bench/runtime harnesses first, so the multi-device and multi-backend behavior can be validated directly before the OpenAI-compatible server path is integrated in a follow-up PR.

## Changes

- Add `DFLASH27B_GPU_BACKEND=cuda|hip` to the DFlash CMake build.
- Link DFlash runtime/test targets against the selected ggml backend (`ggml-cuda` or `ggml-hip`).
- Add a small HIP compatibility shim for the current vendored ggml HIP snapshot.
- Route non-CUDA-WMMA PFlash execution through ggml `flash_attn_ext`.
- Pass the real Q/K/V `ggml_type` into the portable PFlash path so F16 and BF16 are not inferred from byte size.
- Make draft projection weight type selection backend-aware instead of using a CUDA-only SM macro.
- Extend `phase_split_dual_gpu.py` with:
  - `--pflash-backend cuda|hip`
  - `--pflash-visible-devices`
  - `--target-backend cuda|hip`
  - `--target-visible-devices`
  - optional target generation after PFlash compression
  - separate CUDA/HIP GPU resource monitoring
- Add DFlash draft IPC support to `test_dflash`:
  - `--draft-ipc-daemon`
  - `--draft-ipc-bin`
  - `--draft-ipc-gpu`
  - `--draft-ipc-work-dir`
  - `--draft-ipc-ring-cap`
- Extend `bench_he.py` so HumanEval-style DFlash runs can pass through the draft IPC binary/options used for mixed-backend draft split validation.
- Document CUDA/HIP mixed backend placement in `dflash/docs/MIXED_BACKEND.md` and keep `SPEC_PREFILL.md` focused on the spec-prefill daemon reference.

## CUDA/HIP Mixed Backend Design

PFlash and DFlash already have flexible split flows. This PR makes those flows explicit backend-placement points across separate CUDA/HIP builds instead of tying the harness to one GPU vendor or one fixed topology.

The mixed-backend boundary stays at phase/model/process level:

- PFlash phase split can place the PFlash drafter and target generation on different backends, or keep both on one backend.
- DFlash draft split can place the draft model in a separate backend process and feed a target process through host IPC.
- Target execution can remain single-device or use same-backend target layer split.

These split boundaries do not require moving GPU activations directly across vendors, so backend placement can be selected through build-time backend binaries plus runtime harness placement while cross-vendor target layer execution stays outside the design.

For PFlash, the boundary is host-side token/text data:

1. The PFlash daemon returns compressed drafter-token IDs.
2. The harness decodes those IDs to text.
3. The target tokenizer re-encodes that text for target generation.

For DFlash, the boundary is a separate draft process:

1. The target split path captures target-side feature slices.
2. The feature/noise inputs are passed through host IPC to the draft daemon.
3. The draft daemon returns hidden states to the target process for projection and verification.

This keeps mixed backend support compatible with single-device target runs and existing same-backend target layer split behavior while avoiding cross-vendor target layer execution.

## Validation

Validation used mixed CUDA/HIP hardware to cover the cross-backend case. CUDA-only, HIP-only, single-device, and same-backend multi-device placements use the same CMake backend selector and harness plumbing, but the table below focuses on the cases that prove both backend directions:

- CUDA side: RTX 2080 Ti (`sm_75`), NVIDIA driver `595.58.03`, CUDA toolkit `12.0.140` (`nvidia-smi` reports CUDA driver capability `13.2`).
- HIP side: AMD Radeon Pro VII (`gfx906`), ROCm `7.2.1`, HIP `7.2.53211`.

| Path | Direction | Draft/PFlash side | Target side | Model | Test | Result |
|---|---|---|---|---|---|---|
| PFlash phase split | HIP -> CUDA | HIP on 1x Pro VII (`gfx906`) | CUDA target split on 2x RTX 2080 Ti (`sm_75`) | Qwen3.6-27B-Q8_0 | NIAH 4K / 8K / 16K | Passed, key and answer retained, kept-token ratio 4.4-4.8% |
| PFlash phase split | CUDA -> HIP | CUDA on 1x RTX 2080 Ti (`sm_75`) | HIP target split on 2x Pro VII (`gfx906`) | Qwen3.6-27B-Q8_0 | NIAH 4K / 8K / 16K | Passed, key and answer retained, kept-token ratio 4.4-4.8% |
| DFlash draft split | HIP -> CUDA | HIP draft IPC daemon on 1x Pro VII (`gfx906`) | CUDA target split on 2x RTX 2080 Ti (`sm_75`) | Qwen3.6-27B-Q8_0 | HumanEval 10 prompts, n_gen=256 | Passed, mean AL 8.88, accept 50.8%, decode 35.68 tok/s |
| DFlash draft split | CUDA -> HIP | CUDA draft IPC daemon on 1x RTX 2080 Ti (`sm_75`) | HIP target split on 2x Pro VII (`gfx906`) | Qwen3.6-27B-Q8_0 | HumanEval 10 prompts, n_gen=256 | Passed, mean AL 8.69, accept 49.6%, decode 20.86 tok/s |

Additional checks:

- CUDA build passed for the mixed-backend branch.
- HIP build passed for the mixed-backend branch.
- `phase_split_dual_gpu.py` was exercised in both backend directions.
- `bench_he.py` was exercised with mixed-backend DFlash draft IPC in both backend directions.

### DIFF
--- a/dflash/docs/SPEC_PREFILL.md
+++ b/dflash/docs/SPEC_PREFILL.md
@@ -51,78 +51,11 @@ DFLASH_FP_PROFILE=1    # log mean/score/select/forward stage timings
 
 See `src/flashprefill.h` for the full list and defaults.
 
-## Hybrid PFlash phase split
+## Mixed backend placement
 
-PFlash targets the prefill side of long-context requests. The dual-GPU phase
-split harness is an opt-in benchmark/runtime path for measuring the PFlash
-prefill phase as its own resident process. It can now place the PFlash drafter
-phase and the later target generation phase on different ggml GPU backends:
-
-- `pflash_daemon` keeps the Qwen3-0.6B PFlash drafter resident.
-- `scripts/phase_split_dual_gpu.py` sends counted token IDs to the daemon.
-- `--pflash-backend cuda|hip` and `--pflash-visible-devices` select the
-  backend/device set used for the PFlash phase.
-- `--run-target` can launch `test_dflash` afterward with
-  `--target-backend cuda|hip` and `--target-visible-devices`.
-- The report records compressed token/text outputs, PFlash timing, target
-  generation timing, return code, and GPU resource peaks for the PFlash worker.
-
-The cross-backend boundary is host-side token/text data: the PFlash daemon
-returns compressed drafter-token IDs, the harness decodes them to text, and
-the target tokenizer re-encodes that text for `test_dflash`. This keeps the
-hybrid path compatible with the existing PFlash phase split, DFlash target/draft
-split, and DFlash target layer split harnesses without adding cross-vendor GPU
-activation transfers.
-
-Target layer split remains inside one backend binary. For example, use HIP
-PFlash on an AMD card with CUDA target layer split across two NVIDIA cards, or
-CUDA PFlash on an NVIDIA card with HIP target layer split across AMD cards.
-Cross-backend target layer split is intentionally out of scope for this phase.
-
-Build:
-
-```
-cmake -S . -B build-cuda -DCMAKE_BUILD_TYPE=Release \
-  -DDFLASH27B_GPU_BACKEND=cuda
-cmake --build build-cuda --target pflash_daemon test_dflash -j
-
-cmake -S . -B build-hip -DCMAKE_BUILD_TYPE=Release \
-  -DDFLASH27B_GPU_BACKEND=hip \
-  -DDFLASH27B_HIP_ARCHITECTURES=<your-gfx-arch>
-cmake --build build-hip --target pflash_daemon test_dflash -j
-```
-
-Run a HIP PFlash drafter followed by CUDA target layer split:
-
-```
-python scripts/phase_split_dual_gpu.py bench-niah \
-  --build-dir build-hip \
-  --pflash-backend hip \
-  --pflash-visible-devices 0 \
-  --run-target \
-  --target-bin build-cuda/test_dflash \
-  --target-backend cuda \
-  --target-visible-devices 0,1 \
-  --target-gpus 0,1 \
-  --target-layer-split 1,1 \
-  --target-gen-tokens 8 \
-  --contexts 4096 \
-  --local-files-only \
-  --report-dir reports/pflash_hybrid_hip_drafter_cuda_target
-```
-
-Compress a real prompt:
-
-```
-python scripts/phase_split_dual_gpu.py run-prompt \
-  --build-dir build \
-  --prompt-file /path/to/prompt.txt \
-  --local-files-only \
-  --report-dir reports/pflash_phase_split_prompt
-```
-
-The reports include source token count, compressed token count, compression
-ratio, PFlash timing, and GPU resource peaks.
+CUDA/HIP mixed backend placement uses separate CUDA and HIP builds at the
+PFlash phase or DFlash draft-process boundary. See
+[`MIXED_BACKEND.md`](MIXED_BACKEND.md) for the build and harness instructions.
 
 ## Performance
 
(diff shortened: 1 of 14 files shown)


dflash/test/test_dflash.cpp
```diff
@@ -1380,6 +1680,194 @@ static bool copy_feature_ring_range_to_tensor(
...
+static bool stream_status(int stream_fd, int32_t status) {
+#if defined(_WIN32)
+    (void)stream_fd; (void)status;
+    return false;
+#else
+    return write_exact_fd(stream_fd, &status, sizeof(status));
+#endif
+}
+
+static int run_dflash_draft_ipc_daemon(const char * draft_path,
```
howard0su, line 1693: can we abstract interface for draft? so we can have in-context (same gpu), two gpus, two gpus with amd/nvidia. put three different implementation into three files with some common/shared functions.

this file is too large to maintain now.

dflash/src/gpu_runtime_compat.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
```
howard0su, line 1: why not put under hip_compat?

dflash/docs/SPEC_PREFILL.md
```diff
@@ -51,36 +51,64 @@ DFLASH_FP_PROFILE=1    # log mean/score/select/forward stage timings
 
 See `src/flashprefill.h` for the full list and defaults.
 
-## Dual-GPU PFlash phase split
+## Hybrid PFlash phase split
```
howard0su: why override? but overall this file is not well structured. maybe you can replace a good user instruction.

dflash/src/flashprefill_q8.cpp
```diff
@@ -31,7 +33,7 @@ int flash_prefill_forward_q8(
     const void * Q, const void * K, const void * V, void * O,
     int batch, int seq_len, int n_q_heads, int n_k_heads, int head_dim,
     float scale,
-    int elem_size,
+    ggml_type qkv_type,
```
howard0su, line 36: good approach!!

dflash/src/gpu_runtime_compat.h
weicj: I kept it outside hip_compat because it is not a HIP-only shim. The hip_compat directory is currently used for vendored ggml HIP header compatibility, such as cuda_fp16.h. This header is our own runtime compatibility layer for the harness: CUDA builds include cuda_runtime.h through it, while HIP builds map the existing cuda* runtime spellings to hip*.

dflash/docs/SPEC_PREFILL.md
weicj: Agreed. let me move the hybrid CUDA/HIP placement instructions into a dedicated docs/MIXED_BACKEND.md and leave only a short pointer from SPEC_PREFILL.md. That should make it easier to check and follow. :)

dflash/test/test_dflash.cpp
weicj: For this PR, I kept the change focused on validating the CUDA/HIP mixed backend path at the existing bench harness boundary. Pulling the draft interface refactor into the same patch would be a larger cleanup and would make the validation surface larger. But yes it is a good follow-up direction.

howard0su: suggest do the refectoring first then add HIP mixed backend support.

PR #124, opened 2026-05-08
fix(dflash): expose qwen reasoning in streaming and non-streaming response

jkyamog:
First off thank you for making this project.  I have been using for the past week, in particular on open-webui.  I noticed the opening <think> tag is missing.  I did ask codex and reviewed by claude to implement a fix.  I guess the streaming implementation is a bit more complex that its a state machine.

Here is the summary of the implementation:

> Streaming uses a sliding-window state machine (consume_stream_piece) to detect tag boundaries across token boundaries.  Respects chat_template_kwargs.enable_thinking — when disabled, tokens stream as plain content without tag parsing.

### DIFF
--- a/dflash/scripts/server.py
+++ b/dflash/scripts/server.py
@@ -16,6 +16,7 @@ Streams tokens as Server-Sent Events using the OpenAI delta format.
 import argparse
 import json
 import os
+import re
 import struct
 import subprocess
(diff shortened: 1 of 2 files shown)


davide221:
@jkyamog thanks for your first contribution 🚀

PR #125, opened 2026-05-08
feat: disable Qwen3 thinking by default + NIAH bench QoL flags

phazei:
Three small, related changes around Qwen3.x thinking-mode handling and the NIAH harness.

server.py: default `enable_thinking=False`

The OpenAI-compatible endpoint applied the tokenizer chat template with no `enable_thinking` kwarg, which falls through to the Qwen3.x template default of `True`. With thinking on, every reply burns its `n_gen` budget on a chain-of-thought rollout that DFlash's drafter cannot predict, collapsing acceptance rate (typically ~60% -> single digits) and frequently exhausting `max_tokens` before the answer block starts.

The default flips to `False`. Clients keep full control by sending `"chat_template_kwargs": {"enable_thinking": true}` per request, which the existing template-kwargs passthrough merges in. This repo only serves Qwen3.5/3.6, so the default is safe; the kwarg is silently ignored by tokenizers whose templates do not reference it.

bench_niah_cpp.py: `--no-thinking` flag

The NIAH bench builds its own chat prompt (independent of the server path) and hit the same problem at 117K context: Qwen3.6 spent the full generation budget thinking and never reached the answer. Adds a `--no-thinking` flag that injects `enable_thinking=False` into the target's `apply_chat_template` call. Default is off, so existing invocations are unchanged.

niah_gen.py: `--seed-base` flag

Replaces the hardcoded `seed=42 + i` with a `--seed-base` argument (default 42). Lets you generate independent case sets without changing `--n`, e.g. for held-out validation passes.

### DIFF
--- a/dflash/scripts/server.py
+++ b/dflash/scripts/server.py
@@ -277,8 +277,13 @@ def build_app(target: Path, draft: Path, bin_path: Path, budget: int, max_ctx: i
 
         ``template_kwargs`` is passed through to ``apply_chat_template`` so callers
         can toggle template knobs like ``enable_thinking`` per-request.
+
+        Thinking is disabled by default (enable_thinking=False) because Qwen3.6's
+        think mode wrecks DFlash acceptance rates. Clients can opt in by sending
+        ``"chat_template_kwargs": {"enable_thinking": true}`` in the request.
         """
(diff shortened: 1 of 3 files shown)


davide221:
@phazei thanks for your first contribution! Lgtm

PR #127, opened 2026-05-08
feat(dflash): --prefill-skip-park for 32 GB+ GPUs

phazei:
The PFlash compress flow parks the target + draft to VRAM-backed scratch before loading the drafter, then restores them after scoring. The park/restore protects 24 GB cards (target ~16 GB Q4_K_M + draft ~3 GB + drafter KV ~1.3 GB + BSA scratch ~600 MB exceeds 24 GB without it) but adds like 9 s of fixed cost per compress call.

On a 32 GB card (RTX 5090) all three models fit co-resident, so the park/restore round trip is pure overhead.

Plumbs a `skip_park` flag from CLI to daemon:

- `--prefill-skip-park` argparse flag (default off, behaviour unchanged for existing users).
- `PrefillConfig.skip_park` carries the choice through both server.py and server_tools.py compress paths.
- `compress_text_via_daemon` skips the `park target / park draft` and `free drafter / unpark target / unpark draft` round-trips when `skip_park=True`.
- Server start sets `DFLASH_COMPRESS_NO_PARK=1` in the daemon env when the flag is on; daemon `compress` handler then leaves the scorer resident, avoiding the ~2 s reload on every subsequent request.

The flag is opt-in. Users on 24 GB cards who pass it will OOM during compress; the help text calls out the 32 GB minimum.

### DIFF
--- a/dflash/scripts/_prefill_hook.py
+++ b/dflash/scripts/_prefill_hook.py
@@ -59,6 +59,7 @@ class PrefillConfig:
     keep_ratio: float                                  # 0.015..0.125
     drafter_gguf: Optional[Path]                       # drafter weights (Qwen3-0.6B BF16 GGUF)
     drafter_tokenizer_id: str                          # HF repo ID for drafter vocab
+    skip_park: bool = False                            # skip park/unpark on >=32 GB GPUs
 
     @property
     def enabled(self) -> bool:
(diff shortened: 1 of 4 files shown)


davide221:
@phazei thanks for enabling faster execution for >24 gb cards!

PR #129, opened 2026-05-08
draft: sliding window attention for Qwen3.6 draft model

howard0su:
Implement per-layer K/V truncation in the draft graph for SWA layers (Qwen3.6 pattern: layers 0..n-2 use sliding window, last layer full, while Qwen3.5 draft model use all full layers).

When --draft-swa=<window> is set and ctx_len > window, SWA layers:
- Window target_feat to last <window> positions via ggml_view_3d
- Adjust positions_k to start from the window offset
- Use eff_total_k = window + q_len for K/V reshape

Also raises the effective DRAFT_CTX_MAX when SWA is enabled so that ctx_len can exceed the window size and SWA truncation can activate.

Usage: --draft-swa=2048 or DFLASH27B_DRAFT_SWA=2048

 Benchmark (RTX 2080 Ti, Qwen3.6-27B Q4_K_M + DFlash draft)

| Context | Baseline AL | SWA=2048 AL | Δ AL | Baseline tok/s | SWA tok/s |
| :--- | :--- | :--- | :--- | :--- | :--- |
| 6K | 3.27 | 4.57 | +40% | 27.00 | 34.49 |
| 9K | 4.92 | 5.57 | +13% | 41.16 | 42.01 |
| 16K | 4.49 | 7.76 | +73% | 38.10 | 58.29 |
| 24K | 4.34 | 4.41 | +2% | 35.91 | 33.10 |

  Significant AL and throughput improvement at 6K–16K context. No regression on short prompts (HumanEval bench: identical AL at 5.47).

PR #131, opened 2026-05-08
Gemma4 support: pFlash + DFlash + chunked prefill, daemon mode, server routing

dusterbloom:
# Gemma4 support: pFlash + DFlash + chunked prefill, daemon mode, server routing

> **Update 2026-05-10 — three correctness fixes landed in this branch unblock TQ3 KV across all contexts and recover the 64K MoE drafter, lifting decode there from 13 → 36.57 tok/s.** New section below: "Update 2026-05-10". The original benchmarks remain for history.

## Summary

Brings the Gemma4 family (31B Dense, 26B-A4B MoE) to production parity with Qwen3.5: chunked batched prefill, DFlash speculative decoding, pFlash block-sparse attention via a new ggml op, and a daemon mode wired into `scripts/server.py` so the OpenAI-compatible HTTP server can serve Gemma4 with `--pflash`.

## Update 2026-05-10 — post-fix benchmarks

Three commits in this branch fix correctness bugs that were holding back the original numbers:

- `d758ed9bf` (submodule) `fix(fattn): force chunked path for TQ3 K to avoid MMA-intercept FWHT mismatch` — the MMA dequant intercept was dequantising TQ3_0 K to F16 without inverting the FWHT rotation that quantisation applies on write. The chunked / vec FA paths handle this correctly (they forward-rotate Q to match); the MMA intercept silently bypassed those hooks. Fix: force chunked for ALL TQ3 cases unless `DFLASH_TQ3_MMA` opts in. This is the fix for the "TQ3_0 + chunked prefill produces token 0" item in the original Known Limitations section.

- `7b62c07` (parent) `fix(gemma4): allocate+fill SWA mask for n_tokens==1 decode + bump llama.cpp` — the SWA mask was only allocated for `n_tokens > 1` ("batched prefill only" per the comment), so single-token decode fell back to `attn_mask` sized for the full kv_len padding. The K view was 2048 wide, the mask was 256 wide — dimensional mismatch let FA read past the populated region into uninitialised cudaMalloc bytes. Catastrophic with TQ3_0 (turned the cache noise into a `<unused94>`/`'en'` repetition loop), benign-but-wrong with Q8_0.

- `f1f811e` (parent) `fix(mtp): always provide FA mask for head_dim>=512 (any K type)` — the CUDA MMA dispatcher's `gqa_opt_applies` requires both `K->ne[1] % FATTN_KQ_STRIDE == 0` AND `mask != nullptr`. The previous predicate `(kv_is_tq3 && head_dim_fa >= 512) || needs_kv_pad` left Q8 + already-aligned KV with no mask, so the dispatcher returned `BEST_FATTN_KERNEL_NONE` and aborted at `fattn.cu:659`. Reproduced as M4 (MTP+Q8/Q8) crashing at step ~210 in the original benchmarks. Fix: always set `need_mask` when `head_dim_fa >= 512`.

#### Gemma-4-26B-A4B MoE — pFlash + DFlash with `--draft-max 4`

PR's original report had the 64K MoE cell at decode **13 tok/s, accept 1.23/16** because the implicit `--draft-max 16` over-speculated at long context. A `--draft-max ∈ {1, 2, 4, 8}` sweep on a 50K-token code prompt (concatenated HumanEval+ tasks) finds **dm=4 is the optimum** for MoE. With dm=4 the same model + same drafter at 64K runs at **36.57 tok/s — 2.8× the previous published number**. dm=4 also holds at 256K.

| Context | --draft-max | Decode tok/s | AL | VRAM |
|---------|-------------|--------------|-----|------|
| 4K (HumanEval/2) | 4 | 111.92 | 2.64 | 18.95 GB |
| 16K (long_2k Shakespeare) | 4 | 72.32 | 1.66 | 19.27 GB |
| 32K (long_2k) | 4 | 70.54 | 1.66 | 19.45 GB |
| **64K (long_code_50k)** | **4** | **36.57** | **1.79** | **19.74 GB** |
| 128K (long_code_50k) | 4 | 35.21 | 1.77 | 20.40 GB |
| **256K (long_code_50k)** | **4** | **35.30 / 36.63** | **1.79** | **21.73 GB** |

256K fits a 24 GB RTX 3090 with ~2.3 GB headroom. Decode tok/s is essentially flat from 64K to 256K (variance ±5%) because actual KV usage is held constant at 50K and only cache-allocation overhead scales — confirming the long-ctx decode is bound by per-step KV bandwidth at the **populated** position count, not at the cache-size allocation.

#### Gemma-4-31B Dense — DFlash + Q8/Q8 + pFlash, code prompt

| Context | --draft-max | Decode tok/s | AL | VRAM | Notes |
|---------|-------------|--------------|-----|------|-------|
| 4K (HumanEval/2) | 8 | 56.12 | 5.12 | 22.16 GB | code prompt |
| 4K (HumanEval/2) | 16 | 97.81 | 6.56 | 22.18 GB | matches PR's 149/16 reference within task variance |
| 4K (creative `long_open`) | 8 | 23.77 | 2.13 | 22.10 GB | drafter is OOD on creative writing — see "prompt distribution" note below |
| **64K (long_code_50k)** | 8 | **1.78** ← VRAM saturated, paging | 1.94 | 24.00 GB / 24.00 GB | **dense + drafter + Q8/Q8 + 64K is not viable on 24 GB** |
| 64K (target-only, Q8/Q8) | — | 7.96 | — | 22.60 GB | Without drafter Q8/Q8 dense at 64K fits |
| 64K (target-only, TQ3/TQ3) | — | 6.90 | — | 21.25 GB | TQ3 minimum-VRAM dense at 64K |

**For long-context production on a 24 GB GPU, MoE 26B is the model. The dense 31B can run at 64K target-only, but the dflash drafter pushes it into the swap regime and decode collapses.** The MTP drafter unblocks itself at small ctx with the head_dim=512 fix (4K + Q8/Q8 + MTP runs to 256 steps with accept_rate 0.87, decode 34.4 tok/s) but no MoE MTP head exists yet.

#### Drafter quality is a function of (drafter × target × prompt distribution)

The dflash drafter is trained on code-distribution data. On a creative-writing prompt (`long_open`: "Write a short story about a robot…") the drafter accept rate collapses to ~0.20–0.25; on a HumanEval prompt the same drafter at the same context and config runs at 0.64–0.67. **Bench numbers must specify prompt distribution.** PR #131's original 10.67/16 was on code prompts (matches our 6.56/16 dm=16 on HumanEval/2 within task variance); on creative writing the same drafter would have shown the same collapse.

### Updated Known Limitations (vs the original list)

- ~~**TQ3_0 + chunked prefill** produces token 0 (independent of pflash)~~ → **RESOLVED** by submodule commit `d758ed9bf` (force chunked for all TQ3 cases) and parent commit `7b62c07` (proper SWA mask for n_tokens==1 decode).
- ~~**DFlash spec acceptance drops at 64K** (avg 1.23/16 vs 10.67/16 at 32K)~~ → **MITIGATED**. Root cause was a combination of (a) over-speculation: `--draft-max 16` is too aggressive at long ctx for MoE, (b) the drafter's 2096-slot KV cap means it sees only the last ~4% of a 50K prompt. Switching to `--draft-max 4` recovers 36.57 tok/s at 64K and 35.30 tok/s at 256K. Drafter cache size remains an open improvement.
- **NEW**: dense 31B + drafter + Q8/Q8 + 64K saturates 24 GB and decode collapses to ~2 tok/s. Use MoE 26B for long-context on 24 GB consumer GPUs.

### How to reproduce

The bench scripts and BPE-tokenised prompts are in this branch under `.sisyphus/notes/gemma4-baseline/`:

- `prompts/{short_chat,long_open,long_2k,long_50k,long_code_50k,humaneval_2}.txt` — comma-separated token IDs (HF `google/gemma-3-27b-it` tokeniser is byte-identical to the GGUF; verified `vocab_size=262144 BOS=2 EOS=106` match).
- `run_dm_sweep.sh` — the dm sweep that gave the 64K dm=4 result above.
- `run_scaling.sh` — the 16K → 256K MoE ladder.
- `run_64k_v2.sh` — the post-fix dense 64K matrix.

Each runner emits a `SUMMARY.md` with prefill tok/s, decode tok/s, AL, and VRAM per cell; raw per-cell logs are reproducible.

A longer write-up of the debugging journey is in `.sisyphus/notes/gemma4-journey.md` for anyone wanting the narrative behind these numbers.

---

## Original benchmark section

(retained for history; see "Update 2026-05-10" above for the post-fix numbers)

### Gemma-4-31B Dense

| Context | Prefill baseline | Prefill +pflash | Speedup | Decode +pflash | Accept |
|---------|-----------------|-----------------|---------|----------------|--------|
| 4K  | 1438 tok/s | **1502 tok/s** | **+4.5%**  | 149 tok/s | 10.67/16 |
| 8K  | 1284 tok/s | **1497 tok/s** | **+16.6%** | 100 tok/s | 10.67/16 |

### Gemma-4-26B-A4B MoE — pFlash big wins at long context

| Context | Prefill baseline | Prefill +pflash | **Speedup** | Decode +pflash | Accept |
|---------|-----------------|-----------------|-------------|----------------|--------|
| 8K  | 3400 tok/s | (TBD)           | —           | 117 tok/s | 8.0/16   |
| 32K | 2530 tok/s | **3888 tok/s**  | **+53.7%**  | 133 tok/s | 10.67/16 |
| 64K | 1997 tok/s | **4028 tok/s**  | **+101.7%** | 13 tok/s* | 1.23/16* |

\* Decode at 64K shows low DFlash acceptance at the original `--draft-max 16` budget; **the post-fix numbers above show this lifts to 36.57 tok/s with `--draft-max 4`.** Prefill speedup is unchanged.

VRAM stays under 21 GB at 64K on the 24 GB RTX 3090.

### Speedup scaling

pFlash savings scale with KV-len because attention block selection skips blocks proportional to context length. The progression (4.5% → 16.6% → 53.7% → 101.7%) matches the design.

### New `GGML_OP_FLASH_ATTN_SPARSE` ggml op (submodule)

- CUDA dispatch in `ggml-cuda/fattn-sparse.cu` with S↔H transpose for ggml ↔ pFlash layout conversion
- BF16 fast path; falls back to dense `ggml_flash_attn_ext` when no kernel registered
- Q8_0 / Q4_0 K/V supported via `ggml_get_to_fp16_cuda` dequant before the sparse path
- Submodule commits: `5be140d feat(ggml): add GGML_OP_FLASH_ATTN_SPARSE ...`, `866688b feat(ggml-cuda): dequantize K/V to F16 in sparse FA path`

### pFlash + DFlash + chunked prefill on Gemma4

- `ggml_flash_attn_sparse` wired into `build_full_attn_block()` (full-attention layers only; SWA layers stay on dense FA)
- pFlash CUDA kernel registered via new `pflash_ggml_adapter`
- Type-aware dispatch: `pflash_supports = {F16, Q8_0, Q4_0}`; TQ3_0 falls back to dense
- `last_token_logits_only` ported from upstream PR #108 — saves ~1GB output tensor and ~1000x lm_head compute per non-last prefill chunk

### Major correctness fix: SWA mask coordinate frame

The host-built SWA causal mask was in absolute KV coordinates but the FA kernel reads it indexed by view position. For every prefill chunk where `kv_start > 0`, the mask was misaligned by `ring_win_start` columns → kernel saw all `-inf` for SWA layers → softmax NaN.

- Q8/F16 KV: degraded but plausible-looking output (NaN absorbed by saturating arithmetic). Visible: 8K Q8 baseline output token changed from `236770` (broken) to `236799` (correct) after this fix.
- TQ3_0 KV: clean NaN propagation → argmax returns 0.

Fix introduces a shared `compute_swa_view()` helper used by both the graph builder and the test driver so K view + mask stay in lockstep.

### Daemon mode + server routing

- `test_gemma4_dflash --daemon` mirrors the IPC protocol of `test_dflash`: line-based stdin commands, int32 LE token stream on `--stream-fd=N`, `-1` sentinel
- `scripts/server.py` detects GGUF architecture; routes to `test_gemma4_dflash` for `gemma4`, keeps `test_dflash` for Qwen3.5
- New `--pflash` server flag passes through to the daemon
- Smoke test: 16 valid Gemma4 vocab tokens + `-1` sentinel streamed on fd=3 with 26B-A4B + 4096-token prompt

## Run the server

```
python3 dflash/scripts/server.py \
  --target /path/to/gemma-4-26B-A4B-it-UD-Q4_K_M.gguf \
  --draft  /path/to/gemma4-26b-a4b-dflash/ \
  --pflash \
  --port 8000
```

## Test plan

- [x] `test_flash_attn_sparse` (TDD: dense vs sparse @ alpha=1.0 within BF16 tolerance; alpha<1.0 liveness) — passes
- [x] `test_gemma4_dflash --pflash` at 4K/8K Q8 on 31B Dense — output matches baseline at 4K, sparse approx at 8K
- [x] Gemma-4-26B-A4B MoE pFlash + DFlash 32K, 64K — see updated numbers above
- [x] **NEW**: post-fix MoE 16K → 256K ladder + dm sweep — see updated numbers above
- [x] Daemon mode smoke test — 16 valid Gemma4 vocab tokens + `-1` sentinel streamed on fd=3
- [ ] Server end-to-end OpenAI API request against deployed Gemma4 (manual)

## Submodule

Submodule pointer is bumped on `dusterbloom/llama-cpp-turboquant-cuda` `feature/tq3-kv-cache`. The `.gitmodules` URL was updated to that fork because the upstream `Luce-Org/llama.cpp-dflash-ggml.git` doesn't have these commits. Maintainer can rewrite the URL post-merge if the commits get mirrored to a Luce-Org repo.


davide221:
@dusterbloom can you fix the merge conflict? Great contribution!!

dflash/deps/llama.cpp
howard0su, line 1: You may need to provide the patch for ggml.

dflash/scripts/quantize_gemma4_draft_q8.py
howard0su, line 1: why the existing quantize draft to q8 script doesn't work for you? Better you can enhance it to support qwen and gemma4.

dflash/src/f16_convert.cu
howard0su, line 1: I purposely remove this file couple days ago. We can use ggml convert function to do the same thing.

dflash/src/gemma4_target_loader.cpp
howard0su, line 1: we may want to build our loader in a generic way. For this PR, I think it is fine. but this should be a high priority work item to refactor this.

dflash/src/gemma4_dflash_graph.cpp
howard0su, line 1: My read is that draft model is same as qwen3.5. Did I miss anything?

dflash/test/gemma4/smoke_gemma4_draft_forward.cpp
howard0su, line 1: more tests are welcome. better we can group those gemma4 tests into tests/gemma4. we have too Many files on this folder.

howard0su (changes requested):

dusterbloom:
### Update — SWA truncation port + YaRN opt-in (98f72c1)

Just pushed `feat(gemma4): port SWA truncation to draft graph + YaRN opt-in` (commit 98f72c1).

Adapts PR #129 (howard0su/swa, qwen3) to gemma4's cached-KV draft graph layout. Also touches the same surface as #140 / #141 — happy to coordinate with those.

**What's in:**
- `gemma4_dflash_graph.cpp` — `draft_swa_trunc_enabled()` (opt-in via `DFLASH_DRAFT_SWA_TRUNC=1` or `--draft-swa-trunc`); restricts K_full / V_full views and the FA mask (with `ggml_cont`) for SWA layers when `kv_len > sliding_window + n_tokens`.
- `gemma4_dflash_graph.cpp` — `draft_rope()` wrapper around `ggml_rope_ext`; opt-in YaRN via `DFLASH_DRAFT_YARN=1` (default `n_ctx_orig=32768`, override `DFLASH_DRAFT_YARN_NCTX_ORIG`).
- `test_gemma4_dflash.cpp` — `--draft-swa-trunc` CLI flag mirror.

**Bench (RTX 3090, gemma-4-31B-Q4_K_M target + qwen3 5L draft, 50K-token prompt, n_predict=64, ctx=65536, NO_VMM=1):**

| draft cap | SWA trunc | AL | decode tok/s |
|---:|:---:|---:|---:|
| 2096 (default) | off | 1.36 | 1.31 |
| 2096 (default) | on | 1.73 | 1.68 |
| 8192 | off | 1.02 | 4.29 |
| 8192 | **on** | **1.68** | **6.96** (+65% AL, +62% decode, ~5× over default-cap baseline) |

**Caveat:** SWA truncation does not lift AL to "healthy" (~5+). Paired probe (prose-50k vs code-50k at ctx=65536) confirms the residual AL=1 cliff is **purely position-driven**, not content-distribution-driven — the qwen3 draft model itself appears to have been trained at ≤32K positions (matching the in-source comment about "32K → 64K acceptance collapse"). Real fix needs a long-context drafter; this port is a meaningful partial improvement that ships today.

YaRN is included as an opt-in lever for users who want to experiment with draft-side RoPE scaling; default-off, no behavior change.

dusterbloom:
### Update — Submodule rebase complete (`f008033`)

Two new commits on `feature/gemma4-support`:

- **`4b0c158`** *fix(gemma4): TQ3 graph-level FWHT rotation contract* — outer-repo half of the TQ3 fix; wraps `ggml_permute(Q)` in `ggml_cont` before `ggml_turbo_wht` (turbo-wht.cu's kernel writes dst with src strides, so non-contig input scatters writes).
- **`f008033`** *chore(submodule): bump dflash/deps/llama.cpp to feature/tq3-kv-cache-clean* — addresses **r3214286746** by rebasing the fork from a 4-week-old merge-base (`fae3a2807`) onto current `ggml/master` HEAD (`0b047287f`).

#### Submodule rebase summary

Old pin: `dusterbloom/llama-cpp-turboquant-cuda` `feature/tq3-kv-cache@d758ed9bf` — 12 commits ahead of `fae3a2807`, 1 of 12 patches applied cleanly to current `ggml/master`.

New pin: `dusterbloom/llama-cpp-turboquant-cuda` `feature/tq3-kv-cache-clean@daef232a6` — 11 commits on top of `0b047287f`, **4 of 11 apply cleanly**.

Commit triage during the rebase:
- **Dropped** debug probes (`45e492b13`, `3f65b59c4`)
- **Dropped** the broken TQ3 MMA dequant intercept (`580246202`) and its safety gate (`d758ed9bf`) — that intercept produced degenerate output (token loop) and is being held for a kernel-level rework
- **Squashed** WIP rotation kernel (`e2af945b9`) into the FWHT fuse commit (`fd8710abc`)
- **New atomic commits**: `cpy` TQ3→F16 dispatch, graph-level FWHT contract refactor, fused MoE kernel, and a force-chunked-for-TQ3 dispatcher fix that restores SWA-decode correctness post-rebase

#### Verification

- ✅ Submodule build clean
- ✅ FWHT primitives roundtrip self-test: `max_diff=0.000000`
- ✅ Q8/Q8 control byte-identical to pre-rebase baseline
- ✅ TQ3/TQ3 chunked produces coherent output: `next=6596 max=21.250@6596` (matches Q8 control's first prediction); 6.58 tok/s on n_predict=64 — same as the freshly-rebuilt pre-rebase baseline binary

#### Phase-1 upstream-ready patches

Two of the 11 apply cleanly to `ggml-org/llama.cpp:master` and are immediately submittable:

1. **`0006-fix-ggml-cuda-honor-view_offs-in-cpy-data-pointer.patch`** — small, generic bugfix; multi-chunk Gemma4 prefill was silently corrupting KV cache state because `ggml_cuda_cpy()` was using `src->data` raw, ignoring `view_offs`. The fix adds a `ggml_cuda_cpy_data()` helper that resolves `view_src + view_offs` correctly. Applies cleanly to current upstream HEAD.

2. **`0010-feat-ggml-cuda-fused-MoE-kernel-for-IQ2_XS-IQ3_XXS-IQ4_XS.patch`** — generic perf, uses standard upstream quant types only, no TQ3 dependency.

Full mapping at [`submodule-patches/README.md`](https://github.com/dusterbloom/lucebox-hub/blob/feature/gemma4-support/submodule-patches/README.md).

#### Other open feedback (status table)

| Comment | Status |
|---|---|
| **r3214286746** *patch for ggml* | ✅ **This update** — rebased + Phase-1 patches identified |
| **r3214287342** *quantize_gemma4_draft_q8.py refactor* | ⏳ Separate refactor, not in this PR |
| **r3214289240** *f16_convert.cu purposely removed* | ✅ Not reintroduced in the rebase |
| **r3214290950** *generic loader refactor* | ⏳ High-priority refactor backlog |
| **r3214291841** *draft model is same as qwen3.5* | ✅ **Confirmed empirically.** Internal architectural audit (`dflash/REFACTOR_DFLASH_UNIFIED.md` proposed) shows the qwen3 and gemma4 drafts are the same 5L Qwen3-style block-diffusion; ~4-5 engineer-day unification refactor planned, not in this PR. |
| **r3214292648** *group gemma4 tests* | ⏳ Process suggestion |
| cubic-dev-ai P2 bots (8 items) | All addressable in follow-ups |

dusterbloom:
### Phase-1 upstream PR opened

The first generic bugfix from the rebased fork is now an open PR upstream:

> **https://github.com/ggml-org/llama.cpp/pull/22913** — *ggml-cuda: honor view_offs in cpy data pointer*

This is patch `0006` from the rebased `feature/tq3-kv-cache-clean` branch, applied as a single commit on top of current `ggml-org/llama.cpp:master`. Self-contained, generic, no TQ3 dependency.

Once it lands upstream, we can drop `14f90bf60` from the fork and shrink the patch series further. `0010` (fused MoE kernel) is the next Phase-1 candidate — happy to send that one too once you sign off.

dflash/scripts/quantize_gemma4_draft_q8.py
dusterbloom: Addressed in 4935293. Merged `quantize_gemma4_draft_q8.py` back into `quantize_draft_q8.py` with `--arch {qwen,gemma4}` (auto-detect from `config.json`'s `model_type` when present). Q8_0 scale formula and tensor-name mapping were already identical between the two scripts, so the merge is a clean parameterisation behind two `_ARCH_DEFAULTS` dicts. Net -134 lines. Old gemma4 file deleted; no external callers needed updating.

dflash/src/gemma4_dflash_graph.cpp
dusterbloom: Thanks for looking carefully. The gemma4 dflash graph is not a rename of the qwen3 one — there are five material architectural differences:

1. **KV cache** — gemma4 splits into two functions (`build_draft_kv_prefill_graph` + `build_gemma4_draft_graph`) sharing a persistent `GemmaTargetCache`; qwen3 is stateless (recomputes context K/V every call).
2. **Logit softcapping** — gemma4 applies `cap * tanh(logits / cap)` (cap=30) and owns a tied lm_head; qwen3 borrows the target's lm_head with no cap.
3. **Embedding scaling** — gemma4 scales input embeddings by `sqrt(hidden_size)` (Gemma family trait); qwen3 does not.
4. **6 vs 5 captured target layers** — gemma4's FC input width is `6 * target_hidden`, not 5.
5. **YaRN RoPE + mask-aware SWA truncation** — gemma4 adds opt-in YaRN extrapolation and must slice the attention mask tensor when truncating SWA layers (because the mask is causal over the cache); qwen3's SWA only slices `positions_k`.

A unified parameterised builder is feasible but would require addressing all five axes plus the loader divergence, and would change the calling convention. Estimated ~3-5 engineer-days for the refactor + bilateral re-validation. Deferring to a follow-up refactor is the right call for this PR — happy to scope as a separate work item.

dflash/src/f16_convert.cu
dusterbloom: Addressed in 9ccd8270155b9fa439d2c2aa34631b9383a0ffbb. Removed `f16_convert.cu`; replaced 10 call sites (5 distinct blocks) with `copy_target_feat_bf16_to_f32()` using `ggml_cpy(ctx, bf16_view, f32_view)` — dispatched to ggml-cuda's native BF16→F32 path in `cpy.cu`. `ggml_view_2d` with byte offset handles ring-buffer slot arithmetic; `ggml_backend_graph_compute` synchronises internally so the explicit `cudaDeviceSynchronize` is no longer needed. Net -76 lines. CI will verify the full CUDA build.

dflash/src/gemma4_target_loader.cpp
dusterbloom: Acknowledged — agreed, the loader is gemma4-specific and a generic loader would be cleaner. Tracking as a follow-up work item alongside the qwen3↔gemma4 dflash graph unification (which has 5 material deviations per [r3215263575](https://github.com/Luce-Org/lucebox-hub/pull/131#discussion_r3215263575)). Both refactors estimated ~3-5 engineer-days combined; will scope a dedicated PR after this one lands.

dflash/test/gemma4/smoke_gemma4_draft_forward.cpp
dusterbloom: Addressed in 337cee06f5a967378ff9e0e22c5bfee57e0a6429. Moved 8 gemma4-specific test files into `dflash/test/gemma4/`: `test_gemma4_dflash.cpp`, `test_gemma4_kv_tq3.cpp`, `test_mtp_graph_shapes.cpp`, `test_mtp_loader.cpp`, plus 4 `smoke_*` files. Used `git mv` so history is preserved. `test_flash_attn_sparse.cpp` stayed at `dflash/test/` root since it's generic (no gemma4 references). `dflash/CMakeLists.txt` paths updated; cmake configure passes.

dusterbloom:
Closing: fully absorbed into main. All three correctness fixes shipped — TQ3 chunked path via submodule bump bd6a56a, SWA mask for n_tokens==1 via a703915, FA mask for head_dim>=512 via the same submodule bump. Debugging narrative + dm-sweep + prompt-distribution finding shipped as db3ce62 'docs(gemma4): debugging journey blog'. Nothing left to merge.

PR #62, opened 2026-04-29
dflash: split target/draft StepGraphs to fix ggml_gallocr realloc per spec-decode step (issue #55)

dusterbloom:
Fixes #55.

## Root cause

Every spec-decode iteration calls `build_target_step_tree` (target verify, ~3127 ggml graph nodes) at `dflash/test/test_dflash.cpp:1703` and `build_draft_step` (draft forward, ~186 nodes) at `dflash/test/test_dflash.cpp:1556` on the **same** `StepGraph sg`, sharing one `ggml_gallocr`. `ggml_gallocr_needs_realloc` compares `galloc->n_nodes` to `graph->n_nodes`, so every call sees a mismatch left over from the previous call's opposite topology — forcing `ggml_gallocr_reserve` to re-walk the entire graph (CPU cost) and often `cudaFree`+`cudaMalloc` the activation buffer (GPU driver cost).

Reporter on Windows/RTX 4090/CUDA 13 sees `ggml_gallocr_needs_realloc: graph has different number of nodes` log spam every step (the message is `#ifndef NDEBUG` gated, so Linux Release builds are silent but pay the same cost). Decode tok/s halves from 90 @ 16k context to 55 @ 32k context.

## Fix

Split the shared `StepGraph sg` into `target_sg` and `draft_sg`, each with its own `ggml_gallocr`. Target verify settles into the 3127-node topology, draft into 186-node, neither bounces.

```
 dflash/test/test_dflash.cpp | 42 ++++++++++++++++++++++++++++--------------
```

The diff is minimized via a `StepGraph & sg = target_sg;` alias so the existing prefill/target-verify call sites are unchanged; only the draft block (~10 references) swaps `sg.X` for `draft_sg.X`. Daemon-mode reset and the two migrate-cache sites destroy both StepGraphs.

## Verification

Patched `ggml_gallocr_alloc_graph` to unconditionally `fprintf` to stderr at each "needs_realloc returns true" site (removing the `#ifndef NDEBUG` gate). Ran `test_dflash` on a tokenized HE prompt + n_gen=256 + --ddtree-budget=22 + --max-ctx=2048 + --fast-rollback. **Same prompt, same flags, before vs after this commit:**

| | Before | After |
|---|---:|---:|
| `needs_realloc` events over 26 steps | 56 | **3** (initial only) |
| `cudaFree+cudaMalloc` events during decode | 14 | **0** |

Reasons breakdown before fix:
- 26: `n_nodes 186 → 3127`
[...]

PR #39, opened 2026-04-27
feat(dflash): MoE 35B-A3B support + DDTree CUDA graph reuse

dusterbloom:
## Summary

Adds Qwen3.5/3.6 35B-A3B MoE target support to the dflash spec-decode path, plus performance work on the DDTree verify graph and MoE AR.

### MoE 35B-A3B (5 cycles)
- \`feat(dflash): accept qwen35moe arch in GGUF loader (MoE cycle 1)\`
- \`feat(dflash): implement MoE FFN with expert routing + shared expert (cycle 2)\`
- \`feat(dflash): parameterize target graph for 40-layer MoE + full forward test (cycle 3)\`
- \`feat(dflash): parameterize draft model for 35B-A3B MoE + YaRN RoPE (cycle 4)\`
- \`refactor(dflash): parameterize test code for multi-model support (cycle 5)\`

### Perf / fixes
- \`fix(dflash): reshape shared expert tensors for batched MoE FFN\`
- \`perf(dflash): optimize MoE compute and graph allocator reuse\`
- \`fix(dflash): correct YaRN attention scale for MoE draft model\`
- \`perf(dflash): CUDA graph reuse + GPU argmax for MoE AR (64→144 tok/s)\`
- \`perf(dflash): CUDA graph reuse for DDTree target verify\`

## Merge with main

This branch was synced with \`origin/main\` (32 upstream commits, including layer-segmented prefill + sliding-window FA, Blackwell/NVFP4 megakernel, and \`--fa-window\` CLI). Conflicts in \`internal.h\`, \`qwen35_target_graph.cpp\`, and \`test_dflash.cpp\` were resolved by:
- keeping upstream's prefill_only / windowed FA cache + \`build_qwen35_layer\` helper
- preserving the user-side MoE FFN entry point and reusable DDTree graph path
- making the DDTree ancestor mask window-aware (uses \`tree_win_start\` consistent with \`g_fa_window\`)

Verified end-to-end on Qwen3.6-27B-Q4_K_XL + dflash-3.6 drafter:
- \`./build/test_dflash\` builds clean, smoke binaries link
- merge_sort prompt at temp=0 reproduces identical token output and per-step accept stats vs the pre-merge tip (400/912 = 43.9% accept, 7.02 tokens committed/step)
- tps within measurement noise of pre-merge baseline (~85-90 tps on RTX 3090, ddtree budget=22)

## Side note: comparison with the buun-llama-cpp fork

[...]

PR #135, opened 2026-05-09
feat(dflash): native multi-request scheduler with batched target step

javierpazo:
## Summary

Brings concurrent multi-request execution to `test_dflash` on a
single GPU. Internally one cohesive unit; **happy to split into
four sequential PRs (A / B / C / D below) if you prefer per
CONTRIBUTING's "one concern per PR"** — let me know and I'll
re-open as a chain. I kept it bundled because the four pieces
share the same hunks of `test_dflash.cpp` (~+2130 lines) and
splitting cleanly would require careful hunk surgery; doing it on
request is fine.

### A. Multi `TargetCache` slots
- CLI: `--target-cache-slots=N` (alias `--cache-slots=N`)
- prefix `SLOT <id>` routes commands to a specific slot
- `DaemonSlotState` + RAII `ActiveDaemonSlot` for safe switching
- `LIST_TARGET_CACHE_SLOTS` for introspection
- all slots share target/draft weights; only KV / SSM / scratch is
  per-slot
- `create_target_cache` gains an `n_seqs` parameter so a single
  cache can be allocated batched up front

### B. Tagged stream protocol (opt-in)
- `--stream-tagged` emits frames `[-2, request_id, token]` instead
  of bare int32 tokens; sentinels `-4` (CONTINUE), `-1` (DONE)
- parser recognises `REQ <id>` / `REQUEST <id>` headers
- legacy bare-int32 streaming is unchanged when the flag is off
- lets a client demux multiple concurrent requests over the same
  stdout

### C. Native quantum scheduler
- dispatch table for `REQ`/`SLOT`/`START`, `SCHED_STEP`,
  `SCHED_DRAIN`, `LIST_REQUESTS`
- cursor-based fair round-robin between admitted requests
- non-blocking reader thread admits new requests during a drain
- `PendingQuantum{slot, req, epoch, n_gen}` carries the unit of
  work
- `CONTINUE` / `CONT` resumes a slot without re-prefilling
- `REQ <id> CANCEL` invalidates a request and bumps the slot
  epoch so a stale `CONTINUE` is rejected; `RESTORE_CHAIN` and
  legacy `generate` refuse to overwrite a slot that is owned by
  an active scheduler request

### D. Fused batched target step (CUDA path)
- new commands: `SCHED_BATCH_PEEK`, `SCHED_BATCH_PROBE`,
  `SCHED_BATCH_TARGET_TAIL`, `SCHED_BATCH_TARGET_STEP`,
  `SCHED_BATCH_DRAIN`
- `QwenGraphInputs` gains `n_seqs`; `build_delta_net_block`
  accepts `n_seqs > 1`
- `target_feat` is allocated as `[5*hidden, target_feat_cap,
  n_seqs]` when batched and the chain forwards capture features
  per-seq
- rollback for partially accepted draft tokens, multi-token verify
  and parent-id propagation in the batched path are noted as
  follow-ups; today the batched step accepts the cleanest case
  and falls back to single-seq when needed

## Validation

Per CONTRIBUTING ("benchmark before and after on the same hardware,
same warmup"). Single GPU1 RTX 6000 Ada (sm_89), Heretic Q4_K_M
target, Q8 GGUF or FP16 safetensors drafter, `FA_WINDOW=0`, KV
`q4_0/q4_0`:

| Scenario | Result |
|---|---|
| Two concurrent requests, `REQ 4 START SLOT 0 quantum=2` + `REQ 5 START SLOT 1 quantum=2`, then `SCHED_DRAIN` | closes both clean; **slot 0 = 18.41 tok/s, slot 1 = 22.50 tok/s** |
| Mid-drain admission of `REQ 6` | succeeds; `CONTINUE` on slot 0 resumes without re-prefill |
| `batch_probe_compare_ok` over a 2-seq probe | **mismatches = 0** vs the single-seq path |
| `batch_tail_commit` (2 completed pending quanta) | **29.26 ms** |
| `batch_step_commit` followed by `SCHED_DRAIN` | **29.57 ms**, then reverts cleanly back to the DFlash single-seq path |

Methodology: warmup of 1 request before measurement; same `--budget`
and KV-quant settings across runs; nothing else competing on the GPU
during the measurement window.

## Compatibility

- All new behaviour is opt-in. Default invocation of `test_dflash`
  with no scheduler flags keeps the legacy single-request path
  byte-identical.
- Tagged stream gated behind `--stream-tagged`.
- Multi-slot gated behind `--target-cache-slots=N` (default `N=1`).
- Batched target step reached only via the `SCHED_BATCH_*` command
  family; legacy `SCHED_STEP` keeps using the single-seq path.
- Hot-loop diagnostic logs (`sync_us` / `step_debug`) are gated
  behind `DFLASH27B_TIMING_DEBUG` / `DFLASH27B_STEP_DEBUG` so the
  default path is unchanged.

## Verification vs existing community PRs

- **No prior art** in lucebox-hub for the `SCHED_BATCH_*` protocol
  or for a native C++ quantum scheduler with
  REQ/SLOT/CONTINUE/CANCEL + epoch hardening. Checked against:
  - PR #39 (CUDA graph reuse, MoE 35B-A3B + DDTree) — graph reuse
    is single-seq.
  - PR #62 (split target/draft StepGraphs to fix gallocr realloc
    per spec-decode step) — splits but stays single-seq.
- No upstream collision found for tagged stream framing or
  `--target-cache-slots`.

## Notes

- Diff size warning: this branch was extracted from a working tree
  that drifted from `main`. If a hunk fails to apply on a fresh
  rebase or you spot anything off, ping me and I'll fix on the
  spot rather than push through.
- Companion branches with smaller follow-ups (CMake sm_89 / BSA,
  `gguf_draft_loader` fallback, FP16 safetensors drafter, daemon
  scripts improvements, SWA mask wiring + contract test, PFlash
  operator notes) are sitting on
  `https://github.com/javierpazo/lucebox-hub`. Holding off on
  opening those until this one is in a known state.

---

**Javier Pazó** — [@xabicasa](https://x.com/xabicasa) — xabicasa@gmail.com

### DIFF
--- a/dflash/src/internal.h
+++ b/dflash/src/internal.h
@@ -141,22 +141,11 @@ struct TargetWeights {
     int capture_layer_ids[DFLASH27B_DRAFT_N_TARGET_LAYERS] = {1, 16, 31, 46, 61};
 };
 
-struct TargetLoadPlan {
-    int  layer_begin = 0;     // inclusive
-    int  layer_end   = -1;    // exclusive; <0 means all layers
-    bool load_output = true;  // output_norm + lm_head
-};
-
 // Load a Q4_K_M target model from a GGUF file on disk.
 // Returns false and sets last_error on failure.
 bool load_target_gguf(const std::string & path,
                       ggml_backend_t backend,
                       TargetWeights & out);
-
-bool load_target_gguf_partial(const std::string & path,
-                              ggml_backend_t backend,
-                              const TargetLoadPlan & plan,
-                              TargetWeights & out);
 
 void free_target_weights(TargetWeights & w);
 
@@ -230,12 +219,6 @@ struct TargetCache {
     ggml_type kv_k_type = GGML_TYPE_Q8_0;
     ggml_type kv_v_type = GGML_TYPE_Q8_0;
 
-    // When true, K is FWHT-rotated in the graph before writing to the
-    // standard-type cache (Q4_0/Q8_0/etc), and Q is rotated at attention
-    // time. This gives TurboQuant-style outlier spreading with fast FA
-    // kernels that work on all GPU architectures.
-    bool kv_k_rotated = false;
-
     // Full-attention KV cache: one K and one V per full-attention layer.
     // Layout: [head_dim, max_ctx, n_head_kv] f16, contiguous per layer.
     std::vector<ggml_tensor *> attn_k;   // size = n_full_attn_layers (16)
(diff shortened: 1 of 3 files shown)


davide221:
Amazing contribution @javierpazo, thank you! Can you resolve the conflics?

javierpazo:
@davide221 thanks! Rebased on top of fresh main, conflicts resolved. The two collisions were tiny (a comment in internal.h::DraftLayer::is_swa and the [cfg] log line in 	est_dflash.cpp); merged the two log lines so both the upstream draft_swa / draft_ctx_max / 	arget_split_dflash flags and this PR's 	arget_cache_slots / stream_tagged show up. The big ~900-line block from upstream's new target-split path is preserved untouched.

PR #137, opened 2026-05-09
chore(dflash): enforce sm_89 user override and keep BSA enabled

javierpazo:
chore(dflash): enforce sm_89 user override and keep BSA enabled

Two CMake-side rough edges that bit me on Windows MSVC + CUDA 12.x
on RTX 6000 Ada (sm_89, Ada-only):

  1. CUDA architectures: when no explicit override is provided,
     the previous CMakeLists could fall back to `75;86`, which
     caused silent build issues on Ada-only setups. This change
     respects DFLASH27B_USER_CUDA_ARCHITECTURES (e.g. `89`) and
     uses it consistently across the dflash and submodule
     ggml/llama.cpp consumers.

  2. BSA was sometimes silently disabled depending on detection
     order. DFLASH27B_ENABLE_BSA is now respected as an explicit
     opt-in/opt-out and a clear status line is printed at
     configure time.

Net effect: a single-arch Ada-only build with BSA enabled is
reproducible from a clean checkout. Default behaviour (no
DFLASH27B_USER_CUDA_ARCHITECTURES set, BSA on) is preserved for
existing users.

Validation:
  cmake -S dflash -B dflash/build/Release \
    -DCMAKE_BUILD_TYPE=Release \
    -DDFLASH27B_USER_CUDA_ARCHITECTURES=89 \
    -DDFLASH27B_ENABLE_BSA=ON
  cmake --build dflash/build/Release --target test_dflash --parallel 8
  -> BUILD_EXIT_CODE=0, sm_89 single-arch confirmed.

Verification vs existing community PRs:

  COMP-COMPL with #48 ("auto-detect GPU arch to prevent sm_120a on
  consumer Blackwell", open) and #91 ("expose BSA config as CLI
  flags with safety warnings", merged 2026-05-04). #48 covers
  auto-detect; #91 covers runtime CLI. This PR covers the
  build-time CMake side: respect the user's explicit
  DFLASH27B_USER_CUDA_ARCHITECTURES override and keep
  DFLASH27B_ENABLE_BSA honest. The three PRs together give
  sensible defaults per hardware tier.

### DIFF
--- a/dflash/CMakeLists.txt
+++ b/dflash/CMakeLists.txt
@@ -44,59 +44,6 @@ endif()
 # the spec_prefill demo (target_gen path uses standard quant pairs).
 option(DFLASH27B_FA_ALL_QUANTS "Compile ggml-cuda fattn kernels for all KV-quant pairs" ON)
 set(GGML_CUDA_FA_ALL_QUANTS ${DFLASH27B_FA_ALL_QUANTS} CACHE BOOL "" FORCE)
-
-# Resolve the CUDA architecture list up-front so downstream logic (notably
-# the consumer-Blackwell ggml workaround below) can inspect the actual
-# arches nvcc will compile for. The dflash27b target itself is created
-# later; its CUDA_ARCHITECTURES property is applied via
-# set_target_properties once the target exists.
-#
-# Turing (75) and Ampere (86) always; Blackwell consumer (120) and Thor
-# (110 on CUDA 13+) added when nvcc supports them. DGX Spark /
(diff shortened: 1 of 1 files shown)


javierpazo:
@cubic-dev-ai good catch — restored the SM80 guard around \	est_flashprefill_kernels\. The kernel sources are gated on \_dflash27b_min_sm GREATER_EQUAL 80\ further up (\DFLASH27B_HAVE_FLASHPREFILL\), so the test would have failed to link on legacy arches. Force-pushed; only the test target is now conditional, the rest of the CMake cleanup is unchanged.

PR #94, opened 2026-05-04
feat(dflash): support Qwen3.6-27B-DFlash draft (SWA layers) — 106 t/s on RTX 4090

Quitetall:
## Summary

Adds support for `z-lab/Qwen3.6-27B-DFlash` which uses sliding window attention layers. The 3.6 matched draft gives **57% speedup** over the mismatched 3.5 draft.

## Results (RTX 4090, Qwen3.6-27B Q4_K_M target)

| Draft | Speed | Tok/step | Acceptance |
|-------|-------|----------|-----------|
| 3.5 (mismatch) | 67.4 t/s | 3.76 | 23.5% |
| **3.6 (matched)** | **106.1 t/s** | **5.12** | **32.0%** |

Budget sweep (3.6 matched):
```
budget=6:   93.9 t/s, 4.27 tok/step
budget=10: 106.1 t/s, 5.12 tok/step  ← sweet spot
budget=14:  96.5 t/s, 4.74 tok/step
budget=18: 102.7 t/s, 5.82 tok/step
budget=22: 103.5 t/s, 5.82 tok/step
```

## What changed

**`internal.h`**: Added `is_swa` per-layer flag and `swa_window` to `DraftWeights`.

**`safetensors_draft.cpp`**: Reads sibling `config.json` for `layer_types` and `sliding_window`. Sets `is_swa=true` for `"sliding_attention"` layers. Prints diagnostic:
```
[draft] SWA layers: 4/5 (window=2048)
```

**`qwen3_dflash_graph.cpp`**: Plumbs SWA mask through `ggml_flash_attn_ext`. Currently a no-op for short contexts (window=2048 > typical DFlash ctx_len), but the flag is ready for when longer contexts are used.

## How the 3.6 draft differs from 3.5

```
3.5: layer_types = [full, full, full, full, full]
3.6: layer_types = [swa,  swa,  swa,  swa,  full]  sliding_window=2048
```

Same tensor names, same shapes, same architecture otherwise. The key difference is the weights are trained with sliding window attention patterns, giving better acceptance on the Qwen3.6 target.

## Test

```bash
# Download matched 3.6 draft
huggingface-cli download z-lab/Qwen3.6-27B-DFlash model.safetensors config.json --local-dir models/draft-36

# Run (copy config.json next to model.safetensors for SWA detection)
DFLASH27B_KV_K=q4_0 DFLASH27B_KV_V=q4_0 DFLASH27B_FA_WINDOW=0 \
  python scripts/run.py \
[...]

PR #140, opened 2026-05-09
feat(dflash): wire caller-provided SWA mask through draft graph

javierpazo:
feat(dflash): wire caller-provided SWA mask through draft graph

When the Qwen3.6 draft graph is built for a context that exceeds
the SWA window, the caller-provided sliding-window attention mask
must reach `ggml_flash_attn_ext` on the SWA layers.

Previously the mask was constructed and then nullified
post-construction in qwen3_dflash_graph, so SWA layers ran without
the intended visibility constraint at long contexts.

This change makes the wiring explicit and pinable as a contract:

  * `dflash_graph.h` — `DraftGraphInputs` gains an optional
    `attn_mask` field. Documented as caller-owned, type F16, with
    shape `[kv_len, q_len]` (or padded `[kv_pad, q_pad]`), values
    `0` for visible positions and `-inf` for masked positions.
    Two helpers added so callers do not reimplement the same logic:

      bool draft_graph_needs_swa_mask(const DraftWeights & w,
                                      int ctx_len);
      void build_draft_swa_mask(std::vector<uint16_t> & out,
                                int ctx_len, int q_len,
                                int swa_window);

    `lm_head` is normalised to a default-null member at the same
    time (small consistency fix; layout unchanged).

  * `qwen3_dflash_graph.cpp` — when `total_k > swa_window` on a SWA
    layer, the graph wires the caller-provided mask through to
    `ggml_flash_attn_ext` and stops nullifying it. Layers that are
    not SWA still ignore the mask, as before.

  * `smoke_draft_graph.cpp` and `test_vs_oracle.cpp` — small
    alignment so the existing tests can build / fill the draft SWA
    mask when `ctx_len + q_len > swa_window`. No new test scaffolding
    is added in this commit; the focused regression test lives in a
    separate PR (`test(dflash): contract test for draft SWA mask
    wiring`) so each PR keeps one concern.

Validation:

  * Built and ran `smoke_draft_graph` and `test_vs_oracle` on
    RTX 6000 Ada (sm_89), Heretic Q4_K_M target, FP16 safetensors
    drafter, FA_WINDOW=0. Both tests pass before and after; the
    behaviour at ctx_len <= swa_window is unchanged (mask not
    needed and not consumed).
  * At long context the SWA layers now respect the caller mask.

Verification vs existing community PRs:

  COMP-COMPL with PR #94 (open, "support Qwen3.6-27B-DFlash draft
  (SWA layers)", Quitetall) and PR #129 (open Draft, "sliding
  window attention for Qwen3.6 draft model", howard0su).

  * PR #94 wires SWA via masks (same family as this PR).
  * PR #129 wires SWA via per-layer K/V truncation instead.

  The interface added here (caller-mask field + helpers) is small
  enough that it can survive either approach landing first. If
  PR #94 lands first, this commit should rebase cleanly because it
  formalises the mask path #94 already needs internally; if
  PR #129 lands first, the mask path here remains useful for
  callers that prefer mask semantics. Maintainers, happy to
  coordinate ordering.

### DIFF
--- a/dflash/src/dflash_graph.h
+++ b/dflash/src/dflash_graph.h
@@ -1,6 +1,9 @@
 // Shared inputs/outputs for the DFlash draft graph builder.
 #pragma once
 
+#include <cstdint>
+#include <vector>
+
 #include "ggml.h"
 
 namespace dflash27b {
@@ -13,11 +16,15 @@ struct DraftGraphInputs {
     ggml_tensor * target_hidden_cat;// [5*hidden, ctx_len, 1] f32
     ggml_tensor * positions_q;      // [q_len] i32   values [ctx_len..ctx_len+q_len-1]
     ggml_tensor * positions_k;      // [ctx_len+q_len] i32   values [0..ctx_len+q_len-1]
+    // Optional SWA mask for long-context sliding-attention layers.
+    // Shape [kv_len, q_len] or padded [kv_pad, q_pad], type F16, values
+    // 0 for visible positions and -inf for masked positions.
+    ggml_tensor * attn_mask = nullptr;
     // Optional: if non-null, the graph projects final hidden states through
     // this LM head (shape [hidden, vocab]) and returns logits instead of
     // hidden states. Used for DFlash integration where the draft shares the
     // target's lm_head.
-    ggml_tensor * lm_head;
+    ggml_tensor * lm_head = nullptr;
 };
 
 struct DraftGraphOutputs {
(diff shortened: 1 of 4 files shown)


javierpazo:
Closing — superseded by #129 (merged). #129 implemented per-layer K/V truncation for the draft SWA layers (\--draft-swa\ flag, ebcc3bd), which is a different approach than the caller-provided mask wiring this PR proposed. The upstream choice resolves the SWA correctness issue at long context cleanly, so this PR is no longer needed.

The companion focused regression test (#141) remains valid and arguably more useful now: it pins the draft-side SWA semantics as a contract regardless of whether the implementation goes via mask or K/V truncation. I'm updating its description to reflect that.

PR #141, opened 2026-05-09
test(dflash): contract test for draft SWA mask wiring

javierpazo:
test(dflash): contract test for draft SWA mask wiring

Adds a focused regression test that pins the draft-side SWA mask
wiring as a contract, independent of which SWA implementation
eventually lands.

What it tests:

  * When `draft_graph_needs_swa_mask(weights, ctx_len)` returns
    true, the graph reads `attn_mask` from `DraftGraphInputs` and
    propagates it to `ggml_flash_attn_ext` on SWA layers.
  * When SWA is not active, or `total_k <= swa_window`, the mask
    is not consumed.
  * `build_draft_swa_mask` produces the documented shape and
    values (`0.0` for visible positions and `-inf` for masked
    positions, F16 storage).

Why a separate PR:

  * Keeps "one concern per PR" per CONTRIBUTING — the wiring is
    feature, this is test.
  * The test stays useful regardless of the approach that lands
    upstream:
      - PR #94 (mask-style) — exercises exactly the wiring this
        test pins.
      - PR #129 (per-layer K/V truncation) — the contract still
        prevents regressions on the mask code path that callers
        rely on.

Depends on the companion `feat(dflash): wire caller-provided SWA
mask through draft graph` (PR to be opened in parallel). The test
needs the helpers exported in that PR's `dflash_graph.h`.

Build registration:

  Adds the test to `dflash/CMakeLists.txt` next to `test_vs_oracle`
  with the same `EXISTS`-guard pattern, so a `cmake --build ...
  --target test_draft_swa_mask_contract` builds and runs it on a
  clean checkout. No effect on other targets.

Validation:

  * `cmake --build dflash/build/Release --target
    test_draft_swa_mask_contract` succeeds on RTX 6000 Ada (sm_89),
    Windows MSVC + CUDA 12.x.
  * All assertions green: SWA-active + long-ctx case consumes the
    mask; non-SWA / short-ctx cases do not; mask helper output
    matches the documented shape and values.
  * No regressions on `smoke_draft_graph` or `test_vs_oracle`
    (untouched here).


javierpazo:
Closing — this test depended on the caller-mask wiring proposed in #140 (now closed in favor of #129's per-layer K/V truncation, which landed via ebcc3bd). The test exercises \DraftGraphInputs.attn_mask\ and the \uild_draft_swa_mask\ helper, neither of which exists in main now that the SWA approach is K/V truncation rather than mask propagation.

A regression test for the K/V truncation path would be useful but is a different test (assert that SWA layers slice \	arget_feat\ and adjust \positions_k\ correctly when \ctx_len > draft_swa\), not a rebase of this one. If someone wants that, happy to write it as a separate PR — just open an issue or ping me.

PR #149, opened 2026-05-10
docs(pflash): clarify drafter dtype in operator notes benchmark

javierpazo:
docs(pflash): clarify drafter dtype in operator notes benchmark

Follow-up to #136 (merged). The benchmark section described the
target as "Qwen3.6-27B-Q4_K_M, FP16 drafter" while the "Drafter
selection" section above it recommends Qwen3-0.6B **BF16** for
PFlash compress. cubic-dev-ai flagged this as ambiguous and harder
to reproduce.

The real setup of the measurement was:
  - PFlash compress drafter: Qwen3-0.6B BF16 GGUF (the recommended
    one).
  - DFlash drafter on the daemon side (decode after unpark):
    FP16 safetensors on this particular run.

This commit clarifies that in the benchmark paragraph, makes it
explicit that the speedup comes from the compress path itself
(not the drafter dtype), and tells the reader they can substitute
either drafter side with the format they have on disk.

No code or kernel change. Docs only.

### DIFF
--- a/pflash/README.md
+++ b/pflash/README.md
@@ -266,7 +266,12 @@ simultaneously on a 24 GB GPU.
 
 Reproducible comparison vs Ollama native `/api/chat` on the same
 64K unique-prompt summary task, RTX 6000 Ada sm_89,
-Qwen3.6-27B-Q4_K_M, FP16 drafter, FA_WINDOW=0:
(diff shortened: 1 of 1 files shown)


howard0su:
we may want to test out if q8 or even Q4_K_M draft for pfill performance in term of token/s and quality.

davide221:
@javierpazo would you be able to test performance with q8 so that can update usage instructions? I think that with q8 performance will be basically on par with bf16

PR #151, opened 2026-05-11
Preserve exact tool-call text during prompt replay

howard0su:
The OpenAI-compatible dflash server was parsing assistant tool-call output into structured JSON and then rebuilding those turns through the chat template on later requests. That preserved the semantics of the tool call, but not the exact text the model originally emitted. Small formatting differences in the rebuilt assistant turn can change tokenization, which makes prefix and KV reuse less stable for tool-using conversations.

Fix this by keeping a small Python-side tool-memory store in the server path and using it during prompt reconstruction. The server now remembers the original assistant text for generated tool-call turns, keyed by tool-call IDs. When a later request sends those same tool calls back as structured history, the prompt tokenizer looks up the IDs and re-injects the original assistant text verbatim instead of re-rendering canonicalized tool-call objects. If a lookup is missing or inconsistent, the existing structured rendering path still applies.

This change stays intentionally simple and Python-first. It adds a focused ToolMemory helper plus regression coverage for chat-completions and responses round trips, without introducing a native radix-tree backend before profiling shows the server-side store is a real bottleneck.

### DIFF
--- a/dflash/scripts/server.py
+++ b/dflash/scripts/server.py
@@ -697,7 +697,7 @@ def build_app(target: Path, draft: Path | None, bin_path: Path, budget: int, max
         cap=prefix_cache_slots,
     )
     if prefill_cfg is not None and prefill_cache_slots > 0:
-        prefix_cache.init_full_cache(prefill_cache_slots, budget_bytes=prefill_cache_bytes)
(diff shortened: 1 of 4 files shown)


davide221:
lgtm

PR #152, opened 2026-05-11
Add Gemma 4 RTX 4090 backend helpers

adybag14-cyber:
image
image


adybag14-cyber:
RTX 4090 Gemma 4 running at 60 tk/s 31b it abliterated

davide221:
@adybag14-cyber thanks for your contribution! @dusterbloom can you take a look at this

dusterbloom:
Thanks for the contribution and the time you put into the scripts.

We can't take this one as-is. A few specific reasons:

**The Gemma 4 path in this repo is `dflash/`, not libllama.** The PR description says *"DFlash runtime in this repository is still the Qwen/Laguna research path; Gemma 4 uses libllama"* — that isn't accurate. `dflash/src/gemma4_target_graph.cpp`, `dflash/src/gemma4_mtp_graph.cpp`, `dflash/src/gemma4_dflash_graph.cpp`, and `dflash/src/gemma4_target_loader.cpp` are the active Gemma 4 implementation. They power our current Gemma 4 26B-A4B benches up to 1M context with MTP γ=2 — see `.sisyphus/notes/gemma4-baseline/mtp-gamma/`. The scripts in this PR set up a parallel path through `llama-server` and bypass that work rather than building on it.

**The submodule pin is intentional and non-negotiable.** `dflash/deps/llama.cpp` is pinned to our `feature/tq3-kv-cache-clean` branch because it carries TQ3_0 KV quantization, the graph-level FWHT contract, sparse FA, and the chunked attention path that the rest of the codebase depends on. CI builds against that pin; benchmarks reference it. We can't replace it with an arbitrary third-party llama.cpp build per script.

**The scripts hardcode paths from a workstation we don't have access to** — `/mnt/c/Users/adyba/...` and `/home/tdamre/src/llama.cpp-mtp-pr22673/build-mtp-cuda124-speed-faall/bin/llama-server`. We can't ship those, and we can't CI them. The same goes for the `--spec-draft-n-max 4` claim presented as "the measured stable MTP window for this 31B target on the RTX 4090" — no log is cited, and our own γ-sweep on this target (γ ∈ {1, 2, 4, 8} at 4K/16K/64K, RTX 3090) shows γ=2 as the winner past short contexts; γ=4 regresses. Different GPU, different KV layout, possibly different result — but it would need a log to ground the claim.

We're moving toward a declarative config layout — a `configs/backends/` registry for alternate llama.cpp builds and `configs/profiles/` for per-machine deployment configs with measurement provenance — which is the shape we'd take a contribution like this in. **Work-in-progress at #155** — same shape we'd take a re-submission in. If you're still interested once that lands, a re-submission as a small profile + backend descriptor would be the natural way back in.

PR #164, opened 2026-05-13
Update download commands from huggingface-cli to hf in the READ.ME 

nthomisee:
huggingface-cli is deprecated


davide221:
@nthomisee thanks for your first contribution!

PR #165, opened 2026-05-13
Support Chat Completions `max_completion_tokens`

jkyamog:
This PR fixes an OpenAI Chat Completions compatibility gap for clients that send `max_completion_tokens`.

dflash accepted `max_tokens`, but ignored `max_completion_tokens`. When clients requested a larger output budget using `max_completion_tokens`, dflash still used its local `max_tokens` default of 512. On long-running agent workflows this can cause repeated truncation/continuation loops even though the client asked for enough output room.

## User-Visible Failure

Hermes hit repeated length truncation during a heavier terminal/tool workflow:

```text
● yes run it
────────────────────────────────────────

  ┊ 💻 preparing terminal…
  ┊ 💻 $         cd ~/Projects/gpu-cpu-hybrid && ./scripts/rebuild_and_deploy_ik_llama.sh --local-path ~/Projects/ik_llama.cpp  60.5s [error]
⚠️  Response truncated (finish_reason='length') - model hit max output tokens
⚠️  Response truncated (finish_reason='length') - model hit max output tokens
⚠️  Response truncated (finish_reason='length') - model hit max output tokens
 ─  ⚕ Hermes  ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

     Error: Response remained truncated after 3 continuation attempts

 ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 ⚕ luce-dflash │ 55.6K/65.5K │ [████████░░] 85% │ 12h 30m │ ⏲ 6m 50s
```

The server log showed the request included a larger `max_completion_tokens`, but the effective generation cap was still 512 before this fix.

Related upstream user report:

- https://github.com/Luce-Org/lucebox-hub/issues/157#issuecomment-4436186211 reports WebUI answering with decent speed, "but only 512 tokens, then cut off."

## Root Cause

The Chat Completions request model only included:

```python
max_tokens: int = 512
```

So Pydantic ignored `max_completion_tokens` from OpenAI-compatible clients. Generation length was then computed from `req.max_tokens`, often the default 512, instead of the larger requested completion budget.

## What Changed

- `ChatRequest` now accepts `max_completion_tokens`.
- Chat generation length uses `max_completion_tokens` when present, otherwise falls back to `max_tokens`.
- Chat request logging now reports:
  - `max_tokens`
  - `max_completion_tokens`
  - effective max tokens used for generation
- Added a regression test proving `max_completion_tokens` overrides `max_tokens` when constructing the daemon command.

## Validation

Unit tests:

```bash
./.conda/bin/python -m pytest dflash/scripts/test_server.py -q
```

Live smoke test against local dflash:

```text
stream=True  msgs=15 {'system': 1, 'user': 3, 'assistant': 6, 'tool': 5}  tools=22  prompt_tokens=29323  max_tokens=512  max_completion_tokens=16384  effective_max_tokens=16384  max_ctx=65536  model=luce-dflash
```

The request used the client-provided `max_completion_tokens=16384` instead of falling back to 512.

### DIFF
--- a/dflash/scripts/server.py
+++ b/dflash/scripts/server.py
@@ -483,6 +483,7 @@ class ChatRequest(BaseModel):
     messages: list[ChatMessage]
     stream: bool = False
     max_tokens: int = 512
+    max_completion_tokens: int | None = None
     temperature: float | None = None   # 0 = greedy, >0 = sample
     seed: int | None = None             # rng seed for sampling
     top_p: float | None = None         # nucleus, applied when temperature > 0
@@ -1094,6 +1095,9 @@ def build_app(target: Path, draft: Path | None, bin_path: Path, budget: int, max
     def _gen_len_for(prompt_len: int, max_tokens: int) -> int:
         return min(max_tokens, max_ctx - prompt_len - 20)
 
+    def _max_tokens_for(req) -> int:
+        return getattr(req, "max_completion_tokens", None) or req.max_tokens
+
     # ── /v1/chat/completions ────────────────────────────────────────────────
 
     @app.post("/v1/chat/completions")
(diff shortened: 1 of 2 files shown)


davide221:
thanks for the bug fix @jkyamog 🚀

jkyamog:
> thanks for the bug fix @jkyamog 🚀

I am just trying to make my bots work on the under utilized 3090 🤣  
image

PR #177, opened 2026-05-13
fix(gemma4): add long-context KV correctness

dusterbloom:
## Validation (for review)

Upstream `llama-bench` baseline (Gemma4-31B-it Q4_K_M, q4_0 KV, FA on, RTX 3090): **pp16384 = 376.62 tok/s, tg300 = 20.39 tok/s**.

Long-context KV correctness for TQ3 KV (ring-buffer + SWA boundary). Plus: `ggml_backend_buffer_clear` replaces per-tensor 1 MiB chunked memset on cache reset (matches llama.cpp convention).

| Metric | Without PR | With PR |
|---|---|---|
| TQ3 KV decode at 16k+ | ⚠ undefined (KV drift) | ✓ correct |
| AR + TQ3+pFlash @ 16k | n/a | **739 pp / 16.6 tg tok/s** |
| AR + TQ3+pFlash @ 64k | n/a | 474 pp / 8.1 tg tok/s |
| Cache-reset I/O | seconds (chunked memset) | 1 DMA op |
| Byte-identical to upstream Q4 @ temp=0 | n/a | ✓ |

---
PR #6 of the Gemma4 split sequence. Adds the targeted test that validates long-context KV correctness for the Gemma4 target graph.

Depends on #176 → #171 → #170 → #169 → #168. Stacks on `split/gemma4-05-target-graph`; the diff visible here will include those ancestor commits until they merge.

## Scope (2 files)
- `dflash/test/gemma4/test_gemma4_kv_tq3.cpp` — new test (179 lines) covering TQ3-quantized KV, SWA ring wrap, decode mask for `n_tokens == 1`. Uses the 4-arg `create_gemma4_cache(w, max_ctx, backend, cache)` API.
- `dflash/CMakeLists.txt` — wires the `test_gemma4_kv_tq3` target (CUDA-linked, like the other gemma4 tests).

## Note on the implementation hunks listed in `pr06-kv-correctness.md`

The PR06 spec lists four implementation-side categories of changes to `gemma4_target_graph.cpp` / `internal.h`:
- SWA ring/window behavior
- target KV cache alignment + wrapping
- TQ3-compatible K/V handling
- decode mask fixes for `n_tokens == 1`

[...]

PR #176, opened 2026-05-13
feat(gemma4): add target graph execution

dusterbloom:
## Validation (for review)

Upstream `llama-bench` baseline (Gemma4-31B-it Q4_K_M, q4_0 KV, FA on, RTX 3090): **pp16384 = 376.62 tok/s, tg300 = 20.39 tok/s**.

Gemma4 target forward in pure ggml — SWA + full-attn, proportional RoPE, PLE, MoE FFN, shared KV.

| Metric | Without PR | With PR |
|---|---|---|
| AR + Q4 @ 16k prefill | n/a | **678 tok/s** (+80% vs upstream 376.62) |
| AR + Q4 @ 16k decode | n/a | **22.6 tok/s** (+11% vs upstream 20.39) |
| Byte-identical to upstream `llama-cli` AR Q4 @ temp=0 | n/a | ✓ (Q4 lossless) |

---
PR #5 of the Gemma4 split sequence. Adds the basic Gemma4 target graph + cache lifecycle. **No pFlash, no DFlash draft, no MTP** — those land in PRs #7, #9, #10 respectively.

Depends on #171 → #170 → #169 → #168. Stacks on `split/gemma4-04-api-target-loader`; the diff visible here will include those ancestor commits until they merge.

## Scope (4 files exactly per `pr05-target-graph.md`)
- `dflash/src/gemma4_target_graph.cpp` — basic graph builders (`rms_norm_mul`, `build_geglu_ffn`, `build_moe_ffn`, `build_swa_attn_block`, `build_full_attn_block`, `build_gemma4_graph`) + cache lifecycle (`create_gemma4_cache`, `free_gemma4_cache`, `reset_gemma4_cache`) + `compute_swa_view`.
- `dflash/src/internal.h` — adds `GemmaTargetCache`, `GemmaGraphInputs`, `GemmaGraphOutputs`, `SwaView` structs + 5 prototypes (no draft/MTP/pflash fields).
- `dflash/test/gemma4/smoke_gemma4_target_forward.cpp` — minimal forward smoke (uses the stripped 4-arg `create_gemma4_cache` API).
- `dflash/CMakeLists.txt` — wire the new source into `dflash27b` and add the `smoke_gemma4_target_forward` target.

## What was stripped from the source branch (relative to `feature/gemma4-support`)
- `build_full_attn_block`: removed `use_pflash` / `pflash_alpha` params and the entire `ggml_flash_attn_sparse` dispatch branch — only `ggml_flash_attn_ext` remains.
[...]

PR #171, opened 2026-05-13
feat(gemma4): add target API and GGUF loader

dusterbloom:
## Validation (for review)

Upstream `llama-bench` baseline (Gemma4-31B-it Q4_K_M, q4_0 KV, FA on, RTX 3090): **pp16384 = 376.62 tok/s, tg300 = 20.39 tok/s**.

GGUF target loader for Gemma4 (Dense 31B + MoE 26B-A4B). Foundation for every later PR.

| Metric | Without PR | With PR |
|---|---|---|
| Build clean | n/a (Gemma4 unsupported) | ✓ |
| Load Gemma4-31B Q4_K_M | ✗ | ✓ (60 layers, 5376 embd, 262144 vocab) |
| Byte-identical to upstream | n/a | enables the chain |

---
## Purpose

Introduce Gemma4 constants, public error access, and target GGUF weight loading without adding graph execution or decode behavior. This PR brings up the public header and the target model loader so that subsequent PRs can build the graph on top.

## Changed Files

- `dflash/include/gemma4.h` — public constants (`GEMMA4_DRAFT_N_TARGET_LAYERS`, etc.) and `gemma4_last_error` accessor
- `dflash/src/errors.cpp` — shared error accessor, now serves both Qwen35 and Gemma4 callers
- `dflash/src/gemma4_target_loader.cpp` — file mmap utilities, `is_gemma4_gpu_tensor`, `load_gemma4_target_gguf`, `free_gemma4_target_weights`
- `dflash/src/internal.h` — `GemmaTargetLayer`, `GemmaTargetWeights` structs; `load_gemma4_target_gguf` and `free_gemma4_target_weights` prototypes
- `dflash/test/gemma4/smoke_load_gemma4_target.cpp` — smoke test: open a Gemma4 GGUF and verify metadata loads without crashing
- `dflash/CMakeLists.txt` — adds `src/gemma4_target_loader.cpp` to library sources and `smoke_load_gemma4_target` test target

## Excluded (future PRs)

- `load_gemma4_mtp_assistant`, `free_gemma4_mtp_assistant`, `get_mtp_swa_pattern`, `resolve_mtp_donor_layers`, `MtpDrafterWeights` — MTP assistant loader (PR10)
- `GemmaTargetCache`, cache creation/free/reset functions, graph builder — target graph/cache (PR05)
- Draft loader (`GemmaDraftWeights`, `load_gemma4_draft_*`) — PR08

## Dependency Chain

[...]

PR #169, opened 2026-05-13
chore(deps): update llama.cpp for Gemma4 substrate

dusterbloom:
Second PR in the Gemma4 split sequence. Substrate bump only — moves the `dflash/deps/llama.cpp` submodule to the branch carrying the GGML/TQ3/sparse-FA changes required by every Gemma4 PR after this one. **No parent-repo source changes.**

> Depends on #168 (PR 01 — repo hygiene). Stacks on `split/gemma4-01-repo-hygiene`; until #168 merges, the diff here will include the hygiene commit. After #168 merges this PR will rebase onto main cleanly.

## Changes
- `.gitmodules` — point `dflash/deps/llama.cpp` at `https://github.com/dusterbloom/llama-cpp-turboquant-cuda.git`, branch `feature/tq3-kv-cache-clean`.
- `dflash/deps/llama.cpp` submodule pointer → `ecb832bbea1a489c08e53800b50910c420bb33b6`.

## Risk: HIGH
Substrate bump affects every downstream Gemma4 PR. Submodule branch must be reviewed independently before approving the parent pointer here.

## Review checklist
- Review the submodule branch (`feature/tq3-kv-cache-clean` at `ecb832bb`) separately.
- Confirm parent repo has no unreviewed local edits to the submodule.
- `git -C dflash/deps/llama.cpp status --short` is clean.
- `git diff --check` passes.

## Validation
- Submodule status clean.
- CUDA build of GGML targets from the submodule branch (if a CUDA host is available).

## Plan reference
- `dflash/docs/gemma4_pr_split_plan.md` (master)
- `dflash/docs/gemma4-pr-split/pr02-llama-substrate.md` (this PR)

PR #179, opened 2026-05-13
feat(gemma4): route target prefill through pflash

dusterbloom:
## Validation (for review)

Upstream `llama-bench` baseline (Gemma4-31B-it Q4_K_M, q4_0 KV, FA on, RTX 3090): **pp16384 = 376.62 tok/s, tg300 = 20.39 tok/s**.

Routes target prefill through `ggml_flash_attn_sparse` (pFlash). F5 v1 (`a292818`) + F5 v2 (`57c46ca`) gate pflash to decode-only on full-attention layers — without the gate, prefill aborts at `fattn.cu:576` (sparse FA has no mask, dispatcher returns NONE).

| Metric | Without PR | With PR |
|---|---|---|
| AR + TQ3+pFlash @ 16k | ⚠ aborts pre-F5 v1 | **739 pp / 16.6 tg** |
| MTP + TQ3+pFlash @ 16k | ⚠ aborts pre-F5 v2 | **740 pp / 17.0 tg** (γ=1) |
| Prefill vs upstream Q4 (pp16384) | n/a | **+96%** (376 → 739) |
| Byte-identical first 2 tokens vs Q8 | n/a | ✓ (PR #10 validation reproduced) |

---
PR #7 of the Gemma4 split sequence. Wires the sparse-FA adapter (#170) into Gemma4 target prefill. **Performance/dispatch only** — no draft, no MTP, no server routing.

Depends on #177 → #176 → #171 → #170 → #169 → #168. Stacks on \`split/gemma4-06-kv-correctness\`; diff will include ancestor commits until they merge.

## Scope (2 files, +40 / -3)
- \`dflash/src/internal.h\` — adds two fields to \`GemmaGraphInputs\`: \`use_pflash\` (default \`false\`) and \`pflash_alpha\` (default \`0.12f\`).
- \`dflash/src/gemma4_target_graph.cpp\` — re-adds the pFlash dispatch inside \`build_full_attn_block\`:
  - signature gains \`bool use_pflash, float pflash_alpha\` (the existing 18-arg call site is the only caller)
  - inside the function: KV-type support check (\`F16\` / \`Q8_0\` / \`Q4_0\` unconditionally; \`TQ3_0\` only when \`DFLASH_PFLASH_TQ3=1\` is set in env, since the TQ3 path falls back to chunked SGEMM otherwise)
  - one \`if (can_pflash) attn = ggml_flash_attn_sparse(...) else attn = ggml_flash_attn_ext(...)\` branch — the fallback path is byte-identical to the previous unconditional call
  - \`build_gemma4_graph\` passes \`in.use_pflash, in.pflash_alpha\` through to the call site

## Gating logic

\`\`\`cpp
const bool can_pflash = use_pflash
                     && pflash_supports(Kfa->type)
                     && pflash_supports(Vfa->type);
\`\`\`

- Default \`use_pflash = false\` keeps current behavior bit-for-bit when callers don't opt in.
- Unsupported K/V dtypes silently fall back to \`ggml_flash_attn_ext\`. No assertion-fail path.
- \`DFLASH_PFLASH_TQ3\` env gate lets us A/B TQ3 against the slower chunked SGEMM fallback at runtime.

## CMake / build defs
No new CMake hunks. The \`GGML_OP_FLASH_ATTN_SPARSE\` op + \`DFLASH27B_HAVE_CUDA_WMMA_FLASHPREFILL\` compile def landed with PR #170 (sparse-FA adapter). The op exists in the submodule (PR #169).

## Test driver hunks: deferred
The spec mentions \`test_gemma4_dflash.cpp\` for \`--pflash\` coverage. That test file in \`feature/gemma4-support\` is 2962 lines, entangled with draft / MTP / daemon scaffolding the later PRs introduce. Extracting a pflash-only subset would either pull in symbols not yet declared (PR #8, #9) or duplicate the test file. Decision: defer the test-driver integration to the PR that introduces the driver (PR #9 dflash-runtime adds the rest of it), and rely on PR #170's \`test_flash_attn_sparse\` for op-level coverage in this PR. Documented for reviewer.

## Risk: HIGH
First runtime path through the sparse-FA op in the Gemma4 target. Default-off behavior is identical to PR05/06; risk lives entirely behind \`use_pflash = true\` callers, which don't exist yet in this stack (lands with PR #9).

## Review checklist (from spec)
- pFlash gated by supported KV types and backend capability ✓ (\`pflash_supports\` lambda)
- Fallback path unchanged when pFlash disabled ✓ (\`else\` branch is the prior unconditional code)
- Buffer-null guards included with the pFlash code ✓ (no null-deref reachable — \`Kfa\`/\`Vfa\` come from \`ggml_view_3d\` which never returns null on success)
- Test-driver additions only expose/validate pFlash — deferred (see above)

## Validation
- \`g++ -fsyntax-only gemma4_target_graph.cpp\` via project's compile_commands flags → **exit 0**.
- Full \`cmake --build\` of \`test_flash_attn_sparse\` + \`test_gemma4_dflash\` deferred to CI (no CUDA build budget on the dev host this turn).
- Op-level coverage: PR #170's \`test_flash_attn_sparse\` exercises \`ggml_flash_attn_sparse\` directly. End-to-end \`--pflash\` coverage lands with PR #9.


dusterbloom:
Superseded by #193 — re-cut against current main after #175 brought the `gemma4/` skeleton in. The split chain was overtaken by upstream work (PR #175 + #178 + #170 all merged during the rebase window), so re-opening as a single cohesive replacement PR is the cleaner path.

The content of this PR is included in #193. Closing here to consolidate review surface for mrciffa.

PR #180, opened 2026-05-13
feat(gemma4): add draft loader and quantization support

dusterbloom:
PR #8 of the Gemma4 split sequence. Adds Gemma4 DFlash draft weight loading (safetensors + GGUF) and the Q8_0 quantization tool that produces the draft GGUF. **No graph execution, no draft KV cache, no speculative decode runtime** — those land in PR #9.

Depends on #179 → #177 → #176 → #171 → #170 → #169 → #168. Stacks on \`split/gemma4-07-pflash-integration\`; diff includes ancestor commits until they merge.

## Scope (5 files)
- \`dflash/src/gemma4_dflash_graph.cpp\` — **loaders + free function only**. Restored from \`feature/gemma4-support\` and stripped down: 1112 → 803 lines (309 graph-execution lines removed). Keeps:
  - safetensors header parser + tensor mmap helpers (anonymous-namespace block)
  - \`load_gemma4_draft_safetensors(dir, backend, out)\`
  - \`load_gemma4_draft_gguf(path, backend, out)\`
  - \`free_gemma4_draft_weights(w)\`
  Removed: \`build_draft_kv_prefill_graph\`, \`build_gemma4_draft_graph\`, \`draft_rope\`, \`draft_swa_trunc_enabled\` (all graph-execution code, deferred to PR #9).
- \`dflash/src/internal.h\` — adds \`GemmaDraftLayer\` + \`GemmaDraftWeights\` structs and the three loader/free prototypes.
- \`dflash/scripts/quantize_draft_q8.py\` — adds \`--arch gemma4\` mode alongside existing \`--arch qwen\`. Auto-detects from config.json when \`--arch\` omitted. Qwen Q8_0 path preserved.
- \`dflash/test/gemma4/smoke_load_gemma4_draft.cpp\` — minimal smoke that calls \`load_gemma4_draft_safetensors\` → prints weight shapes → \`free_gemma4_draft_weights\`.
- \`dflash/CMakeLists.txt\` — adds \`src/gemma4_dflash_graph.cpp\` to \`dflash27b\` and the \`smoke_load_gemma4_draft\` test target.

## Risk: Medium
Loader code is pure mmap/parse, no GPU execution. Risk is metadata / tensor-name mismatches against future Gemma4 draft GGUFs, which the smoke test catches end-to-end.

## Review checklist (from spec)
- Loader ownership and cleanup are clear ✓ (\`load_*\` allocates ctx+buf, \`free_*\` releases both; smoke test exercises the lifecycle)
[...]

Issue #10, opened 2026-04-22
Oversized `--max-ctx` silently destroys attention throughput (20× prefill slowdown at 32K, 8× at 64K)


dusterbloom:
## Symptom

When running `test_dflash` with `--max-ctx` set larger than the actual required capacity (`prompt_len + n_gen + verify_block`), attention throughput collapses. The slowdown scales roughly with `max_ctx / kv_len` and is especially brutal with `--kv-q4`.

The README example `"128K context in 24 GB (134.78 tok/s at ctx=131072)"` and `scripts/run.py`'s help text `"Q4_0 KV cache (required for max_ctx=131072)"` nudge users toward `--max-ctx=131072`, which turns into a performance trap whenever the real prompt is smaller.

## Repro (RTX 3090, Qwen3.5-27B Q4_K_M target + BF16 draft, pristine `main` at 962c54a)

```bash
# same binary, same prompt, only --max-ctx differs:
./build/test_dflash models/Qwen3.5-27B-Q4_K_M.gguf models/draft/model.safetensors \
    mid_prompt_32k.bin 256 /tmp/out.bin \
    --fast-rollback --ddtree --ddtree-budget=22 --max-ctx=32768   # tight

DFLASH27B_KV_Q4=1 ./build/test_dflash ... --max-ctx=131072        # oversized
```

### Measurements

| Prompt | KV | `--max-ctx` | Prefill | Gen tok/s | Accept-length |
|---|---|---:|---:|---:|---:|
| 32K | Q8_0 (default) | 32768 (tight) | **38 s** | **42.0** | 4.41 |
| 32K | Q4_0 (`--kv-q4`) | 131072 (oversize) | **1035 s** | **1.10** | 4.13 |
| 64K | Q4_0 (`--kv-q4`) | 131072 (oversize) | 1014 s | 3.97 | 4.06 |
| 64K | Q4_0 (`--kv-q4`) | 66048 (tight) | **126 s** | **18.35** | 4.20 |

So at 32K the same prompt runs **27× slower in prefill and 38× slower in decode** purely because of \`--max-ctx\`. At 64K it's **8.1× / 4.6×**.

Output quality is unaffected (\`last_tok\` bit-exact across both paths), so this is strictly a perf bug, not correctness.

## Root cause

[...]

PR #181, opened 2026-05-13
feat(gemma4): add dflash draft runtime

dusterbloom:
PR #9 of the Gemma4 split sequence. Adds Gemma4 DFlash draft graph execution, draft KV cache lifecycle, and a smoke test that exercises the full draft forward path (loader → KV cache alloc → prefill → forward). **No MTP, no γ>1 speculative decode, no server routing** — those land in PR #10/#11/#13.

Depends on #180 → #179 → #177 → #176 → #171 → #170 → #169 → #168. Stacks on \`split/gemma4-08-draft-loader-quant\`; diff includes ancestor commits until they merge.

## Scope (5 files, +761 / -8)
- \`dflash/src/gemma4_dflash_graph.cpp\` — restores the graph-builder block stripped in #180: two helpers (\`draft_swa_trunc_enabled\`, \`draft_rope\` with optional YaRN), \`build_draft_kv_prefill_graph\` (prefix-direct KV materialisation, SGLang style), \`build_gemma4_draft_graph\` (KV-cached forward returning softcapped logits).
- \`dflash/src/gemma4_target_graph.cpp\` — adds \`create_draft_kv_cache\` + \`free_draft_kv_cache\` (these live on \`GemmaTargetCache\` because the draft KV is per-target-cache). Wires \`free_draft_kv_cache(c)\` into \`free_gemma4_cache\` and \`c.draft_kv_pos = 0\` into \`reset_gemma4_cache\`.
- \`dflash/src/internal.h\` — adds the \`draft_kv_*\` fields to \`GemmaTargetCache\` (ctx, buf, K/V vectors, cap, pos) and prototypes for the four new functions.
- \`dflash/test/gemma4/smoke_gemma4_draft_forward.cpp\` — end-to-end smoke: loads target + draft, allocates draft KV, runs \`build_draft_kv_prefill_graph\`, runs \`build_gemma4_draft_graph\`, asserts non-null logits.
- \`dflash/CMakeLists.txt\` — wires the new smoke target.

## Risk: HIGH
First runtime path that actually executes the DFlash draft. Risk lives in:
- KV cache bounds: \`create_draft_kv_cache\` reserves \`sliding_window + block_size + 32\` slots; \`build_draft_kv_prefill_graph\` aborts on overflow.
- Draft RoPE: optional YaRN scaling via env (\`DFLASH_DRAFT_YARN\`), default off.
- SWA truncation: optional via env (\`DFLASH_DRAFT_SWA_TRUNC\`), default off — preserves the pre-PR#129 behavior unless opted in.

## Deferred — \`test_gemma4_dflash.cpp\` integration hunks
The full test driver in \`feature/gemma4-support\` is **2962 lines with 71 MTP references intermixed**. The \`--draft-method dflash\` flag and dflash-specific code paths cannot be cleanly carved out without (a) duplicating large chunks of the driver scaffolding into a PR-specific copy, or (b) pulling in MTP symbols this PR explicitly excludes. The full driver lands with PR #11 (MTP runtime), which already needs both \`--draft-method dflash\` and \`--draft-method mtp\` paths together.

For PR #9, the dedicated \`smoke_gemma4_draft_forward\` covers the same code paths end-to-end (loader → KV cache → prefill graph → forward graph → logits). The test driver integration is documented to maintainers and tracked.

## Forbidden-symbol audit (vs spec exclusion list)
\`\`\`
grep -cw 'mtp\|MTP'                                              → 0  (all 4 files)
grep -nE 'h_prev|MtpDrafter|MtpLayerWeights|build_mtp_step_graph'  → no matches
\`\`\`

## Review checklist (from spec)
- DFlash draft runtime works without MTP code ✓ (compiles independently — verified via syntax-only build of both .cpp files)
- Draft KV cache wrap behavior covered or deferred — **deferred**: prefix-direct prefill aborts on overflow rather than wrapping; the wrap-or-overflow policy is a follow-up decision once decode-loop logic lands. Smoke exercises happy-path only.
- Test-driver flags introduced here are only DFlash — **deferred to PR #11** per above.
- No server behavior is changed ✓.

## Validation
- \`cmake\` reconfigure picks up the new smoke target ✓
- \`g++ -fsyntax-only gemma4_dflash_graph.cpp\` via project's compile_commands flags → **exit 0**
- \`g++ -fsyntax-only gemma4_target_graph.cpp\` → **exit 0**
- Full \`cmake --build smoke_gemma4_draft_forward\` deferred to CI (no local CUDA build budget this turn)
- Runtime invocation requires both a target GGUF and a draft directory on disk.


dusterbloom:
Closing: targets obsolete dflash/src/gemma4_*.cpp paths. Gemma4 moved to vendored deps/llama.cpp/src/models/gemma4-iswa.cpp; spec-decode integration is now via the common/ interface layer.

PR #182, opened 2026-05-13
feat(gemma4): add mtp loader and step graph

dusterbloom:
## Validation (for review)

Upstream `llama-bench` baseline (Gemma4-31B-it Q4_K_M, q4_0 KV, FA on, RTX 3090): **pp16384 = 376.62 tok/s, tg300 = 20.39 tok/s**.

MTP assistant loader + per-step cross-attention graph. Plus: deletes per-token-per-layer `[mtp-fa-types]` stdout flood and a redundant ggml_cont on the non-TQ3 path.

| Metric | Without PR | With PR |
|---|---|---|
| MTP graph builds | ✗ | ✓ (4 MTP layers, 322 MiB GPU) |
| Dense MTP @ 16k decode | n/a | **21.4 tok/s** (γ=1; γ=2 follow-up) |
| Stdout pollution per MTP token | 4-8 lines × n_layer | 0 |
| Byte-identical at γ=1 vs AR (non-accepted tokens) | n/a | ✓ |

---
PR #10 of the Gemma4 split sequence. **Final non-daemon PR.** Adds Gemma4 MTP (Multi-Token Prediction) assistant loading and one-step graph construction. **Not wired into the decode loop** — that integration lands with deferred PR #11.

Depends on #181 → #180 → #179 → #177 → #176 → #171 → #170 → #169 → #168. Stacks on \`split/gemma4-09-dflash-draft-runtime\`; diff includes ancestor commits until they merge.

## Scope (6 files, +1761 / -0)
- \`dflash/src/gemma4_target_loader.cpp\` — **+447 lines, purely additive**. Re-adds the four MTP loader/helper symbols that were intentionally stripped from this file in PR #171 (per the PR04 spec):
  - \`load_gemma4_mtp_assistant(path, backend, out)\`
  - \`free_gemma4_mtp_assistant(w)\`
  - \`get_mtp_swa_pattern(path, out)\`
  - \`resolve_mtp_donor_layers(mtp, target_swa_layers)\`
  None of the PR #171 target-loader behavior is reopened.
- \`dflash/src/gemma4_mtp_graph.cpp\` — **new file, 760 lines**. Defines \`build_mtp_step_graph(...)\` (the single-step MTP forward) and \`free_mtp_step_graph(...)\` (cleanup). Each MTP layer reads target K/V from \`w.layers[il].donor_target_layer\` (resolved at load time).
- \`dflash/src/internal.h\` — **+116 lines**. Adds:
[...]

PR #183, opened 2026-05-13
feat(gemma4): target-graph MTP integration (h_prev capture + asymmetric KV)

dusterbloom:
## Validation (for review)

Upstream `llama-bench` baseline (Gemma4-31B-it Q4_K_M, q4_0 KV, FA on, RTX 3090): **pp16384 = 376.62 tok/s, tg300 = 20.39 tok/s**.

MTP `h_prev` capture in target graph (post-output-norm hidden, fed to MTP cross-attention) + asymmetric-KV override (MTP donor layers forced to Q8 so MTP can read them).

| Metric | Without PR | With PR |
|---|---|---|
| MTP accept rate (γ=1) | 0% (h_prev was pre-norm) | non-zero (post-norm) |
| MTP runs end-to-end | ✗ | ✓ |
| MTP cross-attn on TQ3 cache | ✗ (TQ3 unreadable) | ✓ (donor Q8 override) |
| Byte-identical at γ=1 | n/a | required for accept>0; verified |

---
PR #11a of the Gemma4 split sequence (revised split). Wires the MTP read/write halves into the target graph and cache that #176 / #182 introduced.

**Independent of #175** (the daemon refactor) — pure graph-level integration. The runtime decode loop / daemon wiring move to PR #11b (\`Gemma4Backend\` class) and PR #13 (\`gemma4_daemon\` binary), both of which depend on #175.

Depends on #182 → #181 → #180 → #179 → #177 → #176 → #171 → #170 → #169 → #168.

## Scope (2 files, +124 / -5)
- \`dflash/src/gemma4_target_graph.cpp\`:
  - \`create_gemma4_cache\` signature gains \`extra_q8_layers\` + \`enable_dflash_capture_overrides\` parameters (both default-off / empty so existing callers in #176/#177 keep compiling).
  - Adds the asymmetric KV gate (force Q8_0 on capture layers when TQ3_0 KV is selected and capture overrides are enabled) — the workaround for the upstream FA chunked-kernel trap on Dense 31B \`head_dim=512\`.
  - Adds the MTP donor override (force Q8_0 on caller-specified layer indices that MTP cross-attention reads via \`ggml_cast\` and would otherwise pick up TQ3/FWHT-domain values).
  - Adds the MTP h_prev capture block to \`build_gemma4_graph\` after final RMSNorm. Two modes: single-row (default, γ=1) and batch (\`mtp_h_prev_capture_mode == 1\`, γ>1 partial-accept). Both gated by \`cache.mtp_h_prev_enabled\` — default-off, so non-MTP callers pay nothing.
- \`dflash/src/internal.h\`:
  - Updates the \`create_gemma4_cache\` prototype to match.
  - The \`mtp_h_prev_*\` fields on \`GemmaTargetCache\` already shipped with PR #182 (MTP loader/step graph), so this PR is purely the **writer** half of the MTP h_prev contract — the **reader** (the MTP step graph) is in #182.

## Risk: Medium-High
Touches a hot path in \`build_gemma4_graph\`. Default-off guards mean non-MTP callers see no behavioral change. \`extra_q8_layers\` / \`enable_dflash_capture_overrides\` defaults preserve PR #176's smoke-test setup byte-for-byte.

## What this PR is NOT (deferred)
- **Gemma4Backend** class (\`ModelBackend\` impl): PR #11b, depends on #175.
- **gemma4_daemon** binary entry: PR #13 (revised), depends on PR #11b.
- **\`test_gemma4_dflash.cpp\` runtime hunks** for \`--draft-method mtp\`: those are subsumed by PR #11b's \`Gemma4Backend::generate()\` which dispatches the draft method internally per the \`Gemma4BackendArgs::draft_method\` field. The original test-driver-flag approach is obsolete now that the daemon protocol moves to PR #175's \`ModelBackend\` interface.

## Review checklist
- Default param values preserve existing call sites ✓ (4-arg \`create_gemma4_cache(w, max_ctx, backend, cache)\` still compiles; smoke + KV tests don't enable MTP)
- MTP h_prev write site is at the same post-RMSNorm location used by the MTP step graph readers (PR #182 \`MtpDrafterWeights\`) ✓
- TQ3/FWHT donor-layer override is documented in-source with the cross-attention rationale ✓

## Validation
- \`g++ -fsyntax-only gemma4_target_graph.cpp\` via project's compile_commands flags → **exit 0**
- Full \`cmake --build\` deferred to CI (no local CUDA budget this turn).


dusterbloom:
Closing: targets obsolete dflash/src/gemma4_*.cpp paths. h_prev capture + asymmetric KV would be redone against IExternalDrafterMtp (declared in #237) if Gemma4 MTP is picked up post-#237. The branch content stays as reference via git history.

PR #189, opened 2026-05-13
chore(workspace): introduce uv workspace for dflash, megakernel, pflash

easel:
## Summary

Introduces a root-level **uv workspace** that unifies Python dependency management across the three subprojects (`dflash`, `megakernel`, `pflash`) into a single shared `.venv` and one committed `uv.lock`.

Today each subproject documents its own `python -m venv .venv` + `pip install …` flow, which:
- creates 2–3 separate venvs duplicating ~10 GB of torch + NVIDIA wheels,
- has no lockfile, so PRs silently drift on transitive versions,
- requires `--no-build-isolation` gymnastics for `megakernel` that every contributor has to discover the hard way.

This PR wires all of it through `uv` while keeping the legacy `pip` flows working for anyone who doesn't have `uv` installed.

## What's in the workspace

| File | Role |
|---|---|
| `pyproject.toml` (root) | virtual project (`package = false`); declares workspace members + the PyTorch cu130 index (explicit, only `torch` resolves there) |
| `.python-version` | pins 3.11 for the workspace |
| `dflash/pyproject.toml` | virtual project; pins `fastapi`, `uvicorn`, `transformers`, `jinja2`, `torch`, `numpy`, `gguf`, `datasets`, `huggingface-hub` for `scripts/server.py` + bench harness |
| `megakernel/pyproject.toml` | declares `qwen35-megakernel-bf16`; setup.py retained for the CUDA extension; listed under `[project.optional-dependencies] megakernel` so default `uv sync` skips the compile |
| `pflash/pyproject.toml` | already existed; picked up unchanged as a workspace member |
| `uv.lock` | 1350 lines, fully pinned (torch 2.12.0+cu130, transformers 5.8.1, fastapi 0.136.1, …) |

`no-build-isolation-package = ["qwen35-megakernel-bf16"]` is set at the root so the megakernel CUDA extension builds against the shared `.venv` torch (matches the prior `pip install -e . --no-build-isolation` requirement).

## Usage

```bash
# Everything except the megakernel CUDA extension
uv sync

# Also compile megakernel (sm_86 / auto-detected; needs nvcc)
uv sync --extra megakernel

# Run anything in the workspace
uv run --directory dflash python scripts/server.py --port 1236 --daemon
uv run --directory megakernel python final_bench.py
```

## Doc changes

- Top `README.md`: install blocks for both megakernel and dflash quickstarts switched to `uv sync` (legacy flow kept as a callout)
- `dflash/README.md`: the OpenAI-compatible server snippet now uses `uv run --directory dflash python scripts/server.py …`
- `CONTRIBUTING.md`: adds `uv ≥ 0.5` to the prereq table and a short note that the per-subproject pip flow still works

## CI

New `uv-workspace` job (CPU-only, fast) in `.github/workflows/ci.yml` running:

1. `uv lock --check` — fails the PR if `uv.lock` drifts from any `pyproject.toml`
2. `uv sync --frozen --no-install-package torch` — verifies the resolution is reproducible (skips the ~1 GB nvidia wheels since this job doesn't exercise GPU code)
3. Import smoke for `fastapi`, `uvicorn`, `transformers`, `jinja2`, `numpy`, `gguf`, `datasets`, `huggingface_hub`, and `dflash/scripts/server.py`

The existing `Build (CUDA 12.8)` job is untouched.

## Verified locally

RTX 3090 Ti, CUDA 13.2, Ubuntu 24:

- [x] `uv sync` -> 81 packages resolved, dflash + pflash usable
- [x] `uv sync --extra megakernel` -> `qwen35_megakernel_bf16_C.cpython-311-*.so` built in 37s (sccache hit), `import torch; import qwen35_megakernel_bf16_C` succeeds
- [x] `uv lock --check` clean, `uv sync --frozen` reproducible
- [x] `dflash/scripts/server.py` imports cleanly via `uv run`
- [x] Live test: server restarted from the root `.venv` on `:1236` with `--max-ctx 131072 --prefill-compression auto --budget 16`, `/v1/models` returns 200, daemon healthy

## Test plan

- [ ] CI `uv-workspace` job passes
- [ ] CI `Build (CUDA 12.8)` job still passes (unchanged)
- [ ] On a fresh clone, `uv sync` succeeds without manual torch install
- [ ] On a CUDA-equipped machine, `uv sync --extra megakernel` compiles the extension without an explicit `--no-build-isolation` flag


easel:
@davide221 straw man of a uv project structure -- benefit is pinned deps and executable documentation of the dependency tree. The readme has typically been missing transformers and some others.

PR #190, opened 2026-05-13
feat(dflash): add /props introspection endpoint

easel:
## Motivation

I’m working on a set of agent workflow benchmarks and want to capture enough runtime metadata per run to make benchmark cells self-describing: model path, context window, sampling defaults, speculative mode, KV cache settings, pflash/cache state, and daemon liveness. This adds a llama.cpp-style `/props` endpoint for that purpose, with lucebox-specific details kept as structured extensions.

## Summary

Adds a read-only `GET /props` endpoint returning a JSON snapshot of the live Python-server state for bench-time capture and diagnostics.

The endpoint now uses the cross-server / llama.cpp-compatible shape expected by downstream runtime-props capture:

- top-level `default_generation_settings.{n_ctx, temperature, top_p, top_k, min_p, repeat_penalty}`
- top-level `model_alias`, `model_path`, and `build_info`
- top-level `speculative_mode` with `off`, `dflash`, or `pflash`
- `runtime.backend` plus lucebox runtime extensions such as KV cache types, FA window, lazy draft, and target sharding
- `reasoning.{supported, default, supported_efforts}`
- `sampling.capabilities` for operator-visible request parameter support flags

Old aliases are intentionally not emitted: `runtime.max_ctx`, `model.id`, `model.target_path`, flat `sampling.supports_*`, and `reasoning.default_enabled`.

## Response Shape

Top-level keys:

`{ default_generation_settings, model_alias, model_path, build_info, speculative_mode, server, model, runtime, reasoning, speculative, sampling, pflash, prefix_cache, full_cache, tool_replay, daemon, api }`

Each section is scoped to Python-server state. Daemon build identity, request-rate metrics, and per-endpoint parameter schemas remain v1 non-goals.

## Design Notes

- `server.version` is read from `dflash/pyproject.toml` via stdlib `tomllib`; malformed pyproject logs a warning and falls back to `"0.0.0+unknown"`.
- `server.props_schema = 1` remains the compatibility marker for `/props` parsers.
- `runtime.kv_cache_k/v` report effective daemon KV types via `_effective_kv_type()`.
- `runtime.backend` is best-effort: `DFLASH_RUNTIME_BACKEND` / `DFLASH27B_GPU_BACKEND`, then nearby `CMakeCache.txt`, then `cuda` fallback.
- `speculative_mode` is `pflash` when pflash is enabled, otherwise `dflash` when DDTree speculative decode is supported, otherwise `off`.
- `prefix_cache` and `full_cache` expose cumulative lifetime hit counters.
- `full_cache.disk_bytes` is snapshotted on mutation so `/props` does not walk the filesystem on read.
- `tool_replay` reports the exact tool-call replay memory counters from `ToolMemory`.
- `api.endpoints` is hand-curated with a drift test against FastAPI routes.

## Tests

Focused server suite passes:

- [x] `uv run --extra dev pytest dflash/scripts/test_server.py -q` — 75 passed

Covered areas include endpoint shape, removed old aliases, default generation settings, reasoning shape, speculative mode selection, backend resolution, arch gating, pflash toggle, target-sharding cache behavior, endpoint-list drift, KV type resolution, cache counters, full-cache disk-byte snapshots, and `ToolMemory.stats()`.

## Open Items

- [ ] Run `curl http://localhost:8000/props | jq .` against a live server after deployment/restart and sanity-check runtime values.

### DIFF
--- a/dflash/scripts/prefix_cache.py
+++ b/dflash/scripts/prefix_cache.py
@@ -484,6 +484,9 @@ class PrefixCache:
 
         self.entries: OrderedDict[bytes, int] = OrderedDict()  # hash → slot_id
         self.next_slot = 0
+        # Cumulative hit counter, never decremented. Survives eviction —
+        # unlike a sum over surviving entries, which trends down under churn.
+        self._lifetime_hits = 0
         try:
             self.markers = _resolve_chat_markers(tokenizer)
         except ValueError as e:
@@ -537,9 +540,21 @@ class PrefixCache:
                     best = (self.entries[key], cut)
                 self.entries.move_to_end(key)   # mark fresh
         if best is not None:
+            self._lifetime_hits += 1
             print(f"{self.log_prefix} lookup hit slot={best[0]} prefix_len={best[1]} "
                   f"(of {len(prompt_ids)} total)", flush=True)
         return best
+
+    def stats(self) -> dict:
+        """Snapshot for /props. Lockless: a mutation under daemon_lock can
+        tear in_use vs lifetime_hits; acceptable for an introspection report."""
+        if self.disabled:
+            return {"capacity": 0, "in_use": 0, "lifetime_hits": 0}
+        return {
+            "capacity": self.cap,
+            "in_use": len(self.entries),
+            "lifetime_hits": self._lifetime_hits,
+        }
 
     def mark_all_cleared(self) -> None:
         """Drop every LRU entry after the daemon emits ``[snap] all-cleared``.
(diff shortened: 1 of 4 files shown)


easel:
@cubic-dev-ai thanks for the review — both flagged against commit `5600c93`. Status on current HEAD (`f6f8e97`):

**Issue 1 (server.py:150, unvalidated `float(DFLASH_FP_ALPHA)`)** — already addressed in the first amendment (`d492983`). The parse goes through a new `_parse_optional_float()` helper that returns `None` and logs a warning on non-numeric values rather than raising at request time.

**Issue 2 (server.py:926, `target_sharding` misreport on laguna)** — valid bug, fixed in `f6f8e97`. Root cause: the laguna daemon-spawn path (`if arch in _LAGUNA_ARCHES: cmd = [...]`) doesn't call `cmd.extend(extra_daemon_args)` — that branch is qwen35-only. So on `arch=laguna` with `--target-gpus=...` passed, the flag is silently dropped at spawn but `/props` was still reporting `target_sharding: True`. Now gated:

```python
"target_sharding": (
    arch not in _LAGUNA_ARCHES
    and _extra_daemon_has_target_sharding(extra_daemon_args)
),
```

Added `test_props_target_sharding_false_on_laguna_even_when_args_passed` to lock the behavior in. 72/72 tests passing.

easel:
Superseded by integration PR #226 via 5b67cf2 (server props + thinking controls reworked into the integration line).

PR #202, opened 2026-05-16
Fast CI workflow run

howard0su:
Since we only care if CUDA is broken, we only do one arch build. Also enable pip cache


howard0su:
reduce to 18mins

PR #214, opened 2026-05-17
feat(mtp): MTP-via-daemon end-to-end (incl. MTP infrastructure)

dusterbloom:
## Scope: full MTP stack + daemon wiring + PR #213 daemon fixes

This PR brings TWO things to main that didn't exist there before:

1. **The Qwen3.6 MTP module** (`dflash/src/qwen36/qwen36_mtp.{cpp,h}` and `qwen36_mtp_graph.{cpp,h}`, ~2.8k LOC). This was developed on `feature/mtp-foundation-v2` over the past weeks but never landed on `main`.
2. **MTP-via-daemon wiring** (~500 LOC) — what was originally scoped here: server.py `--mtp-gguf`, test_dflash dispatch, Qwen35Backend MTP integration.
3. **PR #213 daemon fixes** included so this PR is self-contained against `main` (those fixes are also up as PR #213 for narrower review).

Total: **+3324 / -60 LOC**. Above the 3k cap. **Recommend reviewers process this in two passes:**
- First: read PR #213's narrower diff (the 4 daemon fixes) — that's ~140 LOC and already reviewable
- Then: this PR diff minus #213 = ~500 LOC of MTP daemon wiring on top of the (existing on feature branch but new to main) ~2.8k LOC of MTP module

If preferred, this PR can be split into:
- **PR-A**: MTP module land (~2.8k LOC, just the infrastructure, no daemon use)
- **PR-B**: MTP daemon wiring (~500 LOC, atop PR-A)
- **PR-#213**: separate daemon fixes (already up)

Say the word and I'll re-split.

## Headline measurements

**Claude Code on 24K-token system prompt:**

| Speculator | decode | accept | comment |
|---|---|---|---|
| DFlash (DDTree b=22) | 22.0 tok/s | n/a | PR #213 baseline |
| **MTP (γ=3, daemon)** | **35.8 tok/s** | **0.41** | **+63% vs DFlash** |

**Mean across 7 harness clients (3 back-to-back probes, 21 client-cells):**

| Speculator | mean decode | median | max | out=0 cells |
|---|---|---|---|---|
| DFlash | 22.0 tok/s | — | — | 0 / 57 |
| **MTP** | **29.3 tok/s** | **29.6** | **50.6** | **0 / 57** |

**vs Lucebox blog (RTX 3090, Qwen3.6 Q4_K_M, DDTree, published 22.6-29.6 tok/s range):**
- Our MTP mean 29.3, max 50.6 → **at or above the top of the published range**

## Architecture summary

Same daemon binary, same bare-prompt protocol. MTP is a startup mode (`--mtp-gguf` flag). Backend extends `Qwen35Backend` with optional `Qwen36MtpModule` (no new backend class — keeps target loader/snapshot/park/cache-reset shared). Per-request `mtp_module_->reset_chain()` mirrors PR #213's `reset_target_cache(cache_)` so MTP also avoids the state-leak cascade under sustained load.

## Validated

- `pytest dflash/scripts/test_server.py` — 56/56 pass
- `harness/client_test_runner.py probe` × 3 back-to-back × 7 clients = 21/21 cells green on MTP
- Same probe set on DFlash (no `--mtp-gguf`) — 7/7 (no regression from PR #213 baseline)
- `bench_agent.py --bucket 2k --n-sample 8 --skip-ar --budget 22` — mean 53.37 tok/s vs canonical 46.70 = +14% (no regression on file-mode harness path)
- `claude --print --model luce-dflash` real coding task on MTP — valid Heron's formula output
- DFlash regression smoke — "lucebox lucebox lucebox" identical to PR #213

## Known follow-ups (separate PR after this lands)

1. **Pre-norm hidden capture path warning** ("hidden_at_pos_pre_norm returned null"). Currently degrades accept-rate but functional. Fix would push accept from ~0.3 → model-card-recommended 0.83 for γ≤2, lifting MTP decode mean toward 50+ tok/s.
2. **Prefix-cache RESTORE in MTP mode** — disabled because MTP head KV isn't snapshotted. Adding to `PrefixSnapshot` would unlock TTFT savings (DFlash already gets 22s → 6s warm).
3. **mtp_topk draft-source path** plumbed but not implemented in `do_mtp_decode_` — chain path is production default per matrix bench D=3.

## Test plan

- [x] All tests pass
- [x] Harness probes green on MTP + DFlash (no regression)
- [x] Matrix bench no regression
- [x] Real client end-to-end (Claude Code generates valid code)
- [x] Beat blog decode (35.8 tok/s Claude Code vs 22.6-29.6 published)
- [ ] (review) Confirm 2.8k LOC MTP module review acceptable in one pass, or request split
- [ ] (next PR) Pre-norm fix → accept_rate → 0.83+
- [ ] (next PR) MTP head KV snapshot → prefix-cache works in MTP mode

### DIFF
--- a/dflash/src/qwen35/qwen35_backend.h
+++ b/dflash/src/qwen35/qwen35_backend.h
@@ -19,6 +19,7 @@
 #include "dflash_feature_ring.h"
 #include "internal.h"         // TargetWeights, TargetCache, DraftWeights, PrefixSnapshot
 #include "qwen3/qwen3_drafter.h"  // DrafterContext, load_drafter, free_drafter, drafter_score_and_compress
+#include "qwen36/qwen36_mtp.h"    // Qwen36MtpModule
 
 #include "ggml.h"
 #include "ggml-backend.h"
@@ -26,6 +27,7 @@
 #include <memory>
 #include <random>
 #include <string>
+#include <vector>
 
 namespace dflash27b {
 
@@ -54,6 +56,12 @@ struct Qwen35Config {
     float        ddtree_temp     = 1.0f;
     bool         ddtree_chain_seed = true;
     bool         use_feature_mirror = false;
+
+    // MTP (Multi-Token Prediction) speculator — mutually exclusive with draft
+    const char * mtp_gguf_path    = nullptr;   // path to fused MTP GGUF (or nullptr = DFlash)
+    int          mtp_gamma        = 0;         // max speculation depth
+    const char * mtp_draft_source = nullptr;   // "chain" | "mtp_topk" | nullptr -> "chain"
+    int          mtp_draft_topk   = 1;         // top-k for mtp_topk mode
 };
 
 // ── Backend class ───────────────────────────────────────────────────────
@@ -101,6 +109,16 @@ public:
     bool supports_dflash_spec_decode() const override { return true; }
     DFlashTarget * dflash_target() override;
 
+    // Test/bench integration hooks for native-head MTP. These keep the
+    // Qwen3.6 MTP harness on the backend-owned target/cache/context without
+    // exposing the frozen common interfaces.
+    bool ensure_decode_cache(int max_verify_tokens);
+    ggml_context * tensor_context() const;
+
+    // MTP speculator accessors (ModelBackend interface).
+    bool                   supports_mtp() const override { return mtp_module_ != nullptr; }
+    mtp::IMtpModule *      mtp()                override { return mtp_module_.get(); }
+
     void shutdown() override;
 
 private:
@@ -144,6 +162,9 @@ private:
     // ── DFlashTarget adapter (lazy-built) ────────────────────────────
     std::unique_ptr<DFlashTarget> dflash_target_;
 
+    // ── MTP speculator (optional, set when cfg_.mtp_gguf_path != nullptr) ──
+    std::unique_ptr<mtp::Qwen36MtpModule> mtp_module_;
+
     // ── Internal helpers ─────────────────────────────────────────────
     // Prefill a prompt and return the number of tokens committed to KV.
     // kv_offset > 0 resumes from a restored snapshot: tokens are placed at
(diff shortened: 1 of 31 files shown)


davide221:
@dusterbloom thanks for the great feature! Can you fix merge conflics?

dflash/src/qwen36/qwen36_mtp.cpp
```diff
@@ -0,0 +1,2100 @@
+// qwen36_mtp.cpp — see qwen36_mtp.h for contract.
```
howard0su, line 1: check https://huggingface.co/Qwen/Qwen3.6-27B/blob/main/config.json#L7, there is no qwen3.6 model type. keep this as qwen3.5.

dflash/src/qwen36/qwen36_mtp.cpp
```diff
@@ -0,0 +1,2100 @@
...
+    for (int r = 0; r < rows; r++) {
+        float acc = 0.0f;
+        const float * row = A + (size_t)r * cols;
+        for (int c = 0; c < cols; c++) acc += row[c] * x[c];
+        y[r] = acc;
+    }
+}
+
+// Argmax over a float vector; returns index of max element.
+static int32_t argmax(const float * logits, int n) {
```
howard0su, line 153: this is dup code of ggml, which has cpu version of these.

dflash/src/qwen36/qwen36_mtp.cpp
```diff
@@ -0,0 +1,2100 @@
+// qwen36_mtp.cpp — see qwen36_mtp.h for contract.
+//
+// PR 2d-bis (Shape B): implements the full DeepSeek-V3 NextN per-head forward
```
howard0su, line 3: please leverage ggml ops to CPU side of calculation. the reason is not just reduce duplicate code but also enable multi-gpu scenerio that we can offload path A to GPU-B

dflash/src/qwen35/qwen35_backend.h
```diff
@@ -144,11 +162,17 @@ class Qwen35Backend : public ModelBackend {
     // ── DFlashTarget adapter (lazy-built) ────────────────────────────
     std::unique_ptr<DFlashTarget> dflash_target_;
 
+    // ── MTP speculator (optional, set when cfg_.mtp_gguf_path != nullptr) ──
+    std::unique_ptr<mtp::Qwen36MtpModule> mtp_module_;
```
howard0su, line 166: what's the concept of Module here? I prefer that MTP is another set of weights. The current MTP code should be generic enough to handle gemma4 as well.

howard0su (changes requested):
The PR is not ready for checkin:
1. Please leverage ggml for CPU code (or make it generic to support any ggml backend)
2. MTP should be simple as additional weights of modelbackend
3. If a model contains MTP support (no matter gemma4 or qwen3.5), the logic can handle it. In other word, the logic should be in /common which can potentially leverage by any modelbackend if they support mtp.
4. You may want to introduce some additional interface inside modelbackend class.

howard0su:
also, after the change of RoPE, please retest 24k context baseline (it was the bug).

dusterbloom:
@cubic-dev-ai addressed both review comments in 186bccc:

**P1 (`common/mtp_orchestrator.cpp:77` heap UB)** — the `all_prefill_hidden.clear()` branch is gone. Capture is enabled+pinned at MTP attach, so the only way `verify_batch` returns a short chunk is a contract break; we now error out via `result.error = "hidden seq capture invariant violated"` and bail before the next chunk's `memcpy` can write past freed memory.

**P2 (`qwen35_backend.cpp:361` ignored return)** — hoisted the `migrate_prefill_cache` call out of `generate()` into `init_mtp_()`. `max_ctx` and γ are config-time constants, so a single call at attach is correct by construction; the bool return is now checked and a failure tears the module down and fails backend init cleanly (no per-request OOM path → no null `ssm_intermediate` → no segfault).

γ=3 re-bench unchanged: 49.7 tok/s decode, accept 0.78, zero `pre_norm` warnings, 7/7 harness clients green. γ=2 also verified: 49.5 tok/s, accept 0.81, 7/7 green.

dusterbloom:
@howard0su 24K Heron retest with γ=3 on the refactored stack (MTP orchestrator in `common/`, abstract `ModelBackend`, capture-mode pinned at attach):

- decode 49.7 tok/s
- accept rate 0.78 (model card ceiling 0.83)
- zero `pre_norm null` warnings (correct by construction — capture mode is pinned at attach, runtime toggle is a no-op)
- 7/7 harness clients green (claude_code, codex, opencode, openwebui, pi, openclaw, hermes)

γ=2 also tested: 49.5 tok/s, accept 0.81 (within noise of γ=3 on throughput, slightly higher accept). Keeping γ=3 default — tie is in noise, no reason to churn config.

dusterbloom:
superseded by #237

Issue #191, opened 2026-05-14
Nvidia Tesla V100 running e5347801719ad7d45a3d7bd096e9e57778ce23ea


merbanan:
```
python3 scripts/bench_llm.py
[bench] target    = /media/per/work/tmp/lucebox-hub/dflash/models/Qwen3.6-27B-Q4_K_M.gguf
[bench] draft     = /media/per/work/tmp/lucebox-hub/dflash/models/draft/model.safetensors
[bench] ar bin    = /media/per/work/tmp/lucebox-hub/dflash/build/test_generate
[bench] df bin    = /media/per/work/tmp/lucebox-hub/dflash/build/test_dflash
[bench] tokenizer = Qwen/Qwen3.5-27B
[bench] budget    = 22

[bench] ==== HumanEval (n=10, n_gen=256) ====
  [01/10] n_tok=  84  AR= 30.14  DFlash= 102.56  AL=11.64
  [02/10] n_tok= 138  AR= 30.16  DFlash=  77.68  AL= 8.83
  [03/10] n_tok= 134  AR= 30.14  DFlash=  36.16  AL= 4.13
  [04/10] n_tok= 120  AR= 30.15  DFlash=  38.40  AL= 4.34
  [05/10] n_tok= 172  AR= 30.08  DFlash=  29.92  AL= 3.41
  [06/10] n_tok= 118  AR= 30.16  DFlash=  40.87  AL= 4.65
  [07/10] n_tok=  51  AR= 30.15  DFlash=  36.49  AL= 4.06
  [08/10] n_tok= 141  AR= 27.51  DFlash=  36.46  AL= 4.13
  [09/10] n_tok= 125  AR= 25.40  DFlash=  32.44  AL= 3.71
  [10/10] n_tok=  95  AR= 23.62  DFlash=  51.27  AL= 5.82
  HumanEval mean: AR=28.75  DFlash=48.23  AL=5.47  1.68x

[bench] ==== GSM8K (n=10, n_gen=256) ====
  [01/10] n_tok=  45  AR= 25.66  DFlash=  36.25  AL= 4.13
  [02/10] n_tok= 111  AR= 22.17  DFlash=  32.01  AL= 3.66
  [03/10] n_tok=  49  AR= 21.40  DFlash=  26.29  AL= 2.98
  [04/10] n_tok=  70  AR= 20.21  DFlash=  33.55  AL= 3.76
[...]

PR #220, opened 2026-05-18
fix: V100 daemon crashes + Math500 0/10 (issue #191)

davide221:
Fixes the three distinct bugs behind #191. All verified on a Tesla V100
(sm_70, 32 GB) running Qwen3.6-27B-Q4_K_M + the DFlash 3.6 draft.

## 1. `test_dflash` exits 1, writes no output

The DDTree decode loop assigns the tree walk's `next_token` to `last_tok`
and only breaks on `hit_eos`. When the walk yields no continuation,
`next_token` is -1: the next iteration feeds token -1 to
`w.embedder.embed()`, which fails, and the loop `return 1`s with no
message. The output file is never written and the `tok/s` summary never
prints. `bench_llm.py` reports this as `test_dflash exited 1` and aborts.
Under `--daemon` the same -1 kills the server mid-request.

Fix: treat `next_token < 0` (and a trailing EOS) as a normal stop.

## 2. Out-of-bounds argmax read in `do_spec_decode` (regression from #209)

`do_spec_decode` sampled the first token by re-reading `sg_.argmax_tokens`
at `committed % 512`, with `PREFILL_UBATCH` hardcoded to 512. The decode
step graph's `argmax_tokens` tensor is not sized for the prompt, so any
prompt longer than that tensor overruns it:

```
ggml-backend.cpp: GGML_ASSERT(offset + size <= ggml_nbytes(tensor)
                  && "tensor read out of bounds") failed
```

The daemon aborts on the first real-sized request (short prompts only
survive by landing inside the tensor by luck). This is why an agent
client could "issue a few prompts but not do real work".

Fix: `do_prefill` already stores the last-position argmax in
`cache_.last_tok` using the correct per-chunk offset. Read that.

## 3. Math500 scored 0/10 — `bench_llm.py` sent raw prompts

`_wrap_prompt` only applied the chat template under `--no-thinking`; the
default path sent the bare `"Problem: ...\nSolution:..."` string to the
tokenizer. An instruct/thinking model given a raw completion prompt
rambles and never emits a scorable `\boxed{}` answer — 0/10 on every GPU,
not a V100 issue.

Fix: apply the chat template whenever the tokenizer has one;
`--no-thinking` now only toggles `enable_thinking`.

## Verification (V100, sm_70)

- `test_dflash` ddtree run: was exit 1 / no output -> exit 0, output written, summary printed.
- Server probe (chat completions, streaming, tool calls, repeated requests): was garbage / daemon dead after a few requests -> coherent output, valid `tool_calls`, daemon survives, spec-decode ~45 tok/s.
- Math500 (n=3, `--no-thinking`): 0/10 -> 3/3.

Bug 2 is a regression in the merged #209 and crashes the server on the
first agent-sized prompt, so it should land promptly.

### DIFF
--- a/dflash/src/qwen35/qwen35_backend.cpp
+++ b/dflash/src/qwen35/qwen35_backend.cpp
@@ -640,18 +640,13 @@ bool Qwen35Backend::do_spec_decode(int committed, int n_gen,
                                     const DaemonIO & io) {
     const int hidden = w_.n_embd;
 
-    // Sample first token from prefill's last-position argmax.
-    // The last chunk used last_token_logits_only=false, so sg_.argmax_tokens
-    // holds argmax for ALL positions in that chunk. We need the LAST position.
-    int32_t last_tok;
-    {
-        const int PREFILL_UBATCH = 512;
-        int n_last_chunk = committed % PREFILL_UBATCH;
-        if (n_last_chunk == 0) n_last_chunk = PREFILL_UBATCH;
-        ggml_backend_tensor_get(sg_.argmax_tokens, &last_tok,
-                                sizeof(int32_t) * (n_last_chunk - 1),
-                                sizeof(int32_t));
-    }
+    // First token = prefill's last-position argmax. do_prefill already read it
+    // into cache_.last_tok using the correct per-chunk offset. Re-reading
+    // sg_.argmax_tokens here with a hardcoded ubatch indexed it by
+    // `committed % 512`, which overruns the decode-step argmax tensor for any
+    // prompt longer than that tensor (issue #191: ggml "tensor read out of
+    // bounds" -> daemon abort on the first real-sized request).
+    int32_t last_tok = cache_.last_tok;
 
     // Check if we can use speculative decode:
     // - draft model loaded and not parked
--- a/dflash/scripts/bench_llm.py
+++ b/dflash/scripts/bench_llm.py
@@ -323,12 +323,16 @@ def main():
         )
 
     def _wrap_prompt(raw_prompt: str) -> str:
-        if not args.no_thinking:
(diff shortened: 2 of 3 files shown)


pebaryan:
Verified on Tesla V100 (32GB) — IQ4_NL + DFlash draft

Ran the full benchmark suite with the patch applied (commit 1133db6). Math500 went from 0/10 to 1/3 (33%) with structured \boxed{} answers.

dflash/src/qwen35/qwen35_backend.cpp
```diff
@@ -640,18 +640,13 @@ bool Qwen35Backend::do_spec_decode(int committed, int n_gen,
...
-                                sizeof(int32_t) * (n_last_chunk - 1),
-                                sizeof(int32_t));
-    }
+    // First token = prefill's last-position argmax. do_prefill already read it
+    // into cache_.last_tok using the correct per-chunk offset. Re-reading
+    // sg_.argmax_tokens here with a hardcoded ubatch indexed it by
+    // `committed % 512`, which overruns the decode-step argmax tensor for any
+    // prompt longer than that tensor (issue #191: ggml "tensor read out of
+    // bounds" -> daemon abort on the first real-sized request).
+    int32_t last_tok = cache_.last_tok;
```
howard0su, line 649: will this impact the quality of output? this looks like a behavior change and the math is different.

howard0su (approved):

dflash/src/qwen35/qwen35_backend.cpp
davide221: But do_prefill() already knows the actual final chunk size when it runs, so it reads the slot there and stores it in cache_.last_tok. Quality should be the same

PR #221, opened 2026-05-18
mtp: prefix-cache WARM hit (perfect + partial via range-warm)

dusterbloom:
Stacks on top of #214 (`feat/mtp-via-daemon`). The two new commits in this PR are:

- `8409b01 mtp: native-heads MTP speculator (Qwen3.6 NextN, γ-chain)` — identical to the head of #214; merges via #214.
- `0170c40 mtp: prefix-cache WARM hit (perfect + partial via range-warm)` — the new work in this PR. After #214 merges, this PR's effective diff collapses to just this commit.

## What's in the WARM-hit commit

Plugs MTP requests into the existing `server.py` prefix-cache protocol so agent loops with repeated prefixes skip the cold prefill + warm path.

- **Inline snap ack** (`common/prefix_snap.h`) — single source for `[snap] inline slot=N cur_pos=M`. DFlash `do_prefill` and the MTP orchestrator both funnel through `emit_inline_snap_ack` so the format can't drift.
- **Mid-prompt snap in the MTP orchestrator** — chunked prefill clips to `snap_at` (server's `prepare_inline_snap` picks the second-to-last `<|im_start|>` boundary, almost always mid-prompt). Partial `warm_head_kv` fills slots `[1..snap_at]`, `snapshot_save` fires through `ModelBackend`, ack emits. Full warm runs after to cover `[1..prompt_len)`.
- **Snapshot captures `h_{snap_pos-1}`** via `target->last_hidden()` + a shape contract (γ / n_head_kv / n_ctx / n_embd). Restore rejects mismatches before touching state.
- **`INativeMtp::warm_head_kv_range`** — same warm graph as `warm_head_kv`, caller-controlled `slot_start`, used by partial-WARM restore.
- **`restore_and_generate` unifies perfect-WARM and partial-WARM**:
  - Perfect-WARM (`prompt_len == snap_pos`): restore head_kv, no prefill.
  - Partial-WARM (`prompt_len > snap_pos`): restore head_kv, prefill delta, range-warm `[snap_pos..prompt_len]` using `pre_warm_hidden + delta_hiddens`. Slot `snap_pos` is overwritten with the new request's `prompt[snap_pos]` so cross-request first-after-cut divergence cannot corrupt decode.
  - Contract failures fall through to a cold-restart fallback that discards the snapshot's head_kv and runs full cold prefill — byte-correct against the chain runner's `[0..prompt_len)` requirement.
- **Thin head_kv snapshot** — `[0..head_kv_pos+1]` slice instead of full `[key_len, n_ctx, n_head_kv]`. ~10× payload reduction at `max_ctx=65536`.
- **Lazy GPU head_kv alloc** — grows from 8K initial slots to `n_ctx_max` on the first warm that needs more (saves ~256 MiB VRAM on a daemon serving short prompts).
- **`Qwen35Backend::init_mtp_`** passes `cfg_.device.max_ctx` to the module so head_kv tracks the backbone (previous hardcoded 8192 broke any prompt > 8K).

## Tests

- 4/4 `test_prefix_cache_mtp` (perfect bit-equal round-trip, prefill_next mismatch fallback, int32 boundary, shape-contract rejection).
- 4/4 `test_common_mtp_orchestrator` (unchanged from #214).
- Full `cmake --build` green.
- Harness probe 7/7 at `--max-ctx 65536` (claude_code, codex, opencode, openwebui, pi, hermes, openclaw). Zero chain failures, zero ack-missing warnings, zero capacity overflows.

## Bench

`dflash/bench/bench_mtp_warm_hit_2turn.py` is a 2-turn warm-hit gate that drives a live daemon: `prefill_ratio < 0.3` + bit-equal first 16 tokens between turns.

## Follow-ups (not in this PR)

- Tree-MTP arena (B≥2 sibling drafts) is being prototyped on a separate branch.


dusterbloom:
Closing as stale — superseded by the merged MoE-hybrid / prefix-cache / harness infra. Will reopen from a fresh branch on current main if revived.

PR #225, opened 2026-05-19
feat/strix-halo-amd-optimizations — AMD Strix Halo HIP optimizations

a-huk:
## Summary

- **NoPE tail-score fix** — saves pre-RoPE K/Q into persistent buffers and uses them for the tail-score attention instead of post-RoPE tensors. This eliminates the RoPE distance decay that was burying early-position chunks, fixing NIAH at 8K–128K.
- **GPU RMSNorm kernel** (`rms_norm_hip.cu`) — replaces the GPU→CPU→GPU round-trip for B_norm across all 28 drafter layers with a single fused HIP kernel (one block/token, 256 threads, warp+smem reduction). Saves ~0.3s × 28 layers at 8K.
- **HIP drafter BF16 sparse path** — routes the drafter through the BF16 rocWMMA kernel when `DFLASH27B_HAVE_FLASHPREFILL` is defined; fixes the forced-chunk cap so BSA selects the right number of blocks; merges the B-step FFN into a single graph to reduce launch overhead.
- **HIP P1/P2 kernel bug fixes** — tail-bound fix for the P1 sliding-window kernel; BF16 rounding fix for the P2 sparse FA kernel.
- **GEMM M_gemm OOB pad** — allocates `mean_Q`/`mean_K` padded to the next multiple of 16 to prevent GPU faults when M < 16 (rocWMMA loads 16×16 tiles unconditionally).
- **AMD UMA VRAM detection** — daemon boot-readiness check uses `hipMemGetInfo` delta on AMD unified-memory systems instead of nvidia-smi.
- **GPU arch probe + ddtree-budget advisory** — probes `rocm_agent_enumerator` at startup and prints a budget tuning hint based on measured data (gfx1100 benefits from `budget=8`, gfx1151/gfx1201 prefer the default `budget=22`).
- **`--no-thinking` server guard** — server-level flag to strip `<think>` blocks from output.

## Benchmark (Strix Halo, Ryzen AI Max+ 395 / gfx1151, 128 GB UMA)

End-to-end compress time, `keep_ratio=0.05`, warmup + 3 timed runs, median.
Baseline = upstream/main + HIP shims, `SM80_EQUIV=OFF` (q8 ggml-FA, no rocWMMA).

| Tokens | Baseline (q8-FA) | This PR (rocWMMA+all) | Speedup |
|-------:|:----------------:|:---------------------:|:-------:|
| 4K     | 1.280s           | 0.800s                | **1.60×** |
| 8K     | 3.220s           | 1.590s                | **2.03×** |
| 16K    | 9.760s           | 3.380s                | **2.89×** |
| 32K    | 33.070s          | 7.390s                | **4.47×** |
| 64K    | 120.700s         | 16.800s               | **7.18×** |
| 128K   | 471.580s         | 39.260s               | **12.01×** |

**Geometric mean speedup: 3.92×.** Speedup is superlinear — BSA sparsity + KV-transpose coalescing compound at large context while the q8 fallback stays O(S²).

### DIFF
--- a/dflash/CMakeLists.txt
+++ b/dflash/CMakeLists.txt
@@ -273,6 +273,9 @@ if(DFLASH27B_GPU_BACKEND STREQUAL "cuda")
 elseif(DFLASH27B_GPU_BACKEND STREQUAL "hip")
     set_target_properties(dflash27b PROPERTIES HIP_ARCHITECTURES "${_dflash27b_archs}")
     target_compile_definitions(dflash27b PRIVATE DFLASH27B_BACKEND_HIP=1 GGML_USE_HIP)
+    # hip_compat shim is needed by ALL dflash27b sources (peer_access.cpp,
+    # dflash_feature_ring.cpp, flashprefill.cpp), not just the SM80_EQUIV path.
+    target_include_directories(dflash27b PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/hip_compat)
 endif()
 
 # FlashPrefill custom kernels.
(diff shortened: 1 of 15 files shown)


howard0su:
can you split the changes into multi PRs? it will make code review and merge much easier.

davide221:
@a-huk great first contribution! Thank you. Tested and verified on our strix halo.

PR #230, opened 2026-05-20
fix(dflash): Detect bare function tool calls in streaming

jkyamog:
Fixes Python `server.py` streaming tool-call handling for model output that starts with bare Qwen-style function XML instead of a `<tool_call>` wrapper.

Observed real output:

```xml
<function=read_file>
    <parameter=path>
    ~/Projects/gpu-cpu-hybrid/notes.md
    </parameter>
</function>
```

Current upstream can parse this shape in `parse_tool_calls()`, but the streaming state machine only switches into tool-buffer mode when it sees `<tool_call>`. As a result, bare `<function=...>` output can leak to the user as assistant text instead of being returned as structured OpenAI `delta.tool_calls`.

## Root Cause

The parser supports multiple tool-call shapes:

- `<tool_call>...</tool_call>`
- bare `<function=...>...</function>`
- `<tool_code>{...}</tool_code>`
- JSON tool-call objects

But the streaming detector only looked for:

```text
<tool_call>
```

So streaming did not route bare `<function=` or `<tool_code>` output into the parser.

## Fix

The streaming state machines now treat these as tool-call starts when request `tools` are present:

- `<tool_call>`
- `<function=`
- `<tool_code>`

Detection remains disabled when no tools were supplied, so ordinary XML-like assistant text is not accidentally converted into a tool call.

## Test Plan

Added regression coverage for:

- Chat Completions streaming bare `<function=...>` tool call
- Responses streaming bare `<function=...>` tool call
- existing raw tool-call replay test now includes request `tools`, matching the OpenAI contract

### DIFF
--- a/dflash/scripts/server.py
+++ b/dflash/scripts/server.py
@@ -182,6 +182,7 @@ FUNCTION_SIGNATURE_RE = re.compile(
     r"<function=([A-Za-z_][\w.-]*)\((.*?)\)</function>", re.DOTALL)
 TOOL_CODE_RE = re.compile(r"<tool_code>(.*?)</tool_code>", re.DOTALL)
 TOOL_OPEN_TAG = "<tool_call>"
+TOOL_START_TAGS = (TOOL_OPEN_TAG, "<function=", "<tool_code>")
 THINK_OPEN_TAG = "<think>"
 THINK_CLOSE_TAG = "</think>"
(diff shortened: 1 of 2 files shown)


jxnding:
this fixed tool calls in opencode for me! Thanks @jkyamog

Issue #216, opened 2026-05-18
Prefix-cache RESTORE can replay prior completion across requests


easel:
## Summary

Prefix-cache `RESTORE` must reuse only the cached prompt-prefix state and prefill the current request suffix. In the failing run, restore-hit requests sometimes produced output from the wrong request: plain prompts echoed the current user prompt instead of answering, and tool prompts replayed a previous tool call.

This is a correctness bug, not a scoring/model-quality issue. The raw traces show the wrong output after a prefix-cache hit.

## Observed Environment

- Branch under test: `integration/props-uv-squared`
- Server: `lucebox` on `http://127.0.0.1:1236`
- Model: `Qwen3.6-27B-Q4_K_M.gguf`
- GPU: RTX 3090 Ti, 24 GB
- `/props` runtime: `max_ctx=114688`, `ddtree_budget=22`, `lazy_draft=true`, `kv_cache_k/v=tq3_0`
- Prefix cache: `capacity=1`

## Repro 1: Same Prefix, Different Graded Prompts

Run a sequence of non-streaming prompts that share a system prefix:

```bash
python3 dflash/scripts/bench_http_capability.py \
  --url http://127.0.0.1:1236 \
  --questions 4 \
  --max-tokens 256 \
  --timeout 300 \
  --json-out /tmp/lucebox-capability-visible.json \
  --trace /tmp/lucebox-capability-visible-trace.txt
```

Server log shows the first request creating a snapshot and the next request restoring it:

```text
[pc] inline-snap committed slot=0 prefix_len=28
[pc] lookup hit slot=0 prefix_len=28 (of 89 total)
[daemon] ok N=89 gen=53 prefix_len=28 (RESTORE slot=0) stream_fd=6
```

The restore-hit request returned the prompt text instead of an answer:

```text
MODEL_OUTPUT_BEGIN
user
Find the least non-negative residue of 12345 modulo 97.

Solve the problem. At the end, write exactly one final line in this format and do not write anything after it:
Answer: <integer>
MODEL_OUTPUT_END
```

## Repro 2: Same Prefix, Different Tool Prompts

Run the agentic tool-call smoke:

```bash
python3 dflash/scripts/bench_agentic_tools.py \
  --url http://127.0.0.1:1236 \
  --repeat 1 \
  --timeout 300 \
  --json-out /tmp/lucebox-agentic-after-restart.json
```

Observed:

```text
PASS read-props-driver  expected=read_file got=['read_file']
[...]

Issue #167, opened 2026-05-13
Investigate streaming prompt-file lifetime race causing empty responses


jkyamog:
During streamed Chat Completions requests, dflash can sometimes return an empty assistant response after tool-call or follow-up turns, especially in Hermes/OpenClaw style agent loops.

  Observed logs include patterns like:
```
  [pc] lookup hit slot=0 prefix_len=14652 (of 14924 total)
  empty prompt
  [daemon] [snap] restored slot=0 cur_pos=14652
  WARNING dflash.server: 0 output tokens — aborted snapshot reservation
  chat DONE ... out=0 ... finish=stop
```
  The suspected issue is a race between Python streaming cleanup and the dflash daemon reading the prompt .bin file. In streaming mode, Python may unlink the prompt file once the HTTP stream finishes, but the daemon side may still need that
  file for in-flight work. If the daemon opens/reads after cleanup, it can see an empty/missing prompt and produce zero output tokens.

  [fix-streaming-prompt-files](https://github.com/Luce-Org/lucebox-hub/compare/main...jkyamog:lucebox-hub:fix-streaming-prompt-files) is a temporary guard: retain streaming prompt files instead of deleting them at stream finalization. This avoids the suspected race, but it is not the final design.

  Expected final fix: make prompt ownership/lifetime explicit so streaming cleanup cannot race daemon prompt reads, while still avoiding unbounded temp-file growth.

davide221:
@jkyamog thanks for digging into this bug, did you find the root cause?

jkyamog:
> [@jkyamog](https://github.com/jkyamog) thanks for digging into this bug, did you find the root cause?

not really, the empty reply doesn't happen often after my first patch https://github.com/Luce-Org/lucebox-hub/pull/145/changes/a86b1cd5753286e73994ede60af707adef6d0dff

But I think there is a race condition somewhere, so one https://github.com/Luce-Org/lucebox-hub/compare/main...jkyamog:lucebox-hub:fix-streaming-prompt-files it just to turn off the cache clean up.  Since I use docker, after I restart the container to clean up /tmp.

[...]

PR #231, opened 2026-05-20
fix(dflash): abandoned stream drain

jkyamog:
Fixes a Python `server.py` stream-synchronization bug where an SSE client can disconnect before the dflash daemon has emitted its `-1` sentinel. In that case the daemon continues writing tokens to the shared `r_pipe`, but the Python streaming generator exits and releases `daemon_lock`. The next request can then consume stale tokens from the abandoned request, producing impossible timings and wrong responses.

This PR keeps the daemon protocol synchronized by draining abandoned streams to the sentinel while still holding `daemon_lock`.

## Observed Failure

From a real Hermes run:

```text
05:58:06 INFO dflash.server: chat chatcmpl-eb00d86f14f14b9bb4837826  stream=True ...
[pc] lookup hit slot=0 prefix_len=52622 (of 53215 total)
05:58:59 WARNING dflash.server: stream ended before daemon sentinel; retaining prompt .bin for in-flight daemon read
  [daemon] ok N=53215 gen=247 prefix_len=53024 (RESTORE slot=0) stream_fd=5
```

After that, later requests showed impossible near-zero wall time and consumed output that belonged to earlier daemon work:

```text
05:59:17 INFO dflash.server: chat chatcmpl-affa6ec8564f4ef4b6b1de92  stream=True ...
05:59:17 WARNING dflash.server: inline snapshot ack missing - dropped slot reservation
05:59:17 INFO dflash.server: chat DONE chatcmpl-affa6ec8564f4ef4b6b1de92  in=53217 out=235  0.1s  2170.2 tok/s ...
```

That is not realistic model performance. It is stale token-pipe data from a previously abandoned stream.

## Root Cause

Streaming requests share one daemon token pipe (`r_pipe`) protected by `daemon_lock`.

Normal path:

1. Python sends a daemon command.
2. Python reads token IDs from `r_pipe`.
3. Daemon emits `-1`.
4. Python marks `timing["daemon_done"] = True`.
5. Python releases `daemon_lock`.

Failure path:

1. Client disconnects or cancels the SSE response.
2. Python streaming generator exits before seeing daemon `-1`.
3. Daemon keeps generating and writing token IDs to `r_pipe`.
4. Python releases `daemon_lock`.
5. Next request starts and reads stale token IDs from the previous request.

## Fix

When a streaming response exits before `timing["daemon_done"]`:

- log the existing warning
- drain `r_pipe` to the daemon sentinel using `_drain_until_sentinel`
- keep this drain shielded from cancellation so the server reaches a clean protocol boundary
- only then allow the request to release `daemon_lock`

This does not cancel in-flight GPU work. It is a synchronization fix: the next request starts from a clean daemon pipe instead of inheriting stale output.

## Logging And Timing

The abandoned-stream log should stay at warning level, not debug. A client disconnect is normal, but the server reaching the end of an SSE generator before the daemon sentinel is a protocol synchronization hazard: if it is not handled, the next request can read stale token IDs from the previous request.

This PR does not change timing math and does not count prefix-cache time differently. The low/impossible timings disappear because the next request no longer consumes already-buffered tokens from an abandoned prior stream. Normal cached requests should still report fast prefill when a prefix snapshot is restored, but they should not show impossible decode rates caused by stale pipe data.

The drain log includes how many stale daemon tokens were consumed during cleanup, which is enough evidence for this specific synchronization fix without adding broader empty-response diagnostics to this PR.

## Relationship To Existing Issues

Related issue:

- #167: Investigate streaming prompt-file lifetime race causing empty responses

Related but distinct issue:

- #216: Prefix-cache RESTORE can replay prior completion across requests

#216 fixed RESTORE prompt-delta correctness. This PR addresses a different stale-output path: abandoned streaming responses leaving unread daemon tokens in `r_pipe`.

### DIFF
--- a/dflash/scripts/server.py
+++ b/dflash/scripts/server.py
@@ -1284,6 +1284,37 @@ def build_app(target: Path, draft: Path | None, bin_path: Path, budget: int, max
     def _max_tokens_for(req) -> int:
         return getattr(req, "max_completion_tokens", None) or req.max_tokens
 
+    async def _drain_abandoned_stream(kind: str, request_id: str,
+                                      timing: dict) -> None:
+        """Synchronize the daemon stream after an abandoned SSE response.
+
+        If the client disconnects while the daemon is still generating, the
+        daemon continues writing token IDs to r_pipe. Releasing daemon_lock
+        before consuming the sentinel lets the next request read stale tokens.
+        Drain to -1 here so the next request starts from a clean pipe.
+        """
+        if timing.get("daemon_done"):
+            return
+        loop = asyncio.get_running_loop()
+        drain_fut = loop.run_in_executor(None, _drain_until_sentinel, r_pipe)
+        cancelled = False
+        while True:
+            try:
+                drained = await asyncio.shield(drain_fut)
+                break
+            except asyncio.CancelledError:
+                cancelled = True
+                log.warning(
+                    "%s stream cancellation deferred until daemon sentinel %s",
+                    kind, request_id)
+        timing["daemon_done"] = True
+        timing["abandoned_stream_drained_tokens"] = len(drained)
+        log.warning(
+            "%s stream abandoned %s: drained %d stale daemon tokens",
+            kind, request_id, len(drained))
+        if cancelled:
+            raise asyncio.CancelledError
+
     # ── /v1/chat/completions ────────────────────────────────────────────────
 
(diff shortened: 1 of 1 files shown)


jkyamog:
CAVEAT: Also I only tested this behavior in deployment together with PR #230.

PR #232, opened 2026-05-20
 Gemma4: Full DFlash Integration (Speculative Decode + BSA Prefill + Prefix Cache)

howard0su:
Adds complete Gemma4 support to the DFlash inference pipeline — speculative decode, BSA sparse-FA prefill, SWA ring-buffer KV cache, and prefix cache with correct feature snapshot/restore.

Key Changes

Speculative Decode (DFlash target)

 - Implement Gemma4DFlashTarget adapter (embed, verify_batch, project_hidden)
 - Wire spec-decode into Gemma4Backend with feature mirror sync
 - Replace snapshot/replay with KV truncation for ~2× fewer forward passes per spec step
 - Fix SWA causal masking and rope_theta for draft model accuracy

BSA Sparse-FA Prefill

 - Implement gemma4_prefill_bsa() — per-layer BSA for SWA layers, dense FA for full-attention layers
 - Add unified flash_prefill_forward() dispatch (BF16→F16→Q8) in flashprefill.h
 - Simplify Qwen3's prefill dispatch to use the same unified function (−68 lines)

Prefix Cache

 - Add snapshot_save calls in generate()/restore_and_generate() (was completely missing — cache hits returned 0 tokens)
 - Save/restore target_feat + last_tok in snapshots (matching Qwen35) to maintain spec-decode acceptance rate after restore
 - Full feature mirror resync after restore to ensure draft model consistency
 - Add Gemma family detection for chat-turn boundary markers

SWA Ring-Buffer & Architecture

 - Implement G5 SWA ring-buffer indexing with proper causal masks
 - Add fa_window support for full-attention layers (G6)
 - Handle variable head_dim (128 SWA / 256 full), per-layer n_head_kv, KV sharing
 - Fix attention scale, tokenizer decode, and server integration

Misc

 - Fix draft loader to read rope_theta from GGUF (remove hardcoded constant)
 - Rename draft_dflash_graph.cpp → draft_graph.cpp
 - Add graph_compute error checking in BSA prefill path


dusterbloom:
Thanks for the PR @howard0su . 
I tested it locally on `gemma-4-31B-it-Q4_K_M` and `gemma-4-26B-A4B-it-UD-Q4_K_M`.

Gemma4 never actually enters the BSA sparse-FA path right now. Every layer falls through the guard and uses dense WMMA instead. 

Is that expected for this PR? 

In case it is, no worries. It can be added later. 

Only if needed here something to consider:
`dflash/src/gemma4/gemma4_graph.cpp:549`  does `abs_k % swa_size` with no guard. 

`Gemma4Cache::swa_size` default-initialises to 0 (`gemma4_internal.h:161`); current call paths through `create_gemma4_cache` are safe (`swa_size` falls back to `max_ctx`), but any future path that constructs a `Gemma4Cache` outside that helper, or a GGUF where `sliding_window` reads as 0 while `swa_layers[il]` is true, would `SIGFPE` rather than fail cleanly.

Cheap fix:

```
GGML_ASSERT(swa_size > 0 && "SWA branch entered with uninitialised cache.swa_size");
      const int slot = abs_k % swa_size;
```

Both findings reproduced against PR head `de0e4c1`.

howard0su:
good catch. I will check it. I only have a 2080ti which doesn't support BSA. Will try to find a env to debug.

PR #234, opened 2026-05-20
feat(dflash): add default chat thinking toggle

jkyamog:
Adds a server-level opt-in for enabling Qwen thinking mode by default for Chat Completions, while preserving the current default of no-thinking mode.

This is useful for clients such as OpenWebUI that may not forward nested request fields like:

```json
{
  "chat_template_kwargs": {
    "enable_thinking": true
  }
}
```

## Configuration

Enable by environment variable:

```bash
DFLASH_ENABLE_THINKING=1
```

Or by CLI:

```bash
python3 scripts/server.py ... --enable-thinking
python3 scripts/server.py ... --enable-thinking=true
```

Explicit disable is also accepted:

```bash
python3 scripts/server.py ... --enable-thinking=false
```

Per-request `chat_template_kwargs.enable_thinking` still overrides the server default.

## Notes

This controls the Qwen chat-template thinking mode and whether generated `<think>...</think>` output is exposed as `reasoning_content`. It does not guarantee the model will never write reasoning-like prose in normal visible `content`; that can still happen if the model is prompted that way.

### DIFF
--- a/dflash/scripts/server.py
+++ b/dflash/scripts/server.py
@@ -112,6 +112,14 @@ def _print_hip_budget_advisory(budget: int) -> None:
         print(f"  [hip] {arch}: no advisory; using budget={budget}", flush=True)
 
 
+def _parse_bool(value: str | bool | None) -> bool:
+    if isinstance(value, bool):
+        return value
+    if value is None:
+        return False
+    return str(value).strip().lower() in {"1", "true", "yes", "on"}
+
+
 def _extra_daemon_has_target_sharding(extra: list[str] | None) -> bool:
     """True if we spawn test_dflash with multi-GPU target layer split."""
     if not extra:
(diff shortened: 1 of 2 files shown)


davide221:
@jkyamog thanks for the contribution, we will soon move to the cpp server, if you want to start experimenting with that one

jkyamog:
> @jkyamog thanks for the contribution, we will soon move to the cpp server, if you want to start experimenting with that one

I just tried I am pretty impressed, a lot lower VRAM so I might be able to run 128k context.  The tool call are still an issue, so I just need to port some of the python code that was done recently.  I think codex has done some of it already, I just need to test and review... for I forgot about it and too busy juggling some many things.  I am using Qwen 3.5 vision on ik_llama.cpp, I am hoping Lucebox will support vision in the future.

PR #236, opened 2026-05-20
feat(server): add C++ backend placement foundation

weicj:
# Summary

This PR adds the C++ backend placement foundation for the native `dflash_server` path, including the minimal server-side plumbing needed to expose and validate placement options.

Before this change, backend placement wiring mainly lived in Python bench/launcher helpers. Those helpers are still useful for bench execution, but the native C++ server also needs its own small placement structure for backend selection, device selection, logging, and unsupported-combination errors.

This PR does not implement CUDA/HIP mixed execution or native target layer split yet. It prepares the native server/runtime surface those follow-up PRs will use.

# Changes

1. Add a dedicated C++ backend placement directory

   New files:

   - `dflash/src/placement/placement_backend.h`
   - `dflash/src/placement/placement_config.h`

   `placement_backend.h` defines `PlacementBackend` (`auto`, `cuda`, `hip`) plus helpers for parsing, logging, compiled-backend detection, and support checks.

   `placement_config.h` owns `DevicePlacement`, including target backend, primary GPU, target layer-split GPU list, split weights, peer-access flag, and max context.

   This is intentionally a directory instead of one larger `common` header. Later backend placement work can add focused files for parsing, validation, runtime application, memory policy, or placement recommendation without growing `server_main.cpp` or `common/device_placement.h`.

2. Keep existing includes compatible

   `dflash/src/common/device_placement.h` is kept as a compatibility include that forwards to `placement/placement_config.h`.

   Existing model/backend code is updated to include the new placement header directly where practical, without turning this PR into a broad include cleanup.

3. Carry draft placement through the C++ backend boundary

   `dflash/src/common/backend_factory.h` now carries a `draft_device` placement next to the target `device` placement.

   This keeps target and draft placement explicit at the native server/backend factory boundary without splitting backend and GPU id into separate public concepts.

4. Add native server backend-device CLI plumbing

   `dflash/src/server/server_main.cpp` now parses:

   - `--target-device <backend:gpu>` such as `cuda:0`, `hip:0`, or `auto:0`
   - `--draft-device <backend:gpu>`
   - `--target-devices <backend:gpu,...>` for the reserved target layer-split surface
   - `--target-layer-split <weights>`
   - `--peer-access`

   The previous split CLI shape (`--target-backend`, `--draft-backend`, `--gpu`, `--draft-gpu`, `--target-gpus`) is removed from the native server surface so backend and device id are expressed as one placement value from the start.

5. Add backend placement logs and clear unsupported errors

   The native server config block now prints target device, draft device, optional target shard devices, and peer-access state.

   `--peer-access` is bridged into the existing runtime peer-access opt-in path before backend creation.

   Unsupported combinations are rejected explicitly instead of being silently ignored:

   - backend different from the compiled backend
   - mixed target/draft backend before mixed execution is implemented
   - target layer split before native server layer-split support is implemented

# Follow-up Path

This PR intentionally includes the first server-side backend placement plumbing. Follow-up feature PRs can build on this native surface instead of adding more ad hoc parsing in `server_main.cpp`.

For native `dflash_server` behavior, C++ backend placement should become the source of truth for accepted backend placement options and explicit unsupported combinations. Existing Python backend placement helpers remain the orchestration layer for bench/launcher flows, visible-device/env setup, and `test_dflash` daemon arguments until native server support catches up.

### DIFF
--- a/dflash/src/server/server_main.cpp
+++ b/dflash/src/server/server_main.cpp
@@ -9,11 +9,12 @@
 // Usage:
 //   dflash_server <model.gguf> [--draft <draft.gguf>] [--port 8080]
 //                              [--host 0.0.0.0] [--max-ctx 131072]
-//                              [--max-tokens 4096] [--gpu 0]
+//                              [--max-tokens 4096] [--target-device auto:0]
 
 #include "http_server.h"
 #include "common/backend_factory.h"
 #include "common/gguf_inspect.h"
+#include "common/peer_access.h"
 
 #include <cstdio>
 #include <cstdlib>
@@ -21,6 +22,7 @@
 #include <csignal>
 #include <memory>
 #include <string>
+#include <vector>
 
 using namespace dflash::common;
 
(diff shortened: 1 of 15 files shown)


dflash/src/server/server_main.cpp
```diff
@@ -35,6 +119,11 @@ static void print_usage(const char * prog) {
         "  --max-tokens <N>     Default max output tokens (default: 4096)\n"
         "  --gpu <N>            Target GPU device (default: 0)\n"
         "  --draft-gpu <N>      Draft GPU device (default: 0)\n"
+        "  --target-backend <auto|cuda|hip>  Target backend (default: auto)\n"
+        "  --draft-backend <auto|cuda|hip>   Draft backend (default: auto)\n"
+        "  --target-gpus <ids>  Reserved target layer-split GPU list (comma-separated)\n"
```
howard0su: shall we consider to merge targetbackend and target_gpu? named target-device=cuda:0? and draft-device=hip:0?

weicj: yes i think we should do so, let me unify it

PR #237, opened 2026-05-20
feat(dFlash): MTP foundation

dusterbloom:
## Summary

Adds Qwen3.5/3.6 NextN MTP support: `IMtpModule` interface + chain runner in `common/`, `Qwen35MtpModule` concrete impl, qwen35 backend wiring, and `--mtp-gguf` flags in `dflash_server`.

**Works with both GGUF layouts**:
- **Separate files**: backbone GGUF as primary + MTP-head GGUF via `--mtp-gguf`
- **Unsloth single-file combined GGUF** (`Qwen3.6-27B-Q4_K_M-mtp.gguf`): pass the same file as both the primary and `--mtp-gguf`; the loader reads backbone tensors and MTP head tensors from it. Verified greedy-equivalent + same 1.82× speedup as the separate-files setup.

### Single-prompt HTTP (`/v1/chat/completions`, temp=0)

HumanEval-10 via `harness/benchmarks/generation_benchmark.py` on Qwen3.6-27B Q4_K_M + MTP-Q4_K_M (RTX 3090):

| Mode      | mean tok/s | speedup | accept rate |
|-----------|-----------:|--------:|------------:|
| AR        | 32.46      | 1.00x   | —           |
| chain γ=2 | 59.24      | 1.82x   | 0.93–1.00   |

All 10 prompts pass functional checks. Greedy text byte-identical on 8/10; remaining 2 within the AR-vs-AR non-determinism baseline (verified via control run: same prompts produced 9/10 identical text across two AR-only runs of the same server).

Confirmed equivalent between separate-files and unsloth single-file layouts: 1.82× → 1.82× (statistically identical), 8/10 byte-identical text in both, same 95–100% accept rate range.

### Agentic-CLI smoke (OpenAI Responses API)

`pi` (v0.55.3) via `harness/clients/run_pi.sh`: server starts with MTP, SSE streaming delivers tokens, `[mtp_decode] iters=2 proposed=4 accepted=4 accept_rate=1.00`, marker received exact-match. Proves the full HTTP → Responses API → SSE → MTP chain → tokenizer → client round-trip.

### Long-prompt resilience

25K-token prompt under `--mtp-gguf` + `--max-ctx 65536`: server stays alive, MTP fires (iters=16, 50% accept), real response returned. Earlier silent SIGKILL on long prompts (DFLASH27B_MTP_CTX default 8192 not tracking max_ctx) fixed in \`ada7bb0\`; `warm_head_kv` overflow now triggers graceful AR fallback.

### Unit tests

`test_common_mtp_orchestrator`: **20/20 pass** — chain runner state machine (γ propagation, EOS termination, partial-accept rollback, n_gen termination), orchestrator lifecycle, Qwen35MtpModule input-validation guards, T20 ExternalDrafter partial-accept hidden threading (asserts the `set_capture_row` + `consume_captured_hidden` invariant the \`74f708a\` fix added). Mock-based, no GGUF required.

## Scope

- `IMtpModule` + `INativeMtp` + `IExternalDrafterMtp` mixins
- `Qwen35MtpModule` (`INativeMtp` impl) + GGUF loader + step graph
- `MtpChainRunner` + `MtpOrchestrator`
- `dflash_server`: `--mtp-gguf`, `--mtp-gamma`, `--mtp-draft-source`, `--mtp-draft-topk`

## Not in this PR (follow-ups)

- Gemma4 ExternalDrafter MTP module (uses the `IExternalDrafterMtp` interface declared here)
- Per-request DFlash/MTP/auto dispatcher (commit \`3b65ac1\` on \`feat/mtp-prefix-warm-ghost\`)
- `IDrafter` abstraction unifying DFlash + MTP under one runner
- Tree-mode runner + entropy-adaptive selector
- Prefix-cache WARM hit for MTP

### DIFF
--- a/dflash/src/common/backend_factory.h
+++ b/dflash/src/common/backend_factory.h
@@ -21,6 +21,7 @@ namespace dflash::common {
 // ─── MTP source selection ────────────────────────────────────────────────
 // Replaces the old free-form mtp_draft_source string (@howard0su #237, line 59).
 enum class MtpSource {
+    Unset,            // internal sentinel: --mtp-source not provided (never escapes arg parsing)
     None,             // no MTP speculator
     Native,           // MTP heads co-located in the target GGUF (e.g. unsloth single-file)
     ExternalDrafter,  // separate MTP-head GGUF supplied via mtp_gguf_path
@@ -63,13 +64,15 @@ struct BackendArgs {
 
     // MTP (Multi-Token Prediction) speculator — mutually exclusive with --draft.
     // mtp_source drives which loading path is taken:
+    //   Unset          → internal default; --mtp-source not provided; resolved to None after
+    //                    legacy-flag inference (never reaches the backend factory as Unset).
     //   None           → MTP disabled; mtp_gguf_path ignored.
     //   Native         → MTP heads embedded in model_path GGUF (single-file, e.g. unsloth).
     //                    mtp_gguf_path is left nullptr; the factory sets it to model_path.
     //   ExternalDrafter→ Separate MTP-head GGUF at mtp_gguf_path (required).
     //   Auto           → factory calls gguf_contains_mtp_tensors(model_path): if true,
     //                    resolves to Native; otherwise resolves to None.
-    MtpSource    mtp_source       = MtpSource::None;
+    MtpSource    mtp_source       = MtpSource::Unset;
     const char * mtp_gguf_path    = nullptr;  // required only for ExternalDrafter
     int          mtp_gamma        = 0;        // 0 = MTP loaded but not active; >0 = chain depth
     bool         mtp_use_topk     = false;    // false = chain (default), true = mtp_topk strategy
--- a/dflash/src/common/backend_factory.h
+++ b/dflash/src/common/backend_factory.h
@@ -51,6 +51,13 @@ struct BackendArgs {
     float           ddtree_temp      = 1.0f;
     bool            ddtree_chain_seed = true;
     bool            use_feature_mirror = false;
+
+    // MTP (Multi-Token Prediction) speculator — mutually exclusive with --draft.
+    // When mtp_gguf_path is set, the backend ignores draft_path.
+    const char * mtp_gguf_path    = nullptr;
+    int          mtp_gamma        = 0;        // 0 = MTP loaded but not active; >0 = chain depth
+    const char * mtp_draft_source = nullptr;  // "chain" (default) | "mtp_topk"
+    int          mtp_draft_topk   = 1;
 };
 
 // ─── Factory function ───────────────────────────────────────────────────
--- a/dflash/src/common/backend_factory.h
+++ b/dflash/src/common/backend_factory.h
@@ -18,6 +18,15 @@
 
 namespace dflash::common {
 
+// ─── MTP source selection ────────────────────────────────────────────────
+// Replaces the old free-form mtp_draft_source string (@howard0su #237, line 59).
+enum class MtpSource {
+    None,             // no MTP speculator
+    Native,           // MTP heads co-located in the target GGUF (e.g. unsloth single-file)
+    ExternalDrafter,  // separate MTP-head GGUF supplied via mtp_gguf_path
+    Auto,             // probe target GGUF for nextn_predict_layers; Native if found, else None
+};
+
 // ─── Backend creation arguments ─────────────────────────────────────────
 // A superset of all per-arch config fields. The factory reads only those
 // relevant to the detected arch; unused fields are silently ignored.
@@ -53,20 +62,35 @@ struct BackendArgs {
     bool            use_feature_mirror = false;
 
     // MTP (Multi-Token Prediction) speculator — mutually exclusive with --draft.
-    // When mtp_gguf_path is set, the backend ignores draft_path.
-    const char * mtp_gguf_path    = nullptr;
+    // mtp_source drives which loading path is taken:
+    //   None           → MTP disabled; mtp_gguf_path ignored.
+    //   Native         → MTP heads embedded in model_path GGUF (single-file, e.g. unsloth).
+    //                    mtp_gguf_path is left nullptr; the factory sets it to model_path.
+    //   ExternalDrafter→ Separate MTP-head GGUF at mtp_gguf_path (required).
+    //   Auto           → factory calls gguf_contains_mtp_tensors(model_path): if true,
+    //                    resolves to Native; otherwise resolves to None.
+    MtpSource    mtp_source       = MtpSource::None;
+    const char * mtp_gguf_path    = nullptr;  // required only for ExternalDrafter
     int          mtp_gamma        = 0;        // 0 = MTP loaded but not active; >0 = chain depth
-    const char * mtp_draft_source = nullptr;  // "chain" (default) | "mtp_topk"
+    bool         mtp_use_topk     = false;    // false = chain (default), true = mtp_topk strategy
     int          mtp_draft_topk   = 1;
 };
 
 // ─── Factory function ───────────────────────────────────────────────────
 // Inspects model_path GGUF metadata, constructs the correct backend, and
 // calls init(). Returns nullptr on failure (diagnostic printed to stderr).
+// When args.mtp_source == Auto, resolves to Native or None before
+// constructing; the resolved value is not written back into args.
 std::unique_ptr<ModelBackend> create_backend(const BackendArgs & args);
 
 // Returns the detected architecture string without creating a backend.
 // Useful for early dispatch (e.g. printing which backend will be used).
 std::string detect_arch(const char * model_path);
+
+// Returns true if the GGUF at `path` contains MTP-head tensors.
+// Heuristic: presence of `qwen35.nextn_predict_layers` metadata key with
+// a value > 0. Pure metadata scan — no tensor allocation, no GPU touch.
+// Used by create_backend() when mtp_source == Auto.
+bool gguf_contains_mtp_tensors(const std::string & path);
 
 }  // namespace dflash::common
--- a/dflash/src/common/model_backend.h
+++ b/dflash/src/common/model_backend.h
@@ -25,6 +25,10 @@ namespace dflash::common {
 // Return true to continue generation, false to abort.
 using TokenCallback = std::function<bool(int32_t token)>;
 
+// Forward-declare MTP module so the optional mtp() accessor below
+// doesn't pull mtp_interface.h into every backend.
+namespace mtp { struct IMtpModule; }
+
 // ─── I/O handle passed to backend methods that need protocol output ─────
 struct DaemonIO {
     int stream_fd = -1;
@@ -178,6 +182,19 @@ struct ModelBackend {
     // Return the DFlashTarget adapter for this backend. Only valid when
     // supports_dflash_spec_decode() returns true. Default returns nullptr.
     virtual class DFlashTarget * dflash_target() { return nullptr; }
+
+    // ── MTP speculative decode support ───────────────────────────────
+    // Returns true if this backend has an MTP module bolted on (Gemma4's
+    // external drafter, Qwen3.6's native heads, or any future MTP impl).
+    // The MTP path is orthogonal to DFlash: both may be supported by the
+    // same backend (caller picks via request method).
+    virtual bool supports_mtp() const { return false; }
+
+    // Return the IMtpModule adapter for this backend. Only valid when
+    // supports_mtp() returns true. Default returns nullptr.
+    // Forward-declared to avoid a header dependency from model_backend.h
+    // on mtp_interface.h — backends include both as needed.
+    virtual mtp::IMtpModule * mtp() { return nullptr; }
 
     // Release oversized scratch buffers between requests to prevent VRAM
     // growth over time. Default is a no-op.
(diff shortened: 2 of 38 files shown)


dusterbloom:
Addressed cubic review in commit 74f708a:
- F1 (P2): add missing io.emit(-1) on chain-runner failure path in warm_and_decode (mtp_orchestrator.cpp:152)
- F2 (P1): call set_capture_row(accept_n) + consume_captured_hidden() before threading running_hidden on partial accept for ExternalDrafter flavor; NativeHeads path is unaffected (next_hidden is always empty for that flavor, new branch not entered)

Build clean. All 19 unit tests pass. T7 (partial_accept_rollback) exercises the NativeHeads rollback path only — the ExternalDrafter hidden-sync invariant is not covered by the existing suite (noted for follow-up).

E2E HumanEval-10 AR-vs-MTP same-text rate: 8/10 post-fix (unchanged from 8/10 pre-fix). F2 is confined to ExternalDrafter and does not affect the Qwen3.6/NativeHeads path tested here. he_01/he_02 divergences remain — consistent with prior AR-vs-AR control showing GPU non-determinism on those two prompts.

dusterbloom:
Coverage gap closed in bf2d3e9: T20 now exercises the ExternalDrafter partial-accept path that the 74f708a fix addressed (T7 used a NativeHeads stub, so it never reached the fixed branch). 20/20 unit tests pass.

dusterbloom:
Two more fixes from real-workload hermes testing (commit `ada7bb0`):

**A. Silent server death on long-prompt MTP**: `DFLASH27B_MTP_CTX` defaulted to 8192 but the server never propagated `--max-ctx` to that env var. With `--max-ctx 65536`, the backbone KV was allocated at 65536 slots while the MTP head_kv stayed at 8192 — VRAM mismatch caused OOM on the next prefill. (The `warm_head_kv` overflow guard already returned false cleanly — the bug was upstream at init time.) Fix: 5-line propagation in server_main.cpp using `setenv(..., overwrite=0)` so user env still wins.

**B. `--draft` + `--mtp-gguf` conflict**: server now logs a WARNING and clears `draft_path` when both are set (MTP wins).

Verified: 25K-token prompt → server stays alive, MTP ran (iters=16 proposed=32 accepted=16 accept_rate=0.50), real response returned. Conflict invocation prints the warning and starts cleanly with `draft = (none)` in the config banner.

dusterbloom:
Addressed cubic re-review in commit f6e8e94:

**P1**:
- gguf_mmap.h `open()` now idempotent (releases prior mapping first via placement-destroy+reconstruct; leaves object in default state on any failure path)
- gguf_mmap.h Windows `release()` now calls `CloseHandle()` before clearing `handle_` (fix by inspection — Linux build cannot exercise the Windows path)
- qwen35_mtp.cpp GPU warmup off-by-one (`>` -> `>=`): a prompt of exactly `n_ctx` tokens would have written slot `n_ctx` (out-of-bounds; valid range is `[0, n_ctx-1]` with `slot_start=1`). Now declines MTP cleanly and chain runner falls through to AR.

**P2**:
- test_dflash.cpp missing hidden_seq capture now clears `all_prefill_hidden` and breaks out of the prefill loop instead of running `warm_head_kv` on an undefined buffer
- T7 now asserts `restore_kv_at_chain_calls >= 1` — rollback path is confirmed to fire on partial accept

**P3**:
- T6 EOS termination assertion tightened from `total_emitted <= 10` to `total_emitted == 1` (EOS is the first emitted token; runner must stop immediately)

Build clean, 20/20 unit tests pass. HumanEval-10 re-bench: 8/10 byte-identical.

dflash/src/common/backend_factory.h
```diff
@@ -51,6 +51,13 @@ struct BackendArgs {
     float           ddtree_temp      = 1.0f;
     bool            ddtree_chain_seed = true;
     bool            use_feature_mirror = false;
+
+    // MTP (Multi-Token Prediction) speculator — mutually exclusive with --draft.
+    // When mtp_gguf_path is set, the backend ignores draft_path.
+    const char * mtp_gguf_path    = nullptr;
```
howard0su: shall we use https://huggingface.co/unsloth/Qwen3.6-27B-MTP-GGUF which contains MTP in the main model?

dflash/src/common/backend_factory.h
```diff
@@ -51,6 +51,13 @@ struct BackendArgs {
...
+    int          mtp_gamma        = 0;        // 0 = MTP loaded but not active; >0 = chain depth
+    const char * mtp_draft_source = nullptr;  // "chain" (default) | "mtp_topk"
```
howard0su: why not a enum?

dflash/src/common/gguf_mmap.h
```diff
@@ -0,0 +1,219 @@
+// common/gguf_mmap.h — RAII wrapper for platform-conditional mmap of GGUF files.
```
howard0su, line 1: this worth a separate PR

dflash/src/common/model_backend.h
```diff
@@ -148,6 +152,19 @@ struct ModelBackend {
...
+    // external drafter, Qwen3.6's native heads, or any future MTP impl).
+    // The MTP path is orthogonal to DFlash: both may be supported by the
+    // same backend (caller picks via request method).
+    virtual bool supports_mtp() const { return false; }
+
+    // Return the IMtpModule adapter for this backend. Only valid when
+    // supports_mtp() returns true. Default returns nullptr.
+    // Forward-declared to avoid a header dependency from model_backend.h
+    // on mtp_interface.h — backends include both as needed.
+    virtual mtp::IMtpModule * mtp() { return nullptr; }
```
howard0su, line 197: why not return nullptr indicate this model doesn't support mtp?

dflash/src/common/mtp_interface.h
```diff
@@ -0,0 +1,215 @@
...
+struct DFlashTarget;  // forward — see common/dflash_target.h
+
+namespace mtp {
+
+// ── Flavor tag ──────────────────────────────────────────────────────────
+// MtpChainRunner dispatches on this; concrete classes set it via the
+// matching mixin (IExternalDrafterMtp / INativeMtp).
+enum class MtpFlavor {
+    ExternalDrafter,   // Gemma4-style: separate drafter, h_prev chain
+    NativeHeads,       // Qwen3.6-style: MTP heads in the backbone
```
howard0su, line 36: this should hide into ModelBackend.

dflash/src/common/mtp_orchestrator.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+    std::vector<float> all_prefill_hidden((size_t)prompt_len * hidden);
+    int32_t last_tok = -1;
+
+    auto t_prefill0 = std::chrono::steady_clock::now();
+    for (int start = 0; start < prompt_len;) {
+        const int n = std::min(prefill_ubatch, prompt_len - start);
+        std::vector<int32_t> chunk(req.prompt.begin() + start,
+                                   req.prompt.begin() + start + n);
+        if (!target->verify_batch(chunk, start, last_tok, nullptr)) {
```
howard0su, line 72: mtp should only change how we generate the batch (v.s. draft), can we normalize the code with the existing logic?

dflash/src/qwen35/qwen35_mtp_graph.h
```diff
@@ -0,0 +1,133 @@
+// qwen35_mtp_graph.h — CUDA cgraph for Qwen3.6 MTP head step forward.
```
howard0su, line 1: we should use the existing graph API not a new API unless we want to put MTP header into another ggml backend.

dflash/src/qwen35/qwen35_mtp_loader.cpp
```diff
@@ -0,0 +1,225 @@
+// qwen35_mtp_loader.cpp — Discovery loader for Qwen3.6 -MTP-GGUF files.
```
howard0su, line 1: check the previous comment about GGUF loading.

howard0su (changes requested):

weicj:
I’m working on a shared hardware/backend/runtime placement foundation for the cpp native server path. https://github.com/Luce-Org/lucebox-hub/pull/236

Could we align the MTP backend/device selection with that shared placement layer, instead of keeping a separate MTP-specific selection path?

This would avoid duplicated CUDA/HIP/device handling across MTP,DFlash, PFlash, and future speculative paths.

dflash/src/qwen35/qwen35_daemon.cpp
```diff
@@ -30,6 +30,28 @@ int run_qwen35_daemon(const Qwen35DaemonArgs & args) {
     cfg.ddtree_temp        = args.ddtree_temp;
     cfg.ddtree_chain_seed  = args.ddtree_chain_seed;
     cfg.use_feature_mirror = args.use_feature_mirror;
+    cfg.mtp_gamma          = args.mtp_gamma;
+    cfg.mtp_use_topk       = args.mtp_use_topk;
+    cfg.mtp_draft_topk     = args.mtp_draft_topk;
+
+    // Resolve MtpSource to the mtp_gguf_path sentinel that Qwen35Backend expects.
+    switch (args.mtp_source) {
+        case MtpSource::Native:
```
howard0su, line 39: mtp_source can use mtp_gguf_path to determine. I don't see the value of this enum.

dflash/src/qwen35/qwen35_mtp_graph.cpp
```diff
@@ -0,0 +1,404 @@
...
+        return false;
+    }
+    if (!ggml_gallocr_alloc_graph(sg.alloc, sg.gf)) {
+        std::fprintf(stderr, "[qwen35_mtp_warm_graph] alloc_graph failed\n");
+        return false;
+    }
+    return true;
+}
+
+bool build_qwen35_mtp_step_graph(
```
howard0su, line 162: mtp is a normal transformer layer. please leverage the existing code in qwen35.

dflash/src/qwen35/gguf_target_loader.cpp
```diff
@@ -573,9 +591,39 @@ bool load_target_gguf_partial(const std::string & path,
...
+        if (gguf_get_tensor_type(gctx, tid) == GGML_TYPE_F32) {
+            const float * fp = (const float *)((const uint8_t *)mm.addr + off);
+            const size_t n_elem = sz / sizeof(float);
+            size_t n_nan = 0;
+            for (size_t i = 0; i < n_elem; i++) {
+                // Bit-level NaN test: any float whose exponent is all-1 and
+                // mantissa is non-zero. Faster + portable vs std::isnan.
+                uint32_t u;
+                std::memcpy(&u, fp + i, sizeof(u));
+                if ((u & 0x7F800000u) == 0x7F800000u && (u & 0x007FFFFFu) != 0u) n_nan++;
```
howard0su, line 603: why we do such check? we assume gguf is valid in other places.

dflash/src/server/server_main.cpp
```diff
@@ -138,6 +138,17 @@ static void print_usage(const char * prog) {
...
         "\n"
+        "MTP speculative decoding (mutually exclusive with --draft):\n"
+        "  --mtp-source <none|native|external|auto>\n"
+        "                              MTP source (default: auto when --mtp-gamma given)\n"
+        "                                none     = disable MTP\n"
+        "                                native   = MTP heads in target GGUF (unsloth single-file)\n"
+        "                                external = separate GGUF via --mtp-gguf\n"
+        "                                auto     = probe target GGUF; native if found, else none\n"
+        "  --mtp-gguf <path>           MTP GGUF path (required only for --mtp-source external)\n"
+        "  --mtp-gamma <int>           Speculation chain depth (default: 0 = disabled)\n"
```
howard0su, line 149: can we reuse ddtree-budget? In my metal model, MTP is a special type of draft model. it produce a <budget> tokens.

dflash/src/common/mtp_orchestrator.h
```diff
@@ -0,0 +1,44 @@
...
+#include "model_backend.h"
+#include "mtp_interface.h"
+
+namespace dflash::common {
+
+class DFlashTarget;
+
+namespace mtp {
+
+// Drive the full MTP warm + chain decode for one request.
```
howard0su, line 19: what does warm stands here? warm the cache?

dflash/src/qwen35/qwen35_mtp_graph.cpp
```diff
@@ -0,0 +1,404 @@
...
+#include "ggml-alloc.h"
+#include "ggml-backend.h"
+
+#include <cstdio>
+#include <cmath>
+
+namespace dflash::common {
+namespace mtp {
+
+static ggml_tensor * rms_norm_mul(ggml_context * ctx, ggml_tensor * x,
```
howard0su, line 20: in my other PR, this function is moved to a sharable cpp file. (the PR is merged)

dflash/src/qwen35/qwen35_mtp_graph.cpp
```diff
@@ -0,0 +1,404 @@
...
+        int n_head_kv,
+        int key_len,
+        int val_len,
+        int n_rot,
+        int rope_sections[4],
+        float rope_freq_base,
+        float rms_eps,
+        int slot_start,
+        int n_tokens) {
+    qwen35_mtp_warm_graph_free(sg);
```
howard0su, line 83: I didn't get warm part. We should avoid recreate graph if possible. are u trying to reset the buffer?

howard0su (changes requested):
The main feedback is that we should not treat MTP as a special thing. it is another form of draft model. Leverage the existing code to do everything (except the graph is different than the current model).

dusterbloom:
Closing as stale — superseded by the merged MoE-hybrid / prefix-cache / harness infra. Will reopen from a fresh branch on current main if revived.

PR #238, opened 2026-05-20
feat(server): add CUDA/HIP mixed backend placement

weicj:
# Summary

This PR adds server-side CUDA/HIP mixed draft-target placement for the existing `test_dflash` daemon path.

This draft is stacked on top of #236 (`feat(server): add C++ backend placement foundation`).

Before this change, `server.py` could pass target GPU, draft GPU, target layer split, feature mirror, and peer-access flags into `test_dflash`, but it still assumed one backend environment for the whole daemon. That meant CUDA-only and HIP-only server launches worked, but a CUDA target with a HIP draft, or a HIP target with a CUDA draft, had no clear server-level configuration surface.

Now the server can resolve target and draft backend placement separately, export the correct visible-device environment for each backend, and forward the draft IPC options needed by the mixed-backend DFlash path.

# Changes

- Add server CLI fields for backend-aware placement:
  - `--target-backend cuda|hip`
  - `--target-visible-devices`
  - `--draft-backend cuda|hip`
  - `--draft-visible-devices`
- Add server CLI fields for the remote draft daemon path:
  - `--draft-ipc-bin`
  - `--draft-ipc-gpu`
  - `--draft-ipc-work-dir`
  - `--draft-ipc-ring-cap`
- Extend `placement.server_resolver` so it:
  - logs resolved target and draft backends;
  - maps CUDA placement to `CUDA_VISIBLE_DEVICES`;
  - maps HIP placement to `HIP_VISIBLE_DEVICES` and `ROCR_VISIBLE_DEVICES`;
  - rejects conflicting same-backend visible-device lists in one server process;
  - requires `--draft-ipc-bin` when target and draft backends differ;
  - forwards draft IPC args into the `test_dflash` daemon launch args.
- Keep target layer split within one backend binary. Mixed backend draft execution uses the existing remote draft IPC path; it does not split target layers across CUDA and HIP.
- Add focused resolver tests for same-backend placement, invalid same-backend conflicts, missing mixed-backend IPC config, and CUDA/HIP mixed placement arg/env generation.

# Behavior

CUDA-only and HIP-only server configurations keep the existing behavior.

For mixed CUDA/HIP draft-target placement, the server now requires an explicit draft-side `test_dflash` binary built for the draft backend. The target daemon receives the existing `--target-gpus`, `--target-split-load-draft`, `--target-split-dflash`, and `--draft-ipc-*` flags, so the cross-backend boundary stays at the draft IPC feature exchange instead of trying to load CUDA and HIP into one process.

The native C++ server remains single-backend for now and still rejects cross-backend target/draft placement. This PR enables mixed backend support through the existing Python OpenAI-compatible server plus `test_dflash` daemon path.

# Notes

- The current checks cover the server-side placement resolver, CLI parsing surface, generated daemon args, and CUDA/HIP visible-device environment mapping.
- A dedicated SSH validation machine with RTX 3090 + Strix Halo is available in principle, but formal mixed-backend hardware validation should be added after access is opened for this PR.
- Planned hardware validation target: one direction with CUDA target + HIP draft, and the reverse direction with HIP target + CUDA draft, using the existing draft IPC boundary.

### DIFF
--- a/dflash/scripts/placement/server_resolver.py
+++ b/dflash/scripts/placement/server_resolver.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import MutableMapping
 
+from .backend_device import apply_backend_visible_devices
 from .test_dflash_args import TestDflashLaunchArgs
 
 
@@ -14,13 +15,25 @@ class ServerPlacement:
     daemon_args: list[str]
     prefix_cache_slots: int
     prefill_cache_slots: int
+    target_backend: str
+    draft_backend: str
+    target_visible_devices: str | None
+    draft_visible_devices: str | None
     target_gpu: int | None
     draft_gpu: int | None
     target_gpus: str | None
     target_layer_split: str | None
+    draft_ipc_bin: str | None
+    draft_ipc_gpu: int | None
+    draft_ipc_work_dir: str | None
+    draft_ipc_ring_cap: int | None
     draft_feature_mirror: bool
     peer_access: bool
     cache_slots_disabled: bool = False
+
+    @property
+    def mixed_backend(self) -> bool:
+        return self.target_backend != self.draft_backend
 
     def apply_env(self, env: MutableMapping[str, str]) -> None:
         env.update(self.env_updates)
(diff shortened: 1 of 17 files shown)


weicj:
Closed as python path deprecated

PR #240, opened 2026-05-21
qwen35: add BSA (block-sparse attention) for full-attention layers during prefill

howard0su:
Replace ggml_flash_attn_ext with ggml_flash_attn_sparse in build_full_attn_block for the 16 full-attention layers. When the prompt exceeds 1024 tokens and kv_offset==0 (fresh prefill), process the entire prompt in a single graph pass (no 512-token chunking) so Q_len == KV_len as required by the BSA kernel.

On sm_80+ hardware with pflash registered, this dispatches to block-sparse attention (~3x faster at long contexts). On older hardware (e.g., 2080Ti / sm_75), the ggml_flash_attn_sparse op transparently falls back to dense FA via fattn-sparse.cu.

Changes:
- Register pflash kernel at Qwen35Backend::init() (ifdef sm_80)
- Add use_sparse_attn + sparse_alpha to QwenGraphInputs
- Thread flags through build_target_step → build_qwen35_graph
- BSA threshold + single-shot logic in do_prefill
- Alpha configurable via DFLASH_FP_ALPHA env (default 0.12)

### DIFF
--- a/dflash/src/internal.h
+++ b/dflash/src/internal.h
@@ -499,6 +499,8 @@ struct QwenGraphInputs {
     int           fa_window = 0;  // sliding window for FA layers: 0 = full attention
     bool          last_token_logits_only = false; // if true, only compute logits for last token (prefill optimization)
     ggml_tensor * parent_ids = nullptr; // [n_tokens] i32; tree mode when non-null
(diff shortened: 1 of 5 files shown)


davide221:
Kernel head dim is broken,I would not prioritize BSA for the target model

PR #241, opened 2026-05-21
refactor(dflash): rename namespace dflash27b → dflash::common

dusterbloom:
> **Scope clarification:** This PR renames the **C++ namespace** `dflash27b` → `dflash::common`. The **C ABI** (`dflash/include/dflash27b.h`, `dflash27b_last_error()`, `DFLASH27B_*` macros and env vars) is **intentionally untouched** — deferred to a follow-up. Running `rg dflash27b` will surface those C ABI references and is expected; they're listed under "Out of scope" below.
## Summary

Mechanical rename of namespace `dflash27b` → `dflash::common` per @weicj's review on #237. The legacy name baked the first backend (Qwen3-27B) into shared code; renaming removes the backend leak so future backends plug into a neutral namespace.

## Scope

In:
- `namespace dflash27b` → `namespace dflash::common` (all source + tests)
- `dflash27b::*` qualified uses → `dflash::common::*`
- CMake static library target `dflash27b` → `dflash_common`
- `project(dflash27b)` → `project(dflash)`
- Private CMake variables `_dflash27b_*` → `_dflash_*`
- Stale comments referencing the old namespace name

Out (deferred to a follow-up PR with backward-compat shims):
- Public C header `dflash/include/dflash27b.h` (filename + header guard `DFLASH27B_H`)
- C ABI symbol `dflash27b_last_error()`
- Preprocessor macros `DFLASH27B_*` (build flags + compile-time switches)
- Env vars `DFLASH27B_*` (runtime config)

These all sit on the public surface — renaming them in this PR would be an ABI break without a deprecation cycle. They get a dedicated follow-up.

## Verification

- `rg 'dflash27b::'` → zero hits outside the deferred public header
- `rg 'namespace dflash27b'` → zero hits
- `rg 'using namespace dflash27b'` → zero hits
- Clean build green (full rebuild from empty `dflash/build/`)
- `nm libdflash_common.a` → only `N6dflash6common*` mangling, no `dflash27b` symbols (apart from the intentional `dflash27b_last_error` C ABI)
- Smokes: `smoke_qwen3_forward` + Qwen3-0.6B pass, `test_flashprefill_kernels` pass (4/4), `test_kv_quant` pass (T1–T4)
- Diff is symmetric (insertions ≈ deletions — pure rename)

## Build note for reviewers

CUDA 12.6 + GCC 13.3 has a known `_Float128` conflict during CUDA host-compiler ID detection. If you hit it, add:
```
-DCMAKE_CUDA_HOST_COMPILER=/usr/bin/g++-11
```

### DIFF
--- a/dflash/CMakeLists.txt
+++ b/dflash/CMakeLists.txt
@@ -4,9 +4,9 @@ set_property(CACHE DFLASH27B_GPU_BACKEND PROPERTY STRINGS cuda hip)
 string(TOLOWER "${DFLASH27B_GPU_BACKEND}" DFLASH27B_GPU_BACKEND)
 if(DFLASH27B_GPU_BACKEND STREQUAL "cuda")
     set(DFLASH27B_USER_CUDA_ARCHITECTURES "${CMAKE_CUDA_ARCHITECTURES}")
-    project(dflash27b LANGUAGES C CXX CUDA)
+    project(dflash LANGUAGES C CXX CUDA)
 elseif(DFLASH27B_GPU_BACKEND STREQUAL "hip")
-    project(dflash27b LANGUAGES C CXX HIP)
+    project(dflash LANGUAGES C CXX HIP)
 else()
     message(FATAL_ERROR "DFLASH27B_GPU_BACKEND must be 'cuda' or 'hip', got '${DFLASH27B_GPU_BACKEND}'")
 endif()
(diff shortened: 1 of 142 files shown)


dflash/test/test_flash_attn_sparse_mask.cpp
```diff
@@ -0,0 +1,527 @@
+// TDD Step 1 — FAILING TEST for per-Q additive mask in pflash BF16 kernel.
```
howard0su: intentional?

howard0su: also, these files are not included in cmake.

howard0su (changes requested):

dflash/test/test_flash_attn_sparse_mask.cpp
dusterbloom: These files should not be part of the PR. slipped and removed

PR #242, opened 2026-05-21
test(mtp): add RED-bar scripts for converter + e2e determinism

dusterbloom:
## Summary

Two shell test scripts carved out of #241 per @howard0su's review request to split the rename PR from the test additions.

## What's added

- `dflash/test/test_mtp_converter.sh` — **Phase 1 RED** test for `convert_dflash_to_gguf.py --mtp-assistant` mode. Validates the converter's GGUF output matches the AtomicChat gold reference on (1) tensor list, (2) required MTP tensors, (3) required GGUF metadata keys, (4) `requires_target_arch == "gemma4"`. Exits **77** (autotools skip code) cleanly if the gold reference at `$MTP_GOLD_GGUF` is missing.

- `dflash/test/test_mtp_e2e.sh` — **Phase 3 RED** test for `--draft-method mtp` byte-identical output vs `--draft-method none` at `temp=0 seed=0`. Exits **77** cleanly if the `test_gemma4_dflash` binary or the model files are missing.

## Safe to land now

Both scripts:
- Use `#!/usr/bin/env bash` + `set -euo pipefail`
- Have no C++ namespace dependencies
- Exit 77 (skip) when prerequisites aren't on disk
- Have no global state, no side effects beyond writing to `/tmp`

They'll turn GREEN when the corresponding MTP work lands. Until then they're CI-stable skips.

## Not in this PR

The third test file flagged in #241, `dflash/test/test_flash_attn_sparse_mask.cpp`, is deferred. It's a TDD red-bar that intentionally references symbols (`flash_prefill_forward_bf16_masked()`, 6-arg `ggml_flash_attn_sparse()`) that don't exist yet, so adding it to main without the spike that turns it green would leave a permanently broken build target. It will ship with the sparse-mask spike PR.

### DIFF
--- a/dflash/test/test_mtp_converter.sh
+++ b/dflash/test/test_mtp_converter.sh
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+# Phase 1 RED test: Gemma4 MTP assistant converter
+# Should FAIL until convert_dflash_to_gguf.py learns --mtp-assistant mode.
+# When this passes, our converter produces a GGUF byte-for-byte equivalent in
+# tensor list and metadata to AtomicChat's pre-converted reference.
+
+set -euo pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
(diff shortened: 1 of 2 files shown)


davide221:
Submodule rollback hidden in diff, can you fix?

PR #245, opened 2026-05-21
feat(dflash): VRAM management improvements for C++ server

howard0su:
Two changes to reduce VRAM pressure in the C++ server, especially on 22–24 GB GPUs:

1. --lazy-draft: park decode draft when idle

Park the decode draft model (~3.3 GB) between requests to free VRAM for pflash compression. Mirrors the existing Python server.py --lazy-draft behavior.

Flow: startup → park draft | request → compress → free pflash drafter → unpark draft → generate → park draft

Adds --no-lazy-draft flag to disable (enabled by default).

2. Release scratch VRAM buffers between requests

The target gallocr, LM-head projection gallocr, and BSA persistent CUDA buffers grow monotonically but never shrink. After a large-prompt request, subsequent smaller requests suffer VRAM pressure causing KV cache spill to system RAM and ~2× decode slowdown.

Adds ModelBackend::release_scratch() called after each HTTP request completes. Freed buffers are lazily recreated at the exact size needed on the next request.

### DIFF
--- a/dflash/src/common/model_backend.h
+++ b/dflash/src/common/model_backend.h
@@ -174,6 +174,10 @@ struct ModelBackend {
     // supports_dflash_spec_decode() returns true. Default returns nullptr.
     virtual class DFlashTarget * dflash_target() { return nullptr; }
 
+    // Release oversized scratch buffers between requests to prevent VRAM
+    // growth over time. Default is a no-op.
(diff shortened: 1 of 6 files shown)


davide221:
on by default seems a bit of a stretch considering that pflash is not production ready, what do you think?

PR #246, opened 2026-05-21
feat(server): add native mixed-backend draft placement

weicj:
# Summary

This PR adds the native C++ mixed backend foundation for draft/target placement.

It builds on the backend-device placement shape from #236 and is aligned with the `dflash::common` namespace rename from #241. The native server can now describe:

```bash
dflash_server target.gguf \
  --draft draft.gguf \
  --target-device cuda:0 \
  --draft-device hip:0 \
  --draft-ipc-bin /path/to/hip/dflash_draft_ipc_daemon
```

The lower-level rule is unchanged: one native server process still owns one compiled ggml backend. When target and draft use different backends, draft execution moves behind the existing remote draft IPC transport instead of loading both CUDA and HIP runtimes into the same process.

# Problem

After #236, the native server had a clean `DevicePlacement` surface, but mixed target/draft placement was still rejected before backend creation.

That blocked heterogeneous setups where the target model runs on one backend/runtime and the draft model runs on another, even though the DFlash draft side already had an IPC boundary that can carry the draft execution out of process.

# Changes

1. Added shared remote draft configuration

   - Added `RemoteDraftConfig` under `placement/`.
   - Extended `BackendArgs` with `remote_draft`.
   - Kept this model-family neutral at the placement/config layer so server/factory code does not need Qwen-specific placement logic.

2. Added native server mixed-backend CLI plumbing

   - Added `--draft-ipc-bin`.
   - Added `--draft-ipc-work-dir`.
   - Added `--draft-ipc-ring-cap`.
   - Kept draft GPU selection sourced from `--draft-device <backend:gpu>`.

3. Updated server validation

   - Same-backend draft/target keeps the local in-process path.
   - Mixed target/draft backend now requires both `--draft <path>` and `--draft-ipc-bin`.
   - `--draft-ipc-bin` is rejected for same-backend placement.
   - Target layer split across backends remains rejected in this PR.
   - Unsupported model architectures return a clear remote-draft capability error before backend startup.

4. Added backend capability surface

   - Added `ModelBackend::supports_remote_draft()`.
   - Added `arch_supports_remote_draft()` for early native server validation.
   - Unsupported model families keep the default `false` capability and fail explicitly instead of silently falling back.

5. Connected the first adapter

   - Wired Qwen35/Qwen36 backend config to the shared `RemoteDraftConfig`.
   - When remote draft is enabled, Qwen35 starts `DFlashDraftIpcClient` instead of loading draft weights into the target process.
   - Prefill and replay feature captures are synchronized to the remote draft process.
   - Draft park/unpark/shutdown now treat the remote draft daemon as its own lifecycle and do not free local draft weights that were never loaded.
   - Local same-backend draft execution keeps the existing feature mirror path.

6. Added config logging

   - Server startup now prints target device, draft device, local vs remote draft execution, and remote draft IPC settings when used.

# Notes

- The `placement/` directory now owns both device placement and the remote draft transport config.
- The implementation keeps backend/runtime placement generic; Qwen35/Qwen36 is only the first adapter because it already has the DFlash feature boundary.
- Other model families, including Gemma, can opt in through the same backend capability once their adapter exposes a compatible draft/target feature boundary.
- Local hardware validation completed end-to-end generation in both mixed directions on a Tesla P4 CUDA card and an AMD Pro VII HIP card. The target used for this smoke validation was a Q1 21B target pruned from the 27B family, with a little over 5GB of target weights; it is only used to validate the mixed-backend execution path, not as a quality or performance reference.

### DIFF
--- a/dflash/src/common/backend_factory.cpp
+++ b/dflash/src/common/backend_factory.cpp
@@ -17,6 +17,10 @@ std::string detect_arch(const char * model_path) {
     return info.arch;
 }
 
+bool arch_supports_remote_draft(const std::string & arch) {
+    return arch == "qwen35";
+}
+
 std::unique_ptr<ModelBackend> create_backend(const BackendArgs & args) {
     if (!args.model_path) {
         std::fprintf(stderr, "[backend_factory] model_path is null\n");
@@ -38,6 +42,7 @@ std::unique_ptr<ModelBackend> create_backend(const BackendArgs & args) {
         cfg.draft_path         = args.draft_path;
         cfg.device             = args.device;
         cfg.draft_gpu          = args.draft_device.gpu;
+        cfg.remote_draft       = args.remote_draft;
         cfg.stream_fd          = args.stream_fd;
         cfg.fa_window          = args.fa_window;
         cfg.kq_stride_pad      = args.kq_stride_pad;
--- a/dflash/src/ipc/dflash_draft_ipc_main.cpp
+++ b/dflash/src/ipc/dflash_draft_ipc_main.cpp
@@ -1,0 +1,45 @@
+// Standalone DFlash draft IPC daemon entry point.
+
+#include "dflash_draft_ipc.h"
+
+#include <algorithm>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+
+using namespace dflash::common;
+
+int main(int argc, char ** argv) {
+    if (argc < 3 || std::strcmp(argv[1], "--draft-ipc-daemon") != 0) {
+        std::fprintf(stderr,
+            "usage: %s --draft-ipc-daemon <draft.safetensors|draft.gguf> "
+            "--ring-cap=N --stream-fd=FD [--draft-gpu=N]\n",
+            argv[0]);
+        return 2;
+    }
+
+    const char * draft_path = argv[2];
+    int ring_cap = 4096;
+    int draft_gpu = 0;
+    int stream_fd = -1;
+    for (int i = 3; i < argc; i++) {
+        if (std::strncmp(argv[i], "--ring-cap=", 11) == 0) {
+            ring_cap = std::atoi(argv[i] + 11);
+        } else if (std::strcmp(argv[i], "--ring-cap") == 0) {
+            if (i + 1 < argc) ring_cap = std::atoi(argv[++i]);
+        } else if (std::strncmp(argv[i], "--draft-gpu=", 12) == 0) {
+            draft_gpu = std::max(0, std::atoi(argv[i] + 12));
+        } else if (std::strcmp(argv[i], "--draft-gpu") == 0) {
+            if (i + 1 < argc) draft_gpu = std::max(0, std::atoi(argv[++i]));
+        } else if (std::strncmp(argv[i], "--stream-fd=", 12) == 0) {
+            stream_fd = std::atoi(argv[i] + 12);
+        } else if (std::strcmp(argv[i], "--stream-fd") == 0) {
+            if (i + 1 < argc) stream_fd = std::atoi(argv[++i]);
+        } else {
+            std::fprintf(stderr, "[draft-ipc-daemon] unknown option: %s\n", argv[i]);
+            return 2;
+        }
+    }
+
+    return run_dflash_draft_ipc_daemon(draft_path, ring_cap, draft_gpu, stream_fd);
+}
--- a/dflash/src/qwen35/qwen35_backend.cpp
+++ b/dflash/src/qwen35/qwen35_backend.cpp
@@ -20,9 +20,21 @@
 #include <chrono>
 #include <cstdio>
 #include <cstdlib>
+#include <cstdint>
 #include <cstring>
 
 namespace dflash::common {
+
+namespace {
+static float bf16_bits_to_f32(uint16_t bits) {
+    union {
+        uint32_t u;
+        float f;
+    } v;
+    v.u = (uint32_t)bits << 16;
+    return v.f;
+}
+}  // namespace
 
 #define IS_EOS_TOK(tok, w)                                         \
     ( ((w).eos_chat_id >= 0 && (tok) == (w).eos_chat_id)                  \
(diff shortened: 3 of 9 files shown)


dflash/src/common/backend_factory.cpp
```diff
@@ -17,6 +17,10 @@ std::string detect_arch(const char * model_path) {
     return info.arch;
 }
 
+bool arch_supports_remote_draft(const std::string & arch) {
+    return arch == "qwen35";
```
howard0su, line 21: why gemma4 cannot support this?

dflash/src/ipc/dflash_draft_ipc_main.cpp
```diff
@@ -0,0 +1,45 @@
+// Standalone DFlash draft IPC daemon entry point.
```
howard0su, line 1: this is not proper place for this binary main. put it to src/ipc folder.

dflash/src/server/server_main.cpp
```diff
@@ -44,8 +127,14 @@ static void print_usage(const char * prog) {
         "  --host <addr>        Bind address (default: 0.0.0.0)\n"
         "  --max-ctx <N>        Max context length (default: 131072)\n"
         "  --max-tokens <N>     Default max output tokens (default: 4096)\n"
-        "  --gpu <N>            Target GPU device (default: 0)\n"
-        "  --draft-gpu <N>      Draft GPU device (default: 0)\n"
+        "  --target-device <backend:gpu>  Target device (default: auto:0)\n"
+        "  --draft-device <backend:gpu>   Draft device (default: auto:0)\n"
+        "  --draft-ipc-bin <path>         Remote draft IPC daemon for mixed backends\n"
```
howard0su, line 132: if this is IPC on single machine, why we need a process to handle IPC? Why we need a daemon?

dflash/src/qwen35/qwen35_backend.cpp
```diff
@@ -757,6 +807,39 @@ bool Qwen35Backend::do_ar_decode(int committed, int n_gen,
...
+        for (int t = 0; t < n_tokens; ++t) {
+            const int slot = (start_pos + t) % cache_.target_feat_cap;
+            const size_t src_offset = (size_t)slot * src_stride +
+                (size_t)cap_idx * (size_t)feat_hidden * sizeof(uint16_t);
+            ggml_backend_tensor_get(cache_.target_feat, bf16.data(),
+                                    src_offset,
+                                    sizeof(uint16_t) * (size_t)feat_hidden);
+            float * dst = slice.data() + (size_t)t * feat_hidden;
+            for (int h = 0; h < feat_hidden; ++h) {
+                dst[h] = bf16_bits_to_f32(bf16[h]);
```
howard0su, line 866: why convert to f32? if we really want to, use ggml convert functions.

dflash/src/ipc/dflash_draft_ipc_main.cpp
```diff
@@ -0,0 +1,45 @@
...
+        } else if (std::strcmp(argv[i], "--ring-cap") == 0) {
+            if (i + 1 < argc) ring_cap = std::atoi(argv[++i]);
+        } else if (std::strncmp(argv[i], "--draft-gpu=", 12) == 0) {
+            draft_gpu = std::max(0, std::atoi(argv[i] + 12));
+        } else if (std::strcmp(argv[i], "--draft-gpu") == 0) {
+            if (i + 1 < argc) draft_gpu = std::max(0, std::atoi(argv[++i]));
+        } else if (std::strncmp(argv[i], "--stream-fd=", 12) == 0) {
+            stream_fd = std::atoi(argv[i] + 12);
+        } else if (std::strcmp(argv[i], "--stream-fd") == 0) {
+            if (i + 1 < argc) stream_fd = std::atoi(argv[++i]);
```
howard0su, line 37: why no use mmap? stream fd seems slow to send big chunk of data.

howard0su (changes requested):
is this remote daemon work around the problem that CUDA and HIP cannot load into single process?

weicj:
> is this remote daemon work around the problem that CUDA and HIP cannot load into single process?

yes, they cannot be merged into one single process and we'll keep it as-is to avoid possible crash. An IPC daemon will only be introduced to communicate when mixed backend loaded (for single backend IPC won't be loaded)

dflash/src/common/backend_factory.cpp
weicj: I'll make it supported from another PR after gemma4 and other model backend feature fixed and merged into main.

dflash/src/qwen35/qwen35_backend.cpp
weicj: It's on my PR plan for vram optimization afterwards, for now we keep a unified F32 to maximize the compatiblity for all the hardware setup (including legacy ones like Turin/Volta/GCN)

dflash/src/ipc/dflash_draft_ipc_main.cpp
weicj: The stream fd is only used for small control/status messages, not for the large feature payload. The current first-pass transport writes feature/noise tensors to temporary files and sends paths over the control channel. 

But yes I agree mmap/shared memory is the better next step for reducing host-copy and filesystem overhead, let me mark it to the follow-up vram optimization plan.

dflash/src/server/server_main.cpp
weicj: For single backend, IPC won't be called

dflash/src/ipc/dflash_draft_ipc_main.cpp
weicj: Agree, let me move it

davide221:
@weicj can you rebase and solve conflicts please

weicj:
> @weicj can you rebase and solve conflicts please

I've seen Gemma4 merged, will make a separate HIP runtime fix first, and update this one.

weicj:
@davide221 updated, make sure https://github.com/Luce-Org/lucebox-hub/pull/258 is merged before this one, when review passed :)

davide221:
image
Considering that in a same-backend A/B testing the IPC path did not show decode-side overhead, we can integrate smoothly. Thanks for the contribution

PR #247, opened 2026-05-21
fix(server): emit Anthropic tool_use content blocks (non-stream + stream)

sanastasiou:
## Problem

The `/v1/messages` Anthropic endpoint in `dflash_server` sets `stop_reason="tool_use"` correctly when the parser detects a tool call, but the response payload never includes the `tool_use` content blocks. Two places are missing the serialization:

### 1. Non-streaming response builder (`http_server.cpp`)

The `ApiFormat::ANTHROPIC` arm of the response builder only emits `text` and (optionally) `thinking` blocks. Compare to the `OPENAI_CHAT` branch immediately above, which already serialises `emitter.tool_calls()` into `msg[\"tool_calls\"]`. The ANTHROPIC branch was missing the equivalent serialisation. Clients see:

\`\`\`json
{
  \"stop_reason\": \"tool_use\",
  \"content\": [{ \"type\": \"text\", \"text\": \"\" }]
}
\`\`\`

— a tool_use stop with no tool_use block.

### 2. Streaming finish (`sse_emitter.cpp`)

The `ApiFormat::ANTHROPIC` arm of the tool-call dispatch in `emit_finish` was literally a no-op stub:

\`\`\`cpp
case ApiFormat::ANTHROPIC:
    // Anthropic doesn't have streaming tool calls in the same way
    // Tool use blocks would be separate content blocks
    break;
\`\`\`

Clients (claude-agent-sdk, the official Anthropic SDKs) saw the initial text \`content_block_start\`, then a \`content_block_stop\`, then \`message_delta\` with \`stop_reason=end_turn\` — and **no \`tool_use\` events at all**. The model's tool call was silently dropped.

### 3. Trailing \`message_delta\` stop_reason

\`emit_finish\` hard-coded \`stop_reason=\"end_turn\"\` in the final \`message_delta\` regardless of tool calls.

## What this PR does

1. **Non-streaming** — adds tool_use block serialisation to the Anthropic branch, parsing \`tc.arguments\` JSON into the \`input\` object Anthropic expects (object, not string). Falls back to empty object on parse failure.

2. **Streaming** — implements the full Anthropic tool_use lifecycle in \`emit_finish\`:
   - \`content_block_stop\` on the open text/thinking block (if any)
   - For each tool_call:
     - \`content_block_start\` (\`type: tool_use\` with \`id\`/\`name\`/empty \`input\`)
     - \`content_block_delta\` (\`type: input_json_delta\` with arguments JSON)
     - \`content_block_stop\`
   - Tracks via \`active_kind_\` so the final \`emit_finish\` \`content_block_stop\` doesn't double-close.

3. **Trailing stop_reason** — uses \`\"tool_use\"\` when any tool calls were emitted, \`\"end_turn\"\` otherwise.

## Test plan

- [x] New \`test_emitter_anthropic_tool_use_blocks\` unit test exercising the full SSE lifecycle: asserts the byte stream contains a tool_use content block, an \`input_json_delta\` carrying the arguments JSON, and \`stop_reason=tool_use\` in \`message_delta\`. Regression guard on \`content_block_stop\` count (text-block-close + tool_use-block-close).
- [x] End-to-end smoke against \`/v1/messages\` with a get_weather tool: model emits Qwen3 XML, tool_parser converts it, response now contains a proper tool_use block (\`{\"type\":\"tool_use\",\"name\":\"get_weather\",\"input\":{\"city\":\"Tokyo\"}}\`).
- [x] All existing tests still pass (112 → 113 with the new one).

## Impact

Without this fix \`dflash_server\` can't serve any Anthropic-format tool-using client (claude-agent-sdk, Cline, Continue, etc.) — the tool call is generated by the model but lost in serialisation, leaving the client with an empty response and the wrong impression that the model declined to use tools.

### DIFF
--- a/dflash/src/server/http_server.cpp
+++ b/dflash/src/server/http_server.cpp
@@ -827,7 +827,37 @@ void HttpServer::worker_loop() {
                 if (!emitter.reasoning_text().empty()) {
                     content.push_back({{"type", "thinking"}, {"thinking", emitter.reasoning_text()}});
                 }
-                content.push_back({{"type", "text"}, {"text", emitter.accumulated_text()}});
+                // Only emit a text block when there is actual text. When the
+                // model emitted ONLY a tool_call (Qwen3 XML), accumulated_text
+                // is empty — pushing an empty text block confuses Anthropic
+                // SDK clients (they expect tool_use blocks alone).
+                if (!emitter.accumulated_text().empty()) {
+                    content.push_back({{"type", "text"}, {"text", emitter.accumulated_text()}});
+                }
+                // Tool calls — the OPENAI_CHAT branch above does this; the
+                // ANTHROPIC branch was missing the tool_use serialisation,
+                // so stop_reason="tool_use" was returned with empty content.
+                // tc.arguments is a JSON-encoded string; parse to object for
+                // Anthropic's `input` field (Anthropic expects object, not
+                // string). Fall back to empty object on parse failure.
+                if (!emitter.tool_calls().empty()) {
+                    for (const auto & tc : emitter.tool_calls()) {
+                        json input_obj;
+                        try {
+                            input_obj = tc.arguments.empty()
+                                ? json::object()
+                                : json::parse(tc.arguments);
+                        } catch (const std::exception &) {
+                            input_obj = json::object();
+                        }
+                        content.push_back({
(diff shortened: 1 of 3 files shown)


howard0su (approved):
LGTM

PR #263, opened 2026-05-23
feat(server): add mixed-backend PFlash phase split

weicj:
## Summary

This PR adds native C++ server support for PFlash phase split with CUDA/HIP mixed-backend drafter placement.

Previously, C++ server PFlash compression could only use the target backend's in-process drafter path. That forced the PFlash drafter to follow the target backend/device, so the server could not place the target model and PFlash drafter on different CUDA/HIP backends or different backend-local GPUs.

After this change, `--draft-device <backend:gpu>` is also used as the PFlash drafter placement when `--prefill-compression` is enabled. Same-backend PFlash still uses the local typed `backend_.compress(...)` path, while mixed-backend PFlash runs the drafter through the backend-neutral IPC process layer from #261.

## Changes

- Add a PFlash compression mode to the shared backend IPC process.
  - Add `PFlashDrafterIpcClient`.
  - Add `pflash-compress` as a second mode on the shared `backend_ipc_daemon` process layer from #261.
  - The server sends drafter-tokenized prompt IDs to the daemon and receives compressed drafter token IDs back.
  - Mixed-backend PFlash requires `--draft-ipc-bin`; missing IPC config fails early with a clear error.
  - A failed remote PFlash IPC compression closes the IPC client so a later request can start a fresh daemon.

- Make PFlash use the server's draft placement surface.
  - `--draft-device <backend:gpu>` now controls the PFlash drafter placement when `--prefill-compression` is enabled.
  - `--prefill-compression off` is accepted without requiring `--prefill-drafter`.
  - Startup logs show the resolved PFlash drafter GPU and whether PFlash drafter execution is local or remote IPC.

- Add a focused C++ placement helper for PFlash drafter resolution.
  - New `src/placement/pflash_placement.h` resolves target backend, drafter backend, backend-local drafter GPU, and whether the drafter must run through remote IPC.
  - Drafter placement is only validated when it is actually used by decode draft or PFlash; target-only startup is not blocked by an unused `--draft-device`.

- Pass backend-local drafter GPU selection through the PFlash compression path.
  - `CompressRequest` now carries the resolved drafter GPU.
  - The Qwen3-based PFlash drafter loader can select a backend-local GPU instead of always binding GPU 0.
  - Qwen35 PFlash compression passes the same drafter GPU through the typed compression path.

- Make required PFlash compression failure explicit.
  - When PFlash is selected for a request, drafter startup / IPC / compression failures now return a request error instead of silently falling back to the uncompressed prompt.
  - This avoids false-positive validation where the server returns HTTP 200 while PFlash placement or mixed-backend IPC was actually not working.

## Notes

- Depends on the native mixed-backend placement foundation from #246 and the backend-neutral IPC process layer from #261.
- Checked CUDA sm61 and ROCm 6.3.3 / gfx906 HIP builds for `dflash_server`, `backend_ipc_daemon`, and `test_server_unit`; unit tests passed on both builds.
- Runtime smoke passed with a qwen35-family 21B IQ1 target and Qwen3-0.6B BF16 PFlash drafter, using CUDA target on Tesla P4 and HIP remote drafter on Pro VII; non-streaming `/v1/chat/completions` returned HTTP 200 with PFlash remote IPC enabled.
- The reverse HIP-target/CUDA-drafter direction was also exercised with the same `pflash-compress` IPC mode. On this local P4 setup it passes after applying the separate legacy-CUDA F32 drafter compatibility patch; that compatibility patch is intentionally kept out of this PR and should land separately.

### DIFF
--- a/dflash/CMakeLists.txt
+++ b/dflash/CMakeLists.txt
@@ -240,6 +240,7 @@ add_library(dflash_common STATIC
     src/common/dflash_capture.cpp
     src/common/dflash_draft_ipc.cpp
     src/common/dflash_draft_ipc_daemon.cpp
+    src/common/pflash_drafter_ipc.cpp
     src/common/dflash_draft_graph.cpp
     src/common/dflash_spec_decode.cpp
     src/qwen35/graph_builders.cpp
@@ -675,6 +676,7 @@ if(DFLASH27B_TESTS)
     if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/src/ipc/backend_ipc_main.cpp")
         add_executable(backend_ipc_daemon
             src/ipc/backend_ipc_main.cpp
+            src/common/pflash_drafter_ipc_daemon.cpp
         )
         target_include_directories(backend_ipc_daemon PRIVATE ${DFLASH27B_SRC_INCLUDE_DIRS})
         if(DFLASH27B_GPU_BACKEND STREQUAL "hip")
--- a/dflash/src/common/backend_factory.cpp
+++ b/dflash/src/common/backend_factory.cpp
@@ -21,6 +21,10 @@ bool arch_supports_remote_draft(const std::string & arch) {
     return arch == "qwen35";
 }
 
+bool arch_supports_pflash_compression(const std::string & arch) {
+    return arch == "qwen35" || arch == "qwen3";
+}
+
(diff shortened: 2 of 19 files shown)


davide221:
@weicj can you rebase please

weicj:
> @weicj can you rebase please

Updated

PR #262, opened 2026-05-23
Split MoE weights between CPU & CUDA, support qwen35moe models

howard0su:
Add Support for arch of qwen35moe, (including Qwen35-35B-A3B, Qwen35-35B-A3B, etc.)

Measured on RTX 2080 Ti (Qwen3.6-35B-A3B-UD-Q4_K_M, all GPU):
    - AR decode: 37.1 tok/s, 2644 graph nodes
    - Per-token compute: ~26.9ms

- [x] Support loading all experts on GPU
- [ ] Support hybrid prefill
- [ ] Support hybrid decode

 ┌──────────┬──────────┬──────────┬─────────┬────────┬─────────┐
 │ Budget   │ Hot/Cold │ GPU %    │ tok/s   │Prefill │ Decode  │
 ├──────────┼──────────┼──────────┼─────────┼────────┼─────────┤
 │ 1000 MB  │  549/9691│   5.4%   │ 10.8    │ 1.6s   │ 23.7s   │
 │ 2000 MB  │ 1101/9139│  10.8%   │ 12.4    │ 1.7s   │ 20.7s   │
 │ 5000 MB  │ 2756/7484│  26.9%   │ 12.6    │ 1.4s   │ 20.3s   │
 │ 9000 MB  │ 4963/5277│  48.5%   │ 13.8    │ 1.4s   │ 18.6s   │
 │12000 MB  │ 6618/3622│  64.6%   │ 16.5    │ 1.3s   │ 15.5s   │
 │15000 MB  │ 8274/1966│  80.8%   │ 18.7    │ 1.1s   │ 13.7s   │
 └──────────┴──────────┴──────────┴─────────┴────────┴─────────┘

Key observations:

 - Going from 5% → 81% GPU experts only improves decode from 10.8 → 18.7 tok/s (1.73× speedup)
 - The curve is nearly linear — not showing the expected hockey-stick where having all hot experts on GPU dominates
 - Even with only 5% hot experts (549/10240), we still get 10.8 tok/s — this suggests CPU FFN is not the bottleneck; the attention/DeltaNet layers (always on GPU) dominate per-token cost
 - Prefill scales better (1.6s → 1.1s) but is also modest

PR #269, opened 2026-05-23
feat(cpp-server): thinking-budget v2 + multi-dialect reasoning aliases + spec

easel:
Adds per-request thinking-budget controls, multi-dialect reasoning emission, and JSON model-card sidecars to the native C++ dflash server.

## What changed

- **Thinking-budget mechanism (Level 2 BudgetHook + watchdog).** Backend AR / spec-decode injects the model's `</think>` close-token sequence once when `n_gen − committed ≤ hard_limit_reply_budget` — KV-continuous, mid-stream, in-process, uniform across streaming and non-streaming. Degenerate-decode watchdog detects post-close runaway and aborts cleanly, surfaced as a flag in `finish_details` and bench output. Sidecar `thinking_terminator_hint` ↔ tokenizer resolves `think_close_token_ids` at startup so the hook is family-aware without hardcoded archs. Resolution order: CLI > sidecar > family fallback > hard fallback (default `hard_limit_reply_budget` = 4096; terse models override down).
- **Multi-dialect reasoning emission.** SSE emitter splits reasoning ↔ content for OpenAI Chat (`reasoning_content` delta), Anthropic (separate thinking / text content-block lifecycle), and Responses API (reasoning stripped per Codex r1 P2). Qwen3.6 `<think>` / `</think>` special-token ids are forwarded as text into the emitter; gemma4 `<|channel>` / `<channel|>` are mapped onto the same channel so all archs share one state machine. `first_content_token_index()` derives the natural-close split from the REASONING→CONTENT transition; leading `<think>` opener is detected up-front so `thinking_tokens` accounts correctly for the Qwen3.6 streamed-thinking path.
- **/props endpoint.** Wholesale `model_card` (verbatim sidecar JSON, validates against `share/model_cards/_schema.json`) + `budget_envelope` (effective `think_max_tokens`, `default_max_tokens`, `hard_limit_reply_budget`, `effort_tiers`, `model_card_source` label) + `runtime` fields (`chunk`, `target_device`, `draft_device`, `speculative_enabled`, `fa_window`, `ddtree_budget`, `kv_cache_k/v`, `runtime_backend`). Captured by `server_main` at startup so the handler doesn't crack `BackendArgs`.
- **Model-card sidecars.** `share/model_cards/{qwen3.6-27b,gemma-4-26b-a4b-it,gemma-4-31b-it,laguna-xs.2}.json` each ship `max_tokens`, `complex_problem_max_tokens`, `hard_limit_reply_budget`, `thinking_terminator_hint`, sampling defaults, and `reasoning_effort_tiers` as applicable. `share/model_cards/_schema.json` provides the JSON Schema for validation. `model_card.cpp` resolver does keyword + family-fallback lookup, with the startup banner reporting the resolved `model_card_source` so operators can confirm which envelope is in force.
- **Spec docs.** `docs/specs/thinking-budget.md` (mechanism, resolution order, CLI surface, `finish_details` / `close_kind` contract), `docs/specs/model-cards.md` (sidecar field reference, family fallback table, ship/override guidance), `docs/specs/props-endpoint.md` + `docs/specs/openapi-props.yaml` (shape + OpenAPI 3 schema for the `/props` payload).

## Validation

- `cmake --build dflash/build -j` clean (warnings only on pre-existing `-Wunused-result` `fread` calls in unrelated test code).
- `ctest -V -R server_unit` → 1542 assertions, 0 failures, including:
  - SSE emitter reasoning split across OpenAI Chat / Anthropic / Responses.
  - `first_content_token_index` across natural-close, never-closed, content-only, and Qwen3.6 streamed-thinking (new regression test added in this PR — leading `<think>` opener no longer captures `fci=0`).
  - `/props` body shape (wholesale sidecar + family-fallback null).
  - `usage.timings` across all three response shapes.
- Confirmed the regression test catches the underlying bug: reverting `dflash/src/server/sse_emitter.cpp` to pre-fix state makes `test_emitter_first_content_index_qwen36_streaming_thinking` fail with `em.first_content_token_index() > 0`.

## Out of scope / follow-ups

- **31b backend wiring.** `share/model_cards/gemma-4-31b-it.json` ships but the dense-30.7B backend path is not yet integrated; the sidecar is in place for when it lands.
- **gemma4 MoE expert-split.** `gemma-4-26b-a4b-it` runs end-to-end at this PR's level of integration, but the MoE expert-split optimization is howard0su's PR #262 territory.
- **MTP (Multi-Token Prediction).** Tracked by upstream PR #23398; orthogonal to the thinking-budget / sidecar work here.

### DIFF
--- a/dflash/CMakeLists.txt
+++ b/dflash/CMakeLists.txt
@@ -655,6 +655,7 @@ if(DFLASH27B_TESTS)
         add_executable(dflash_server
             src/server/server_main.cpp
             src/server/http_server.cpp
+            src/server/model_card.cpp
         )
         target_include_directories(dflash_server PRIVATE ${DFLASH27B_SRC_INCLUDE_DIRS})
         if(DFLASH27B_GPU_BACKEND STREQUAL "hip")
@@ -720,7 +721,9 @@ if(DFLASH27B_TESTS)
 
     if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test/test_server_unit.cpp")
         add_executable(test_server_unit test/test_server_unit.cpp)
-        target_sources(test_server_unit PRIVATE src/server/http_server.cpp)
+        target_sources(test_server_unit PRIVATE
+            src/server/http_server.cpp
+            src/server/model_card.cpp)
         target_include_directories(test_server_unit PRIVATE ${DFLASH27B_SRC_INCLUDE_DIRS})
         if(DFLASH27B_GPU_BACKEND STREQUAL "hip")
             target_compile_definitions(test_server_unit PRIVATE DFLASH27B_BACKEND_HIP=1 GGML_USE_HIP)
--- a/dflash/src/common/model_backend.h
+++ b/dflash/src/common/model_backend.h
@@ -46,6 +46,39 @@ struct DaemonIO {
 
 // ─── Generate request/result ────────────────────────────────────────────
 
+// Thinking-budget force-close hook. Mirrors antirez/ds4 ds4_eval.c's
+// hard_limit_reply_budget semantics: when the budget remaining (n_gen
+// minus tokens committed so far) falls to hard_limit_remaining, the
+// next sampled tokens get overridden with close_token_ids in order,
+// giving the model the remaining budget to write a visible answer
+// after the injected close-tag sequence.
+//
+// Single vs multi-token close:
+//   Qwen3.6: </think> is one added_token (id 248069). close_token_ids
+//            has size 1. One override + budget_close_injected=true.
+//   DeepSeek/laguna: </think> tokenizes to 3 ordinary tokens
+//            ([1718, 37947, 32] for DS-V3). close_token_ids has
+//            size 3. Three consecutive overrides, then resume.
+//
+// This is "Level 2" of our thinking-budget migration: in-process
+// mid-stream force-close, KV-continuous. Beats Level 1's phase-2
+// reprompt because the model never sees a fresh prefill — its KV
+// state continues naturally after the injected close.
+//
+// Current implementation: AR-decode only. When budget_hook is set,
+// backends MAY route generation through their AR path (skipping spec
+// decode) — the perf trade-off is acceptable since this only kicks in
+// for thinking-enabled requests. Spec-decode integration is a follow-up.
+struct BudgetHook {
+    // Multi-token close sequence injected when `(n_gen - committed)`
+    // drops to `hard_limit_remaining`. For Qwen3.x this is the
+    // canonical "Considering the limited time..." summarize-and-stop
+    // lead-in (tokenized at server startup); for non-qwen arches it's
+    // a single close-tag token. Empty = hook disabled.
+    std::vector<int32_t> close_token_ids;
+    int                  hard_limit_remaining = 0;
+};
+
 struct GenerateRequest {
     std::vector<int32_t>       prompt;
     int                        n_gen       = 0;
@@ -65,6 +98,8 @@ struct GenerateRequest {
     // When non-null, the spec decode loop uses these as draft overrides,
     // bypassing draft model computation for covered positions.
     const std::vector<int32_t> * hint_tokens = nullptr;
+    // Optional thinking-budget hook — see BudgetHook docs above.
+    BudgetHook                 budget_hook;
 };
 
 struct GenerateResult {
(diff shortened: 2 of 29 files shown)


easel:
Followed up with `codex review` after the cubic round. Three findings; one P1 fixed in this push (`63bd7a8`), two P2 deferred with rationale.

### P1 (fixed) — Force-close budget used absolute KV frame

`dflash/src/qwen35/qwen35_backend.cpp:769` — the lambda computed `remaining = n_gen - committed_now`, but `committed_now` is the absolute KV position (= prompt_len + tokens generated this call) while `n_gen` is gen-only count. The subtraction treated prompt-length tokens as if they were generated output, firing `--hard-limit-reply-budget` `prompt_len` tokens early on every prompted request. For the spec-decode → AR tail-off path (line 985 caller), `n_gen` is remapped to the remaining-budget headroom, so the subtraction could go negative and force-close immediately as soon as AR took over — defeating the purpose of having a hard-limit reply budget for the visible answer.

Fix: capture `committed_at_entry = committed` at entry and compute `generated = committed_now - committed_at_entry; remaining = n_gen - generated`. Both call sites (post-prefill AR fallback and spec-decode tail-off) now have correct budget arithmetic in the "generated since entry" frame.

Practical impact on the bench (~200-token prompts, n_gen=16000, hard_limit=512): force-close was firing at decoded≈15288 instead of 15488 — only ~1.3% short of intended. The headline pass/fail flip wasn't affected because the model still had plenty of budget. Tail-off path was effectively broken (`remaining` going negative) but rarely hit, which is why bragi's audit #47 (`c39326a`) caught related tail-off bugs but missed this one.

### P2 (deferred) — Phase-1 token count includes post-`</think>` content

`dflash/src/server/http_server.cpp:1504-1505` — `finish_details.thinking_tokens` and `usage.completion_tokens_details.reasoning_tokens` both use `phase1_tokens.size()`, but when the model naturally emits `</think>` mid-stream, `phase1_tokens` contains both the reasoning AND the visible answer that follows. Today we count the lot as thinking and report `content_tokens` as only the phase-2-reprompt output, so the split is wrong whenever phase-1 doesn't get force-closed.

Why deferred: the fix needs the emitter to expose its REASONING→CONTENT transition point in token-position terms (it already knows the text-mode transition via `<think>/</think>`; just needs to surface "first content token index" so we can split `phase1_tokens` at that boundary). Reasonable but touches the emitter API. Filing for a focused follow-up so this PR stays tight.

### P2 (deferred) — `/props` cache snapshot is technically UB

`dflash/src/server/prefix_cache.cpp:440-442` — `stats()` reads `entries_.size()` and `lifetime_hits_` without atomics or a lock; a worker-thread `confirm_inline_snap()` mutation racing with a client-thread `/props` read is data-race UB by the C++ memory model, not just tearable values.

Why deferred: the spec already explicitly accepts the trade-off ("lockless snapshot … acceptable for an introspection report — matches the Python implementation's semantics"). To make it strictly well-defined we'd move `lifetime_hits_`/`full_lifetime_hits_` to `std::atomic<int64_t>` (cheap) and either snapshot `entries_.size()` under the daemon_lock or replace it with an atomic counter (medium). Worth doing — but not gating this PR. Filing as a separate small fix.

### Cubic's P2 follow-up still outstanding

The duplicate `system`-block normalization between `/v1/messages` and `/v1/messages/count_tokens` is still its own follow-up PR; haven't extracted the helper here.

Force-push: `63bd7a8` over `0b9a8ac`. Diff vs previous push: one C++ patch on `qwen35_backend.cpp` (P1 fix above).

easel:
`docs/specs/thinking-budget.md` rewritten as a self-standing design doc in `d4c71a8` — no more "this PR" / cross-link references, no shipping-status framing, no v1-deprecation history. Same content surface, just structured as a permanent spec.

Sections:
1. **Background** — why a single `max_tokens` cap is insufficient for reasoning-capable models.
2. **Terminology** — phase 1 / phase 2 / budget envelope / close kind, all defined inline.
3. **Configuration** — `--think-max-tokens` / `--default-max-tokens` / `--hard-limit-reply-budget` with defaults table.
4. **Request shape** — `thinking:{type:"enabled"}` opt-in vs the `reasoning:{effort:...}` legacy-cap path.
5. **Close strategies** — Level 1 phase-2 reprompt and Level 2 in-process force-close each fully defined (mechanism, support matrix, budget arithmetic).
6. **Response shape** — multi-dialect reasoning aliases (provider comparison table + dflash emit) and `finish_details` schema.
7. **Close-kind taxonomy** — `natural` / `hard` / (reserved) `soft`.
8. **Streaming** — same caps, format-appropriate SSE deltas.
9. **Out of scope** — per-request override, soft close, per-token close-info, phase-split `reasoning_details`.

325 lines (down from 429). Same content landed on integration as `8c37f3c`.

PR #271, opened 2026-05-24
fix(cpp-server): harden C++ streaming tool calls

jkyamog:
I had codex fix some of tool calls last week, it seems there is already been progress to this.  So I asked it to use the upstream which already includes PR #252 and only add the missing things I already had.  Been testing this with Hermes and with my PR #270 as I needed better logging. Below is codex elaboration of this fix:

Follow-up to upstream PR #252 for the native C++ `dflash_server` path.

Upstream PR #252 added C++ streaming detection for:

- `<tool_call>`
- bare `<function=...>`
- `<tool_code>`

This PR keeps that behavior but hardens two remaining edge cases:

- tool-like XML should only enter tool-buffer mode when request tools are present
- malformed/incomplete tool XML should not be emitted back as visible assistant text if parsing fails

## Observed Failure

In testing, the model exposed tool XML as visible assistant text:

```xml
<tool_call>
<function=terminal>
<parameter=command>
ls -la /tmp/lop/
</parameter>
```

That should not be shown as assistant text when the request contains tools. If it parses cleanly, it should become a structured tool call. If it is malformed/incomplete, it should be suppressed rather than leaked.

## Fix

The C++ SSE emitter now:

- detects `<tool_call>`, bare `<function=`, and `<tool_code>` as tool-call starts
- only enables that detection when request tools are present
- suppresses malformed/incomplete buffered tool XML instead of leaking it as visible content
- preserves XML-like text as ordinary assistant content when no tools were supplied

## Tests Added

C++ unit coverage for:

- streaming `<tool_call>...</tool_call>` with request tools
- streaming bare `<function=...>` tool call
- malformed/incomplete tool XML not leaking as assistant content
- no-tools requests preserving XML-like text as ordinary content

### DIFF
--- a/dflash/src/server/sse_emitter.cpp
+++ b/dflash/src/server/sse_emitter.cpp
@@ -17,7 +17,24 @@ static const char FUNCTION_OPEN[] = "<function=";
 static const char TOOL_CODE_OPEN[] = "<tool_code>";
 static constexpr size_t THINK_OPEN_LEN  = 7;
 static constexpr size_t THINK_CLOSE_LEN = 8;
-static constexpr size_t TOOL_OPEN_LEN   = 11;
+
+static bool has_request_tools(const json & tools) {
+    return tools.is_array() && !tools.empty();
+}
+
+static bool find_tool_start(const std::string & text, size_t & pos) {
+    size_t idx = text.find('<');
+    while (idx != std::string::npos) {
+        if (text.compare(idx, sizeof(TOOL_OPEN) - 1, TOOL_OPEN) == 0 ||
+            text.compare(idx, sizeof(FUNCTION_OPEN) - 1, FUNCTION_OPEN) == 0 ||
+            text.compare(idx, sizeof(TOOL_CODE_OPEN) - 1, TOOL_CODE_OPEN) == 0) {
+            pos = idx;
+            return true;
+        }
+        idx = text.find('<', idx + 1);
+    }
+    return false;
+}
 
 static std::string gen_item_id() {
     static std::atomic<uint64_t> ctr{0};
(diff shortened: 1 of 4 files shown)


dflash/src/server/sse_emitter.cpp
```diff
@@ -17,7 +17,25 @@ static const char FUNCTION_OPEN[] = "<function=";
 static const char TOOL_CODE_OPEN[] = "<tool_code>";
 static constexpr size_t THINK_OPEN_LEN  = 7;
 static constexpr size_t THINK_CLOSE_LEN = 8;
-static constexpr size_t TOOL_OPEN_LEN   = 11;
+
+static bool has_request_tools(const json & tools) {
+    return tools.is_array() && !tools.empty();
+}
+
+static bool find_tool_start(const std::string & text, size_t & pos) {
```
howard0su, line 25: we may need to consider regex here. scan the text three times is not a good idea.

dflash/test/test_server_unit.cpp
```diff
@@ -392,6 +419,54 @@ static void test_emitter_anthropic_tool_use_blocks() {
...
+        TEST_ASSERT(args["command"] == "ls -la /tmp/lop/");
+    }
+    TEST_ASSERT(em.accumulated_text().find("<function=terminal>") == std::string::npos);
+}
+
+static void test_emitter_does_not_leak_malformed_tool_xml() {
+    auto em = make_emitter(ApiFormat::OPENAI_CHAT, false, weather_tools());
+    em.emit_start();
+    em.emit_token("Let me list files.\n\n");
+    em.emit_token("<tool_call>\n"
```
howard0su, line 445: even no </tool_call>?

howard0su (changes requested):

dflash/src/server/sse_emitter.cpp
jkyamog: single pass is now implemented.

dflash/test/test_server_unit.cpp
jkyamog: I didn't really fully understand here, but codex understood your comment, I hope this is correct, see below:

Second concern: the parser could handle bare malformed Qwen/OpenClaw-style output like:
```
  <function=terminal>
  <parameter=command>
  ls -la /tmp/lop/
  </parameter>
  </function>
```
  But if the model started a wrapper and forgot to close it:
```
  <tool_call>
  <function=terminal>
  <parameter=command>
  ls -la /tmp/lop/
  </parameter>
  </function>
```
  the inner function could parse, while the outer `<tool_call>` wrapper was not part of the removed span. That risked leaking `<tool_call>` into visible assistant text. I fixed the parser so when it parses a bare `<function>...</function>`, it also
  removes a preceding <tool_call> wrapper if only whitespace exists between the wrapper and the function.

  I added a regression test for exactly that missing outer-close case.

howard0su (approved):

PR #274, opened 2026-05-24
feat(pflash): prefill compress up to 128k -> 2-12× prefill (content-dependent), decode at  parity

dusterbloom:
## TL;DR

Adaptive composition stack: pflash compression + dflash spec-decode + ee7 drafter + anchor-transitive cascade + gated rare-token bridge + adaptive `anchor_radius` (64K NIAH cliff fix) + per-request `fa_window_override` + C2 mechanism gate + **type-gate regime router (default-OFF)**.

Prefill speedup is content-dependent:
- ~12× prefill-TPS on sparse/NIAH cells (Round5),
- ~2.2× wall on dense agentic 64–128K (full-sweep 2026-05-29),
- **3.00× prefill-TPS / 2.91× wall on real dense >32K agentic turns with the router ON** (verified on this branch's tip, 2026-05-31 — see §Verified A/B).

Decode at parity (composition); +35% with the router on (shorter compressed KV).

## Adaptive levers

1. **Adaptive `anchor_radius`** (`4a838dd`): `<32K=2 / 32–64K=4 / ≥64K=8` based on `n_chunks`. Eliminates the 64K NIAH cliff. Override via `PFLASH_COMPRESS_ANCHOR_RADIUS`.

2. **Per-request `fa_window_override`** (`2726608`, Design 1): when pflash compresses, `http_server` sets `fa_window_override = effective_prompt + 256`. Never caps visibility.

3. **C2 mechanism gate** (`2726608` + refactor `548be0a`): if `eff_fa_window ≤ 2 × cfg_.fa_window` (= 4096 default), spec-decode runs. Else AR fallback. Extracted to `server/src/qwen35/c2_gate.h` as a pure tested predicate + 5 unit tests. Zero behavior change.

4. **Effective-size admission gate + keep-ratio guard** (`8c1f37d`): the context-window admission check gates on the **post-compression** effective size, not the raw prompt — a prompt larger than `max_ctx` that compresses to fit is admitted (previously a hard 400) without forcing an oversized KV reservation. Default `--prefill-keep-ratio` 0.05→0.10 (anchors yield ~40% effective; `keep=0.25` over-forces the cascade to ~100% = near-no-op). Verified at >128K: raw 166,911 → eff 71,110 (42.6%), admitted at `max_ctx=139264`.

5. **Type-gate regime router** (`fbc2d41` / `b31544f` / `8fc961b`, **default-OFF** behind `PFLASH_ROUTER_ENABLE`): correct-by-construction admission-time gate. `server/src/common/regime_router.h` is pure stdlib (96 LOC, 59 standalone `-Werror` tests). Logic:
   - **agentic** request (structured `tools` array / `tool_use` / `tool_calls` present) → keep floor **0.25**, transitive cascade **OFF**. The cascade keys on rare-token co-occurrence, and the 151K drafter vocab makes ~every token "rare" → it over-forces dense agentic content to 70–100% keep (a near-no-op). Disabling it + flooring at 0.25 is what delivers the win below.
   - **retrieval** request → full keep, cascade **ON** (cascade is correct for QA/NIAH recall).
   - **empty/degenerate guard**: a compressed agentic turn that returns empty/degenerate skips the keep-bandit and one-shot recovers to full keep next turn (`PFLASH_GUARD_MIN_TOKENS`, default 8).
   - **floor-clamp**: the online keep-bandit (PR #264) is clamped up to the 0.25 agentic floor so it can't undercut the router.

## Verified A/B (router ON vs OFF) — 2026-05-31

Binary built from this branch tip (`8fc961b`), verified to contain both the admission-gate (`dflash::common::check_admission` symbol) and the router. 4 cold-prefill turns on **real organic agentic sessions** (claude-code transcripts, user turns at 34–37K est tokens, native tool blocks + `tools` array → agentic path fires). Fresh server per arm (true cold prefill). Run dir: `bench/2026-05-31_definitive_ab/run_20260531_111334/`.

| Metric | ROUTER OFF | ROUTER ON | Speedup |
|--------|-----------|-----------|---------|
| keep % | 72.3% (44.7–100) | 24.9% (25) | — |
| prefill_TPS | 1248 | 3747 | **3.00×** |
| wall_s | 64.9 | 22.3 | **2.91×** |
| decode_TPS | 20.6 | 27.8 | +35% |
| tool_parse | 4/4 OK | 4/4 OK | held |

All 4 ON turns routed `type=agentic, keep=0.25, cascade=off`. 0 floor-clamps, 0 guard fires. The admission-gate does **not** suppress the cascade over-forcing (OFF still 72% avg), so the router win is real on the shipped base. N=4 cold turns; numbers match the earlier-base measurement (2.85×/2.89×) within noise.

## Reproducibility

```bash
./server/build/dflash_server /path/to/Qwen3.6-27B-Q4_K_M.gguf \
  --draft /path/to/dflash-draft-3.6-q4_k_m.gguf \
  --prefill-drafter /path/to/Qwen3-0.6B-Q8_0.gguf \
  --prefill-compression always --prefill-keep-ratio 0.10 \
  --prefill-threshold 32000 --max-ctx 139264 \
  --chat-template-file qwen3-coder-chat-template.jinja
```

Env: `PFLASH_COMPRESS_ANCHOR_TRANSITIVE=1 PFLASH_DRAFTER_EARLY_EXIT_N=7 PFLASH_DRAFTER_SCORE_LAYERS=7 DFLASH27B_KV_K=tq3_0 DFLASH27B_KV_V=tq3_0`. Router: add `PFLASH_ROUTER_ENABLE=1`.

Bench dashboard: `bench/2026-05-28_adaptive_stack/Round5_4arm_matrix_174641/DASHBOARD.md`. Public evidence: https://github.com/dusterbloom/pflash-evidence.

## Notes

- **The "+47% decode TPS" figure is C_pflash_only vs A_baseline**, not D_composition vs A_baseline. Pflash compression delivers it. Composition decode TPS is at parity-or-slightly-below pflash-only because the C2 gate (correctly) disables spec-decode on compressed KV.
- The router (lever 5) is **default-OFF**; merging it changes no default behavior. It adds opt-in code + 401 LOC of tests.

### DIFF
--- a/server/src/common/score_range.h
+++ b/server/src/common/score_range.h
@@ -1,0 +1,48 @@
+// Pure helper: compute the [score_layer_start, score_layer_end) range for
+// tail-attention scoring given the forward-pass layer limit and the optional
+// SCORE_LAYERS count.
+//
+// Parameters:
+//   n_layer        - total number of layers in the model (e.g. 28)
+//   score_layers   - value of PFLASH_DRAFTER_SCORE_LAYERS (-1 = all)
+//   fwd_layer_limit - number of layers actually computed (== early_exit_n when
+//                    early-exit is active, else n_layer)
+//
+// Semantics: SCORE_LAYERS is interpreted as "how many of the computed layers
+// to score", counted from the END of the forward range [0, fwd_layer_limit).
+// This way SCORE_LAYERS=7 with early_exit_n=7 scores layers [0,7) instead of
+// producing the empty interval [7,7) that the old code yielded.
+#pragma once
+
+#include <algorithm>
+
+namespace dflash::common {
+
+struct ScoreRange {
+    int start; // inclusive
+    int end;   // exclusive
+    int count() const { return end - start; }
+    bool empty() const { return start >= end; }
+};
+
+// Compute the scoring layer range.
+// When early-exit is active, SCORE_LAYERS counts from 0 upward within the
+// computed range [0, fwd_layer_limit), not from the end of the full model.
+inline ScoreRange compute_score_range(int n_layer, int score_layers, int fwd_layer_limit) {
+    // score_layers <= 0 means "use all computed layers"
+    const int effective_n = fwd_layer_limit;
+    int start;
+    if (score_layers > 0 && score_layers < n_layer) {
+        // Clamp: can't request more layers than were computed.
+        int want = std::min(score_layers, effective_n);
+        start = effective_n - want;
+    } else {
+        start = 0;
+    }
+    int end = fwd_layer_limit;
+    // Clamp start to never exceed end.
+    if (start > end) start = end;
+    return { start, end };
+}
+
+} // namespace dflash::common
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,7 @@ fix-plan.md
 # Harness test artifacts
 .harness-work/
 health
+
+# Workdir editor backup suffixes
+*.git-head
+*.pre-pflash-rename
--- a/MORNING_BRIEF.md
+++ b/MORNING_BRIEF.md
@@ -1,0 +1,200 @@
+# PFlash + Adaptive Bandit MVP — Overnight Production Brief
+Date: 2026-05-22
+GPU: NVIDIA GeForce RTX 3090 (24 GB), TQ3_0 KV, Qwen3.6-27B Q4_K_M + Qwen3-0.6B BF16 drafter
+
+---
+
+## Headline Numbers (all empirically validated)
+
+- **ee7 at 128K NIAH: 9.29x** drafter speedup (69.48s → 7.48s) — commit d3fbad3
+- **ee7 at 64K NIAH: 3.68x** (10.41s → 2.83s) — same commit
+- **ee7 at 32K NIAH: 3.51x** (5.05s → 1.44s) — same commit
+- **ee7 on claude_code agentic 28.7K: 3.68x** drafter_fwd (4.31s → 1.17s) — multiclient bench 2026-05-22
+- **ee7 on hermes 14.1K: 3.25x** drafter_fwd (2.18s → 0.67s), accept_rate +11pp (13.8% → 25.0%)
+- **ee7 on opencode 5.4K: 3.46x** drafter_fwd (0.83s → 0.24s)
+- **ee7 broad agentic (ee7_broad Pass B, claude_code ~5.3K): 3.07x** (1.72s → 0.56s)
+- **MVP bandit (day5): Pareto-dominates keep=0.20** — 3s faster wall (16s vs 19s) + 6.5pp higher accept_rate (31.9% vs 25.4%) — commit 1a1a0f6
+
+## Ship-it Config
+
+```
+PFLASH_DRAFTER_EARLY_EXIT_N=7
+PFLASH_DRAFTER_SCORE_LAYERS=7
+```
+
+Recommended default for RTX 3090, all contexts >= 1K. Super-linear speedup with context (3.5x at 32K, 9.3x at 128K) because scoring dominates at long context and ee7 cuts it from 28 layers to 7.
+
+---
+
+## Master Prefill / Decode / Context Table
+
+Sorted by Ctx_in (S, pre-compress tokens) then condition. All RTX 3090 unless noted.
+drafter_fwd = drafter forward+score time (the prefill bottleneck).
+Decode tok/s from spec-decode log lines. Accept = spec-decode accepted ratio.
+Speedup is drafter_fwd baseline / condition drafter_fwd.
+
+### Section A: Named Client Multi-Client Bench (2026-05-22_multiclient_ee7)
+
+Source binary: PFLASH_DRAFTER_EARLY_EXIT_N=7 PFLASH_DRAFTER_SCORE_LAYERS=7
+Config: pflash=always keep=0.05 ddtree=ON budget=16 max_tokens=512
+
+| Client     | Ctx_in | Ctx_kept | Ctx_out | Condition | drafter_fwd | Decode tok/s | Accept    | Quality | Wall  | Speedup_vs_baseline |
+|------------|--------|----------|---------|-----------|-------------|--------------|-----------|---------|-------|---------------------|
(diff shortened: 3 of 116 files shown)


dflash/src/qwen3/qwen3_loader.cpp
```diff
@@ -96,6 +96,26 @@ bool load_qwen3_drafter_model(const std::string & path,
...
+    ggml_type wtype = GGML_TYPE_BF16;
+    {
+        int tidx = gguf_find_tensor(gctx, "blk.0.attn_q.weight");
+        if (tidx >= 0) {
+            // gguf_context created with no_alloc=false builds a ggml_context
+            // internally; use gguf_get_tensor_type to read the stored type.
+            // Fallback: check the tensor size to distinguish Q8_0 from BF16.
+            // BF16 blk.0.attn_q is [1024, 2048] = 2097152 elements * 2 bytes = 4194304 B
+            // Q8_0 blk.0.attn_q is 2097152 * 1.0625 bytes = 2228224 B (32 bytes per 32 elems + 2 byte scale)
+            size_t tsz = gguf_get_tensor_size(gctx, tidx);
```
howard0su: can we use gguf_get_tensor_type ?

dflash/bench/run_agentic_ee7_passbv.py
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env python3
```
howard0su: worth to keep in the repo?

dflash/src/qwen3/qwen3_loader.cpp
dusterbloom: Done — replaced the size-byte heuristic at line 109 with `gguf_get_tensor_type(ctx, tidx)`. Also updated `tidx` declaration from `int` to `int64_t` to match the return type of `gguf_find_tensor`. Build confirmed clean.

dflash/bench/run_agentic_ee7_passbv.py
dusterbloom: Kept. Added a 6-line docstring at the top explaining purpose, when to run it, and which bench dir it writes to. It's the evidence trail for the pass-B agentic validation in this PR.

dusterbloom:
Addressed all cubic + howard0su review feedback. PR body updated with correct Q4_K_M target (was Q3_K_S — factual error) and per-context drafter speedup table (1K-128K + per-client agentic). Howard0su's two fixes applied (gguf_get_tensor_type, passbv docstring). All P1+P2 cubic items fixed across 6 commits.

server/src/common/score_range.h
```diff
@@ -0,0 +1,48 @@
+// Pure helper: compute the [score_layer_start, score_layer_end) range for
```
howard0su, line 1: this is not really model specific, right? can we do the same for other models?

dusterbloom: it can be general. so moving it to  `dflash/src/common/score_range.h` thanks for spotting it!

dusterbloom: Good catch — moved to dflash/src/common/score_range.h in 24f28bc. The helper is already in dflash::common namespace and has zero qwen3-specific code, so it's now trivially reachable from qwen35/gemma4. Wiring those backends into ee7 itself is a separate change (per-backend graph rewrites), but the scoring-clamp logic is now ready to share.

davide221:
why layer 7, and not some other layer? Env naming should be PFlash not DFlash to make it more clear that is the prefill drafter

dusterbloom:
great question @davide221 I started removing from layer 14 then it worked. Then I push and it workd on layer 7. I stopped but actually am benching right now to see if it works on layer 5 or even 3. Let's see

As for the dFlash is a typo, agree to rename all for making sure it is clear this is about prefill!

dflash/bench/run_tier1_proof.py
```diff
@@ -0,0 +1,408 @@
...
+
+def start_server(cond: dict) -> tuple:
+    drafter = cond["drafter"]
+    score_layers = cond["score_layers"]
+    name = cond["name"]
+
+    env = os.environ.copy()
+    env["GGML_CUDA_NO_VMM"] = "1"
+    if score_layers is not None:
+        env["PFLASH_DRAFTER_SCORE_LAYERS"] = str(score_layers)
```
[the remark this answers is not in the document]
dusterbloom: Fixed in 4f7efb9. Set + clear both PFLASH_DRAFTER_EARLY_EXIT_N and PFLASH_DRAFTER_SCORE_LAYERS per condition. Historical tier1_proof results are preserved as the spike that discovered the scoring-clamp bug (fixed in 90e6b27) — not being re-run.

dusterbloom:
@cubic-dev-ai re: `prev_forced` is refreshed at the top of every transitive iteration:

Looked at this carefully — the implementation handles the case you're describing, just differently than the diff suggests. Two pieces:

1. **`anchor_scan.cpp:115-120` seeds the worklist from `forced` directly on `it==0`**, not from `forced && !prev_forced`. So pass-1's anchor results enter the rare-token worklist regardless of the snapshot.

2. **`anchor_scan.cpp:133` makes the worklist self-cascading within a single iteration.** When a rare-token hop forces a new chunk, that chunk is `push_back`'d into the same worklist and processed by the same for-loop (line 121 indexes a growing vector by `wi`). The header comment on lines 107-108 says it: "worklist-driven so cascades within a pass are caught (e.g. hop3 forces hop2 which forces hop1 in one outer iteration)."

So the rare-token closure transitively exhausts in a single outer iteration by design. The purpose of `it > 0` is the **4-gram re-scan** at lines 148-159, which is gated by a fresh `prev_forced` reassignment on line 158 right before `scan_and_force` runs — so iteration N+1's rare-token phase does see new chunks from the 4-gram re-scan via the `it==0` all-forced seeding mechanism applied at the start.

**T6 in `test_anchor_transitive.cpp:298-344` is the executable proof**: it builds a 20-link rare-token chain with `max_iters=25` and `max_forced_count=5`, and asserts the cap engages. If the worklist were empty after `it==0` as your comment describes, the chain couldn't propagate past chunk 0 and the cap would never fire. T6 passes — the in-iteration worklist cascades chunks 0→1→...→N until the cap stops it.

Not a bug — resolving as by design. Will add a clarifying inline comment if the area is touched again.

dusterbloom:
Closing — stale (last updated Jun 9) and CONFLICTING against current main. Please reopen with a rebase if still needed.

PR #275, opened 2026-05-24
feat(drafter): ee3 as production default (depends on #274)

dusterbloom:
## Summary

Changes the recommended early-exit default from N=7 to N=3 based on empirical N-sweep. Depends on PR #274 landing first (introduces `PFLASH_DRAFTER_EARLY_EXIT_N`).

## Source change

`dflash/README.md` — documents ee3 as the recommended default with reproduction instructions.

## Evidence

Results not committed. Reproduce via:

- N-sweep NIAH @ 32K/64K/128K (ee3=3/3 everywhere, 24.3× drafter speedup at 128K vs baseline): `dflash/bench/run_ee_n_sweep.sh`
- 5-client multi-client accept_rate (ee3 +1.2 pp vs ee7, all clients within ±2 pp): `dflash/bench/run_ee_n_multiclient.sh`
- Sweep plan + rationale: `bench/2026-05-25_ee_n_sweep/PLAN.md`

Headline numbers (RTX 3090, Qwen3.6-27B-Q4_K_M, Qwen2.5-0.5B-BF16):
- ee3 drafter speedup: 6.9× @ 32K, 24.3× @ 128K
- accept_rate vs ee7: +1.2 pp mean (claude_code +0.0, hermes -0.4, opencode +1.7, pi +0.0, codex +4.6)
- NIAH 3/3 at 32K, 64K, 128K

## Dependency

Requires PR #274 to merge before this can land (env vars are introduced there).

## Reviewer note

Bench results not committed; run the scripts above to regenerate.

### DIFF
--- a/bench/2026-05-25_ee_n_sweep/PLAN.md
+++ b/bench/2026-05-25_ee_n_sweep/PLAN.md
@@ -1,0 +1,131 @@
+# ee N-sweep Plan: baseline / ee3 / ee5 / ee7
+
+Scope: NIAH @ 32K/64K/128K + 5-client multi-turn bandit session, all four conditions.
+Goal: find the smallest N where quality and accept_rate hold vs ee7.
+
+## Trigger
+
+User says "GPU is free, run the ee_n sweep."
+
+Pre-flight GPU check:
+
+    flock -n /tmp/lucebox-gpu.lock echo ok
+
+Exit 0 means GPU is free. Exit 1 means another process holds the lock — wait or
+ask the user.
+
+## Pre-flight: generate NIAH case files (once, CPU only)
+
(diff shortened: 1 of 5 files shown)


davide221:
NIAH is a simple benchmark, it is a very interesting result. Can you check if we can push deeper this finding on agentic coding use cases/benchmarks? Worth exploring before default to N layers

dusterbloom:
> NIAH is a simple benchmark, it is a very interesting result. Can you check if we can push deeper this finding on agentic coding use cases/benchmarks? Worth exploring before default to N layers

Absolutely 100% agreed. It has to work. Initial testing seems it works but Qwen3.6 tool calling is acting wierd and not really functioning at least on claude-code. Digging deeper on that today

PR #276, opened 2026-05-25
fix(server): Qwen3.6-27B tool calling for claude-code Anthropic path

dusterbloom:
## Summary

Qwen3.6-27B (Q3_K_M / Q4_K_M, non-UD) tool calling silently fails for real claude-code traffic on `/v1/messages` — same model works fine through hermes (OpenAI shape). Confirmed community bug ([HF discussion](https://huggingface.co/Qwen/Qwen3.6-27B/discussions/13), [reference fix repo](https://github.com/abysslover/qwen36_tool_calling_failure)).

Symptoms reproduced on a captured real claude-code request (22.8K-token system prompt + 24 tools):
- Pre-fix: model emits 16 leading newlines, then `` ```bash\nls -lS / ``, then `finish=stop`. No structured `tool_use`.
- Post-fix: model emits proper `content_block_start type:tool_use, name:Bash` with `{"command":"ls -lhS /tmp 2>/dev/null | head -30"}`, `finish=tool_calls`.

## What's in the PR

Six coordinated server-side fixes, sequenced because each exposes the next:

1. **`normalize_tools_for_qwen()`** — convert Anthropic `input_schema` → OpenAI `parameters` shape so the Jinja template sees the schema key Qwen3-Coder was trained on.
2. **`<bash>` / `<read>` / `<write>` / `<edit>` / `<ls>` / `<grep>` / `<glob>` native-tag parser fallback** — Pattern 6 in `tool_parser.cpp`. Gated on \`tools\` being present (avoids fabricating phantom calls from prose).
3. **`scrub_schema_metadata()`** — strip JSON-Schema metadata (\`\$schema\`, \`additionalProperties\`, \`\$defs\`, \`oneOf\`/\`anyOf\`/\`allOf\`/\`not\`) before the Jinja \`render_extra_keys\` macro turns them into garbage XML tags that hallucinate function names like \`<function=cls>\`.
4. **`truncate_description()`** — cap each tool description at 500 bytes (paragraph break → sentence boundary → UTF-8-safe hard cut). Claude-code embeds 12KB of \"use other tools instead\" recipes inside Bash's own description, which steered Qwen to pick Write.
5. **Closed `<think>` prefill in Jinja renderer** when thinking is disabled — mirrors the hardcoded Qwen renderer. Handles trailing whitespace variants in the template. (Diagnosed by Codex.)
6. **`resolve_param_alias()`** — Q3 quant hallucinates short forms (\`cmd\`→\`command\`, \`path\`↔\`file_path\`, \`expr\`→\`expression\`, etc.). Resolves the emitted parameter name to the schema's canonical name via case-insensitive direct match plus a small alias table.

Plus PR #271's \`find_tool_start\` is extended to recognize the 7 native tags (rebase-resolved cleanly).

## Test plan

- [x] 1013 LOC diff: 337 production + 676 tests (~2:1 test:prod ratio)
- [x] New unit coverage: native-tag parsing, scrub recursion incl. combinators, description truncation incl. UTF-8 2/3/4-byte safety, alias resolution, no-tools-no-fabrication gate, Anthropic→OpenAI shape normalization
- [x] All sources compile clean (only pre-existing fread warnings)
- [x] End-to-end verified against captured real claude-code request: \`finish=tool_calls\`, structured \`tool_use\` block, valid \`command\` argument
- [x] Underwent self-review cycle (cubic-style); 3 P1 + 5 P2 findings addressed in commit \`129e606\`
- [ ] Local server interactive verification: launch with \`--chat-template-file <Unsloth Qwen3-Coder template>\` against your dflash_server build, run a multi-turn claude-code session, confirm agentic loop completes
- [ ] Regression check on hermes (OpenAI shape) — should be unaffected; Pattern 6 gate ensures no behavior change when tools shape is already OpenAI

## Recipe to deploy

\`\`\`bash
dflash_server <Qwen3.6-27B GGUF> \\
  --host 127.0.0.1 --port 18099 --max-ctx 131072 \\
  --draft <dflash draft GGUF> \\
  --chat-template-file <Unsloth Qwen3-Coder Jinja>
\`\`\`

Plus \`~/.claude/settings.json\` env block:
\`\`\`json
{\"env\": {
  \"CLAUDE_CODE_ATTRIBUTION_HEADER\": \"0\",
  \"CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC\": \"1\",
  \"CLAUDE_CODE_DISABLE_TELEMETRY\": \"1\"
}}
\`\`\`
(\`CLAUDE_CODE_ATTRIBUTION_HEADER=0\` must be in JSON, not shell \`export\` — see Unsloth docs.)

## Known caveats

- **Q3_K_S works** with all 6 fixes. **Q4_K_M** structurally also works but model prefers writing Python scripts over running Bash directly — Qwen3-Coder training bias amplified at higher precision.
- **UD-Q3_K_XL / UD-Q4_K_XL** (Unsloth Dynamic, calibrated on tool-calling examples) likely needs fewer of these patches — not tested in this PR since download wasn't available in-session.
- **Native-tag fallback (\`<bash>\` etc.)** is opportunistic: \`<edit>\` and \`<write>\` can't satisfy Anthropic's Edit/Write tool schemas without \`file_path\` / \`old_string\` / \`new_string\`. They're kept for parser correctness but won't produce useful calls; flagged in code comments.

### DIFF
--- a/server/src/server/chat_template.cpp
+++ b/server/src/server/chat_template.cpp
@@ -419,13 +419,25 @@ std::string render_chat_template_jinja(
         // for tool use. Mirror the hard-coded behavior here when the rendered
         // prompt ends with a bare assistant generation prompt.
         if (!enable_thinking) {
-            static constexpr char kAssistantPrefix[] = "<|im_start|>assistant\n";
-            static constexpr char kNoThinkPrefill[] = "<think>\n\n</think>\n\n";
-            if (rendered.size() >= sizeof(kAssistantPrefix) - 1 &&
-                rendered.compare(rendered.size() - (sizeof(kAssistantPrefix) - 1),
-                                 sizeof(kAssistantPrefix) - 1,
-                                 kAssistantPrefix) == 0) {
-                rendered += kNoThinkPrefill;
+            // The hard-coded Qwen renderer follows <|im_start|>assistant with a
+            // closed <think> block to put the model in non-thinking decode mode.
+            // Tolerate template variants that emit extra trailing whitespace
+            // after the assistant marker (single \n, double \n\n, trailing
+            // space). Strategy: trim trailing whitespace, check for the BARE
+            // assistant marker (no newline), then re-emit marker + prefill.
+            static constexpr char kAssistantBare[]    = "<|im_start|>assistant";
+            static constexpr char kAssistantPrefill[] = "<|im_start|>assistant\n<think>\n\n</think>\n\n";
+            size_t trim_end = rendered.size();
+            while (trim_end > 0) {
+                char c = rendered[trim_end - 1];
+                if (c != ' ' && c != '\t' && c != '\n' && c != '\r') break;
+                --trim_end;
+            }
+            const size_t blen = sizeof(kAssistantBare) - 1;
+            if (trim_end >= blen &&
+                rendered.compare(trim_end - blen, blen, kAssistantBare) == 0) {
+                rendered.resize(trim_end - blen);
+                rendered += kAssistantPrefill;
             }
         }
 
--- a/server/src/server/chat_template.cpp
+++ b/server/src/server/chat_template.cpp
@@ -411,7 +411,25 @@ std::string render_chat_template_jinja(
         jinja::runtime rt(ctx);
         jinja::value results = rt.execute(*prog);
         auto parts = jinja::runtime::gather_string_parts(results);
-        return parts->as_string().str();
+        std::string rendered = parts->as_string().str();
+
+        // The hard-coded Qwen renderer appends a closed think prefill when
+        // thinking is disabled. Some Qwen3.6 Jinja templates omit that final
+        // assistant suffix, which leaves the model in the wrong decoding state
+        // for tool use. Mirror the hard-coded behavior here when the rendered
+        // prompt ends with a bare assistant generation prompt.
+        if (!enable_thinking) {
+            static constexpr char kAssistantPrefix[] = "<|im_start|>assistant\n";
+            static constexpr char kNoThinkPrefill[] = "<think>\n\n</think>\n\n";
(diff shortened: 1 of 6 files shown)


dflash/src/server/chat_template.cpp
```diff
@@ -293,7 +293,37 @@ std::string render_chat_template_jinja(
...
         auto parts = jinja::runtime::gather_string_parts(results);
-        return parts->as_string().str();
+        std::string rendered = parts->as_string().str();
+
+        // The hard-coded Qwen renderer appends a closed think prefill when
+        // thinking is disabled. Some Qwen3.6 Jinja templates omit that final
+        // assistant suffix, which leaves the model in the wrong decoding state
+        // for tool use. Mirror the hard-coded behavior here when the rendered
+        // prompt ends with a bare assistant generation prompt.
+        if (!enable_thinking) {
```
howard0su: this is general code path. based on your comment, we should check the model or arch here.

dusterbloom:
Addressed in 0e3c79a: added `arch_hint` (ChatFormat) parameter to `render_chat_template_jinja`, gated the closed-think prefill block on `arch_hint == ChatFormat::QWEN3`. Call site in http_server.cpp passes `chat_format_` so Laguna and Gemma4 Jinja templates are untouched.

dusterbloom:
Addressed in 0e3c79a: added `arch_hint` (ChatFormat) parameter to `render_chat_template_jinja`, gated the closed-think prefill injection on `arch_hint == ChatFormat::QWEN3`. Call site in http_server.cpp passes `chat_format_` so Laguna and Gemma4 Jinja templates are untouched.

dusterbloom:
Closing — stale (last updated May 28) and CONFLICTING against current main. Please reopen with a rebase if still needed.

PR #284, opened 2026-05-27
fix(draft): read rope_theta from config.json in safetensors loader

a-huk:
Commit 03aeda5 correctly removed the hardcoded 10M default and made rope_theta come from GGUF metadata, but did not update the safetensors loader path. With rope_theta defaulting to 0.0f, powf(0, -small) = +inf in the rope kernel, producing NaN in every Q and K and causing all draft hidden states to be NaN — spec-decode fails immediately.

Fix: read rope_theta from config.json in the same directory as the .safetensors file. HF models always ship config.json alongside weights. Load fails hard if the key is missing, matching the GGUF loader's behaviour.

### DIFF
--- a/server/src/draft/draft_safetensors_loader.cpp
+++ b/server/src/draft/draft_safetensors_loader.cpp
@@ -30,6 +30,7 @@
 #include "internal.h"
 
 #include <climits>
+#include <cmath>
 #include <cstdint>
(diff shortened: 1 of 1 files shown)


davide221:
@a-huk thanks for the contribution!

PR #285, opened 2026-05-27
feat(lucebox): docker stack + CLI + bench/profile + harness + luce-bench in-tree

easel:
This PR turns Lucebox into a one-command local inference deployment and ships the two tools that operate it: **`lucebox`** (the host CLI that runs and tunes the server) and **`luce-bench`** (the benchmark + grading framework that measures it). All three ship together so a fresh box goes from nothing to a tuned, benchmarked server with a single install.

The three pieces, what each is, and how to use it:

---

## 1. Docker — the server image

A CUDA 12.8 image (`ghcr.io/luce-org/lucebox-hub:cuda12`) that builds the dflash server and bundles `server/`, `lucebox/`, `harness/`, and `luce-bench/`. The entrypoint dispatches `serve` (default), `benchmark`, any `lucebox` subcommand, or `shell`. An in-container autotune fallback picks VRAM-tiered defaults and resolves the draft GGUF by target architecture (gemma4 → gemma drafter, qwen3.6 → dflash-draft-3.6).

**Use it directly:**
```bash
docker run --rm --gpus all -p 8080:8080 \
  -v ~/.local/share/lucebox/models:/opt/lucebox-hub/server/models \
  ghcr.io/luce-org/lucebox-hub:cuda12
# OpenAI + Anthropic-compatible API on :8080
curl -s http://localhost:8080/v1/models
```

Image tags: `:cuda12`, `:vX.Y.Z-cuda12`, `:X.Y-cuda12`, `:sha-<short>-cuda12`. Built and pushed by `.github/workflows/docker.yml`; `docker-bake.hcl` has a `cuda13` slot ready.

## 2. `lucebox` — the host CLI

`lucebox.sh` is the host-side wrapper (deps: `docker` + `nvidia-smi` only). It probes the host, writes a tuned `config.toml`, runs the container as a user-systemd service, and delegates provisioning/workloads to the in-container Python CLI (`models`, `autotune`, `profile`, `smoke`, `config`, the client drivers).

**Stand a server up:**
```bash
lucebox check            # driver / docker / NVIDIA Container Toolkit / VRAM / systemd / WSL2 probe
lucebox pull             # docker pull the cuda12 image
lucebox models download  # pull target + DFlash draft GGUFs  (verbs: list, download)
lucebox autotune         # VRAM-tiered DFLASH_* defaults → ~/.lucebox/config.toml  (autotune --sweep picks a winner empirically)
lucebox install          # install the user-systemd unit
[...]

PR #287, opened 2026-05-27
feat(gemma4): wire prefill/decode timing into GenerateResult

easel:
Mirrors `qwen35_backend.cpp` commit 3b80fa8.

Before this, every gemma4 `/v1/chat/completions` response reported
`usage.timings.prefill_ms = 0` and `decode_tokens_per_sec = 0`, so
benchmarks couldn't measure prefill/decode wall time on gemma4 targets.
`qwen35_backend.cpp` got this instrumentation when 3b80fa8 landed;
gemma4 was missed in that pass.

## What's in this PR
- `server/src/gemma4/gemma4_backend.cpp` (+18 lines):
  - `generate()`: wrap `do_prefill` in `steady_clock` and stamp
    `result.prefill_s`. Same for the decode loop → `result.decode_s`.
  - `restore_and_generate()`: same for the delta-prefill + decode path.

`http_server.cpp` already reads `result.prefill_s` / `result.decode_s`
via the `GenTimings` struct and surfaces them in `usage.timings`; no
server-side changes needed.

## Validation
Sindri gemma-4-26b ds4-eval bench (PR #285 territory) now reports
non-zero `decode_tokens_per_sec` per case (typical: 30–80 tok/s with
spec-decode, 12–20 tok/s AR-only).

### DIFF
--- a/server/src/gemma4/gemma4_backend.cpp
+++ b/server/src/gemma4/gemma4_backend.cpp
@@ -632,7 +632,13 @@ GenerateResult Gemma4Backend::generate(const GenerateRequest & req,
 
     cache_.cur_pos = 0;
 
+    // Timing: prefill wall-clock — surfaced in usage.timings.prefill_ms
+    // by http_server.cpp via the GenTimings struct from sse_emitter.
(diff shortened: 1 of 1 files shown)


howard0su (approved):
LGTM

PR #288, opened 2026-05-27
fix(laguna): correct chat template + eos_chat_id fallback to </assistant>

easel:
### `server/src/server/chat_template.cpp` (+71 / -14)
Rewrites the `ChatFormat::LAGUNA` branch of `render_chat_template()`.
The previous template was a DeepSeek copy-paste using tokens
(`〈|begin▁of▁sentence|〉`, `〈|User|〉`, `〈|Assistant|〉`) that **do
not exist in the laguna vocab** — the model saw replacement-character
garbage in its prompt and degenerated into echoing the user message
back with `<���Assistant���>` artifacts.

New template matches `poolside/Laguna-XS.2/chat_template.jinja`
(verified 2026-05-25):

```
〈|EOS|〉<system>
{system_content_or_default}
</system>
<user>
{user_content}
</user>
<assistant>
<think>     ← if enable_thinking
</think>    ← if NOT enable_thinking (empty think block, model jumps to answer)
```

Tokens 18/19/23/24/25/26 are special (`<think>`, `</think>`,
`<assistant>`, `</assistant>`, `<tool_call>`, `</tool_call>`).
`<user>` / `<system>` / `</user>` / `</system>` are not added-tokens
and tokenize as regular bytes — matches the upstream template.

Includes a default system prompt (matches Poolside's template
fallback), tool-response wrapping, and `add_generation_prompt` with
`enable_thinking` toggle.

### `server/src/laguna/laguna_target_loader.cpp` (+10 / -1)
`eos_chat_id` falls back to `24` (`</assistant>`) when not in GGUF.
Laguna's GGUF only ships `tokenizer.ggml.eos_token_id` (id 2 = `〈|EOS|〉`);
the chat-template end-of-turn marker is the unrelated `</assistant>` (id 24).
Without this fallback, the decoder check `next == eos_chat_id` matches
-1 (impossible), so the model never stops mid-stream — it just emits
`</assistant>` and keeps going, re-greets the user, and answers again
until `max_tokens`.

### DIFF
--- a/server/src/server/chat_template.cpp
+++ b/server/src/server/chat_template.cpp
@@ -147,27 +147,87 @@ std::string render_chat_template(
     }
 
     case ChatFormat::LAGUNA: {
-        // Laguna/DeepSeek format:
-        //   <｜begin▁of▁sentence｜>system content
-        //   <｜User｜>user content<｜Assistant｜>
-        result = "<｜begin▁of▁sentence｜>";
-        for (const auto & msg : messages) {
-            if (msg.role == "system") {
+        // Laguna XS.2 format (verified against
+        // poolside/Laguna-XS.2/chat_template.jinja, 2026-05-25):
+        //
+        //   〈|EOS|〉<system>
+        //   {system_content_or_default}
+        //   </system>
+        //   <user>
+        //   {user_content}
+        //   </user>
+        //   <assistant>
+        //   <think>      ← if enable_thinking (gen prompt)
+        //   </think>     ← if NOT enable_thinking (gen prompt — empty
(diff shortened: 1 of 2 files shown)


server/src/server/chat_template.cpp
```diff
@@ -147,27 +147,87 @@ std::string render_chat_template(
     }
 
     case ChatFormat::LAGUNA: {
```
howard0su, line 149: we may want to move chatformat related logic into modelbackend as more and more different logic around.

howard0su (approved):
LGTM.

PR #308, opened 2026-05-29
fix(server): route Qwen3.6/Laguna think-mode reasoning to reasoning_content channel

easel:
## Problem

For models whose chat template appends `<think>` to the prompt suffix when
`enable_thinking` is honored (Qwen3.6, Laguna), the model generates
reasoning tokens directly into the output stream with no opening tag.
`SseEmitter` was hardcoded to start in `StreamMode::CONTENT` and only
transitioned to `REASONING` when it saw a `<think>` literal in the
generated stream. Result: reasoning text leaked into `content`,
`reasoning_content` stayed empty, and the `</think>` close tag appeared
verbatim in the user-visible answer.

`parse_reasoning()` already supported a `started_in_thinking=true` mode
(see `reasoning.h`) — the receiving end of the contract existed. The
sending end never threaded the flag through. Both halves were correct
in isolation; no caller connected them.

## Fix

Plumb a `started_in_thinking` bit from the chat-template renderer to the
SSE emitter so the emitter starts in the correct stream mode when the
prompt itself pre-opens the reasoning channel.

1. `render_chat_template` and `render_chat_template_jinja` now return
   `PromptRenderResult { text, started_in_thinking }`.
   - QWEN3 / LAGUNA builtins set `started_in_thinking = true` iff
     `enable_thinking && add_generation_prompt`.
   - GEMMA4 always sets `false` (its reasoning channel is opened by the
     model emitting `<|channel>`, which the server already forwards as
     `<think>`).
   - Jinja path: suffix-sniffs the rendered prompt for a trailing
     `<think>` and logs a `[WARN]` when the sniff fires, so
     template/model-card mismatches surface at runtime.

2. `SseEmitter` takes an `initial_mode` constructor parameter (defaulted
   to `CONTENT`). When `REASONING`, the Anthropic first
   `content_block_start` is emitted as `thinking` so SDK clients don't
   see a spurious empty `text` block. The existing one-time
   `<think>`-strip guard is preserved so a redundant model-emitted
   opener (template/card mismatch) is still removed cleanly.

3. `http_server.cpp` threads `render_result.started_in_thinking` through
   `ParsedRequest` into the emitter's `initial_mode`. Streaming and
   non-streaming paths share the emitter, so both response shapes are
   fixed by one wire.

### C++ unit (`test_server_unit`, runs in CI)

1656 assertions, 0 failures. Coverage:
- Renderer end: QWEN3/LAGUNA/GEMMA4 set `started_in_thinking` correctly
  across `enable_thinking` ∈ {on, off} and `add_generation_prompt` ∈
  {on, off}; Jinja suffix-sniff positive and negative cases.
- Emitter end: with `initial_mode=REASONING`, tokens before `</think>`
  route to `reasoning_content`; unclosed channel keeps everything in
  reasoning; redundant `<think>` opener is stripped; Anthropic first
  block is `thinking`, not `text`.
- **Wire integration**: three tests chain `render_chat_template →
  propagate flag → SseEmitter(initial_mode=…) → emit tokens` end-to-end
  for QWEN3-on, LAGUNA-on, and QWEN3-off. Mirrors the production wiring
  at `http_server.cpp`.

### CPU-only HTTP integration (`test_stub_integration.py`, runs in CI)

Closes the original "CI can't load a real model" gap that allowed this
bug to ship. A new test driver — `spike_no_gpu_http_server` — links
against the production `dflash_common` library but instantiates a
deterministic `StubModelBackend` instead of a CUDA-backed model. The
driver runs with `CUDA_VISIBLE_DEVICES=""`; the real Qwen3.6 tokenizer
loads from a stripped GGUF fixture (`server/test/fixtures/
qwen3.6-tokenizer.gguf`, 11MB via LFS, vocab/merges/specials only, no
tensor weights).

`StubModelBackend::generate()` decodes the request prompt to text via
the real tokenizer, matches it against a JSON scenario file (longest
`prompt_suffix` wins), then replays the scripted token stream through
the production `req.on_token` / `io.on_token` callbacks. Streaming
behavior comes from the production code path — the stub feeds the same
SseEmitter that GPU traffic does.

The new pytest suite (4 tests, 0.43s) covers:
- Qwen3.6 enable_thinking non-streaming OpenAI: reasoning_content
  populated, content has no `<think>` leakage.
- Qwen3.6 enable_thinking non-streaming Anthropic: thinking block
  precedes text block; no raw tags.
- Qwen3.6 enable_thinking streaming OpenAI: per-token
  `reasoning_content` deltas before `</think>`, per-token `content`
  deltas after.
- Qwen3.6 enable_thinking streaming Anthropic: first
  `content_block_start` has `type:"thinking"` (not `text`);
  `thinking_delta` events precede `text_delta` events.

Scenarios are JSON files in `server/test/scenarios/` and trivially
extensible — any deterministic prompt→token-stream pairing can be
captured as a file. Synthesizing edge cases (unclosed `</think>`,
redundant openers, error responses, finish_reason variants) is just
authoring more files.

### Manual smoke (existing `test_server_integration.py`, run by deploy)

The Python integration test for a live GPU server has had its reasoning
assertions tightened: prior versions used "doesn't crash" assertions
that would have passed on the broken code. Now require non-empty
`reasoning_content` and no raw `<think>`/`</think>` in either channel.
This is the deploy-time check; the CPU integration test above is the
pre-merge check.

## Files

```
server/src/server/chat_template.{h,cpp}        +91/-6
server/src/server/http_server.{h,cpp}          +30/-8
server/src/server/sse_emitter.{h,cpp}          +25/-4
server/test/test_server_unit.cpp              +330/-10
server/scripts/test_server_integration.py      +32/-6

# New CPU integration infrastructure
server/test/stub_model_backend.{h,cpp}        +130 new
server/test/scenario_store.{h,cpp}            +220 new
server/test/spike_no_gpu_http_server.cpp      +110 new
server/test/test_stub_integration.py          +220 new
server/test/scenarios/qwen3_enable_thinking_basic.json  new
server/test/fixtures/qwen3.6-tokenizer.gguf   11MB (LFS) new
server/test/scripts/strip_gguf_to_tokenizer.py +90 new
.github/workflows/ci.yml                       +14
.gitattributes                                 +1
server/CMakeLists.txt                          +25
```

No model-card JSON or schema changes. No production code outside
`server/src/server/`.

## Out of scope

- Optional `thinking_prompt_pre_opens_think` model-card override field.
  The renderer is the source of truth; an override is belt-and-suspenders.
- A "record mode" that wraps the real GPU backend and captures
  `(prompt, emitted tokens, finish)` triples into scenario files
  automatically. Hand-authored scenarios are sufficient for the bug
  classes we care about today.
- Additional scenarios (Laguna, Gemma4, multi-turn, tool calls inline
  with reasoning). The infrastructure supports these — each is a new
  JSON file.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,3 +5,4 @@ assets/banner.png -filter -diff -merge -text
 *.jpeg filter=lfs diff=lfs merge=lfs -text
 *.mp4 filter=lfs diff=lfs merge=lfs -text
 *.webm filter=lfs diff=lfs merge=lfs -text
+*.gguf filter=lfs diff=lfs merge=lfs -text
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
             -DCMAKE_BUILD_TYPE=Release
           cmake --build build --target \
             test_dflash test_generate test_flash_attn_sparse \
-            dflash_server test_server_unit spike_no_gpu_http_server \
+            dflash_server test_server_unit replay_http_server \
             -j$(nproc)
 
       - name: Run C++ server unit tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: recursive
+          lfs: true
           token: ${{ secrets.SUBMODULE_PAT || secrets.GITHUB_TOKEN }}
 
       - uses: Jimver/cuda-toolkit@v0.2.35
@@ -55,7 +56,7 @@ jobs:
             -DCMAKE_BUILD_TYPE=Release
           cmake --build build --target \
             test_dflash test_generate test_flash_attn_sparse \
-            dflash_server test_server_unit \
+            dflash_server test_server_unit spike_no_gpu_http_server \
             -j$(nproc)
 
       - name: Run C++ server unit tests
@@ -69,6 +70,19 @@ jobs:
         # [tool.uv.sources]) and pflash drags in setuptools. Megakernel is
         # in the optional `megakernel` extra so its build does NOT run yet.
         run: uv sync --frozen
+
+      - name: Run CPU integration tests (stub backend, no GPU)
+        # End-to-end exercise of HttpServer + render_chat_template +
+        # SseEmitter with a deterministic stub model backend. No GPU
+        # required: the spike driver runs under CUDA_VISIBLE_DEVICES=""
+        # and the tokenizer fixture is a stripped Qwen3.6 GGUF (metadata
+        # only). Covers the regression class from PR #308 end-to-end —
+        # streaming and non-streaming, OpenAI and Anthropic formats.
+        env:
+          CUDA_VISIBLE_DEVICES: ""
+        run: |
+          uv run --frozen --with pytest --with requests \
+            pytest -v server/test/test_stub_integration.py
 
       - name: Build megakernel via uv sync (sm_75)
         env:
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -769,6 +769,38 @@ if(DFLASH27B_TESTS)
         endif()
     endif()
 
+    # ─── Spike: HttpServer with stub backend, no GPU bound ──────────────
+    # Validates that the server constructs and serves /health without a
+    # GPU — the foundation for a CPU-only integration test driver. Links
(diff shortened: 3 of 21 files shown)


easel:
Closing in favor of #343, which extracts the unique CPU-only HTTP-server test rig from this branch (8 new test sources + LFS fixture + additive CI/CMake wiring) without the stale reverts that had accumulated here.

The reasoning-channel routing fix that this PR originally landed is already on `main` (commit db79a7d) and continues to be exercised by #341. #343 is purely additive test infrastructure with no production-code changes.

PR #311, opened 2026-05-30
fix(qwen35): correct + transparent prefix cache with spec-decode

OmarB97:
## Summary

The dflash prefix cache (`--prefix-cache-slots > 0`) produced corrupted / empty generations on cache hits, which forced it off (`--prefix-cache-slots 0`) and lost the cross-request system-prompt prefill speedup. This fixes three distinct bugs so cache-on output is **bit-identical to no-cache**, then the cache can be safely re-enabled.

### Bugs fixed

1. **Stale draft feature mirror on restore.** `restore_target_cache` restores KV / recurrent state / `target_feat`, but `restore_and_generate` never re-synced the draft-side feature mirror. The draft then emitted token `-1`, `verify_batch`'s embed rejected it, and the whole turn returned empty. Fix: re-sync the mirror tail from the restored `target_feat` (as `do_prefill` already does).

2. **Prefill chunk-shrink → FP-nondeterministic state divergence.** Prefill shrank the ubatch chunk to land exactly on `snap_pos`, which realigned every later chunk and changed GPU batch sizes vs the no-cache path. The resulting FP drift flipped near-tie greedy decisions, so cache-on diverged from cache-off (e.g. an early stop with no tool call). Fix: keep uniform chunks; snapshot at a chunk boundary instead of shrinking.

3. **Snapshot rounding broke cross-request reuse.** Snapshotting at the first boundary `>= snap_pos` pushed the snapshot to prompt end (the full prompt incl. the user message) when the next boundary exceeded `prompt_len`. A later request sharing only the system-prompt prefix then restored a state polluted with the prior user message → garbage → instant EOS. Fix: round **down** to the largest boundary `<= snap_pos` (chunk-aligned AND strictly inside the shared prefix).

Also hardens the `fa_window` struct defaults (2048 → 0) across the qwen35/dflash config structs. `fa_window>0` sparsifies the full-attention layers at decode, dropping the system prompt on long prompts and breaking tool calls; `0` (full attention) is the only correct value for this full-attn model and matches gemma4. The CLI already passes `0`; this just removes the footgun.

## Verification (Qwen3.6-27B-Q4_K_M, RTX 5090)

- **Correctness oracle:** 6/6 cache-on output **bit-identical** to cache-off across single-turn + 5 growing multi-turn conversations.
- **Cross-request** (shared system prompt, different user messages — the realistic case): 4/4 **bit-identical** to cache-off.
- **~18× faster** warm requests (27.75s → 1.5s) by skipping the 17k-token system-prompt prefill.

## Files

- `server/src/qwen35/qwen35_backend.cpp` — the three correctness fixes (mirror re-sync; uniform-chunk + round-down snapshot).
- `server/src/qwen35/{qwen35_backend,qwen35_daemon,qwen35_layer_split_adapter}.h`, `server/src/common/{backend_factory,dflash_layer_split_runtime}.h` — `fa_window` default `2048 → 0`.

### DIFF
--- a/server/src/common/backend_factory.h
+++ b/server/src/common/backend_factory.h
@@ -42,7 +42,7 @@ struct BackendArgs {
     int             chunk        = 512;
 
     // qwen35-specific speculative decode options
-    int             fa_window        = 2048;
+    int             fa_window        = 0;  // 0 = full attention. qwen3.6 full-attn layers must see the whole context; a finite window drops the system prompt/tools -> breaks tool calls.
     int             kq_stride_pad    = 32;
     int             draft_swa_window = 0;
     int             draft_ctx_max    = 4096;
--- a/server/src/common/dflash_layer_split_runtime.h
+++ b/server/src/common/dflash_layer_split_runtime.h
@@ -19,7 +19,7 @@ namespace dflash::common {
 
 struct LayerSplitRuntimeConfig {
     int kq_stride_pad  = 32;    // KQ_MASK_PAD default; 256 when TBQ KV active
-    int fa_window      = 2048;  // flash-attn sliding window
+    int fa_window      = 0;  // 0 = full attention. qwen3.6 full-attn layers must see the whole context; a finite window drops the system prompt/tools -> breaks tool calls.
     int draft_ctx_max  = 4096;  // draft context cap
     int draft_swa_window = 0;   // draft SWA window (0 = disabled)
 };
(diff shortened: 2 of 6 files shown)


server/src/common/backend_factory.h
```diff
@@ -42,7 +42,7 @@ struct BackendArgs {
     int             chunk        = 512;
 
     // qwen35-specific speculative decode options
-    int             fa_window        = 2048;
+    int             fa_window        = 0;  // 0 = full attention. qwen3.6 full-attn layers must see the whole context; a finite window drops the system prompt/tools -> breaks tool calls.
```
howard0su, line 45: @davide221 Is it time to completely remove fa_window configure? Although it gives perf but breaking tools call make it not useful.

howard0su (approved):
LGTM

PR #314, opened 2026-05-30
fix(common): retry empty spec-decode output through AR

OmarB97:
## Summary

dflash speculative decoding can emit **EOS as the very first token** on certain agentic decision turns, returning an **empty completion** (0 tokens, `finish=stop`, no `tool_call`). The agent loop then has nothing to run and stalls.

At temperature 0, spec-decode output must match autoregressive (AR) greedy output. On the captured qwen35 cases it does not, so this is treated as a spec-decode correctness failure for the empty-output class.

## What changed

Follow-up to review feedback asking whether the fallback can live in a common upper layer:

- Added shared `ModelBackend` wrappers for `generate` and `restore_and_generate` that detect `ok && spec_decode_ran && tokens.empty()` and retry once with `GenerateRequest::force_ar_decode`.
- Routed the daemon loop and HTTP server through those common wrappers.
- Made spec-capable backends honor `force_ar_decode` instead of duplicating the empty-output retry at backend call sites: qwen35, gemma4, layer-split, and qwen35moe hybrid.
- Preserved spec-decode telemetry on the retry result so accept-rate/session logic still sees that spec decode was attempted.
- Added unit coverage for the common generate and restore retry wrappers.

## How to review

Start with `server/src/common/model_backend.h`: this is now the shared retry policy. Then check the call sites in `daemon_loop.cpp` and `http_server.cpp`, and finally the backend-specific changes, which should only be mechanism: skip spec decode when `force_ar_decode` is set.

## Evidence

Original qwen35 root-cause validation was isolated with a reproducible stateless lane (`dflash-nocache`, no prefix-cache cross-request state). On 55 real captured agentic turns:

- 9 turns produced empty output under spec-decode deterministically.
- The same 9 turns produced correct non-empty output on the AR path and on stock llama.cpp Q6.
- After the fallback, 9/9 empty turns produced non-empty output byte-identical to AR.
- Full 55-turn sweep: empty count 9 -> 0, zero regressions to empty, tool-call set unchanged.
- Prefix-cache correctness oracle remained 6/6 bit-identical.

## Verification

- Local hygiene: `git diff --check HEAD~1..HEAD` passed.
- Local prerequisites fixed for follow-up work: `cmake` is installed locally (`cmake version 4.3.3`) and local submodules are initialized recursively.
- Clean GPU-host verification on taro in isolated worktree `/tmp/lucebox-pr314-verify-20260530T220140` at `f4a0d8b567f3318def29dc2b2a119b9eda63cd5b`.
- CUDA/toolchain proof: CUDA 13.3 `nvcc` on RTX 5090 (`NVIDIA GeForce RTX 5090`, driver `595.71.05`).
- Configure passed:
  `CUDACXX=/usr/local/cuda-13.3/bin/nvcc CUDAToolkit_ROOT=/usr/local/cuda-13.3 cmake -S server -B server/build-pr314 -DCMAKE_BUILD_TYPE=Release -DCMAKE_CUDA_ARCHITECTURES=120 -DCMAKE_CUDA_COMPILER=/usr/local/cuda-13.3/bin/nvcc -DCUDAToolkit_ROOT=/usr/local/cuda-13.3`
- Unit target passed:
  `cmake --build server/build-pr314 --target test_server_unit -j$(nproc)` and `./server/build-pr314/test_server_unit` -> `1620 assertions, 0 failures`, `ALL PASSED`.
- Server target passed:
  `cmake --build server/build-pr314 --target dflash_server -j$(nproc)` -> built `server/build-pr314/dflash_server`.

## Risks / gaps

- Scope is still the empty-output failure only. Other stalls that emit an action preamble and then stop with no tool call remain separate.
- The failing case now retries at the common request boundary, so it may pay a full AR retry cost. Normal non-empty spec-decode turns do not hit the retry path.
- The underlying first-token spec-decode divergence is treated, not root-caused, in the verify path.

## Collaborators

- Omar Baradei (operator) — reported the failure, directed the fix.
- Claude (Opus 4.8) — initial reproducible eval, root-cause isolation, and qwen35 fallback.
- Codex — common-layer refactor responding to review feedback.

### DIFF
--- a/server/src/common/daemon_loop.cpp
+++ b/server/src/common/daemon_loop.cpp
@@ -292,7 +292,7 @@ int run_daemon(ModelBackend & backend, const DaemonLoopArgs & args) {
             req.do_sample = do_sample;
             req.stream    = false;
 
-            auto result = backend.generate(req, io);
+            auto result = backend.generate_with_empty_spec_fallback(req, io);
             if (!result.ok) {
                 std::printf("err %s\n", result.error.c_str());
                 std::fflush(stdout);
@@ -349,7 +349,7 @@ int run_daemon(ModelBackend & backend, const DaemonLoopArgs & args) {
             req.snap_pos  = snap_pos;
             req.snap_slot = snap_slot;
 
-            auto result = backend.restore_and_generate(slot, req, io);
+            auto result = backend.restore_and_generate_with_empty_spec_fallback(slot, req, io);
             if (!result.ok) {
                 std::printf("err %s\n", result.error.c_str());
                 std::fflush(stdout);
@@ -398,7 +398,7 @@ int run_daemon(ModelBackend & backend, const DaemonLoopArgs & args) {
             req.snap_pos  = snap_pos;
             req.snap_slot = snap_slot;
 
-            auto result = backend.generate(req, io);
+            auto result = backend.generate_with_empty_spec_fallback(req, io);
             if (!result.ok) {
                 io.emit(-1);
                 std::printf("err %s\n", result.error.c_str());
--- a/server/src/common/layer_split_backend.cpp
+++ b/server/src/common/layer_split_backend.cpp
@@ -103,7 +103,9 @@ GenerateResult LayerSplitBackend::run_from_state(const GenerateRequest & req,
             return result;
         }
         auto t_decode_start = std::chrono::steady_clock::now();
-        const bool ok = adapter_->can_dflash_decode()
+        const bool use_dflash = !req.force_ar_decode && adapter_->can_dflash_decode();
+        if (use_dflash) result.spec_decode_ran = true;
+        const bool ok = use_dflash
             ? adapter_->decode_dflash(req.prompt, base_pos, last_tok, req.n_gen,
                                       result.tokens, out_io)
             : adapter_->decode_ar(last_tok, base_pos + (int)req.prompt.size(), req.n_gen,
--- a/server/src/common/model_backend.h
+++ b/server/src/common/model_backend.h
@@ -11,6 +11,7 @@
 #pragma once
 
 #include <cstdint>
+#include <cstdio>
 #include <functional>
 #include <string>
 #include <vector>
@@ -100,6 +101,10 @@ struct GenerateRequest {
     const std::vector<int32_t> * hint_tokens = nullptr;
     // Optional thinking-budget hook — see BudgetHook docs above.
     BudgetHook                 budget_hook;
+    // Common retry knob. Upper layers set this after a speculative decode
+    // path returns success but emits no tokens, so each backend can route the
+    // retry through its existing AR path without copying retry policy.
+    bool                       force_ar_decode = false;
 };
 
 struct GenerateResult {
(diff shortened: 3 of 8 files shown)


howard0su:
The idea is great. I am wondering if we can fix this in the upper layer (common). We have several models now, shall we fix this in every backend?

OmarB97:
@howard0su addressed in f4a0d8b.

I moved the fallback policy up to the common `ModelBackend` boundary instead of keeping the zero-token check in the qwen35 backend:

- `ModelBackend::generate_with_empty_spec_fallback` / `restore_and_generate_with_empty_spec_fallback` now detect `ok && spec_decode_ran && tokens.empty()` and retry once with `GenerateRequest::force_ar_decode`.
- `daemon_loop.cpp` and `http_server.cpp` call those common wrappers.
- qwen35, gemma4, layer-split, and qwen35moe hybrid now only honor the common `force_ar_decode` mechanism, so future backends do not need to copy the empty-output retry policy.
- The retry result keeps the first attempt's spec-decode telemetry so existing accept-rate/session logic still sees that spec decode ran.
- Added unit coverage for both common generate and restore retry paths.

Verification is now complete too. I fixed the environment blockers from the earlier note: local `cmake` is installed, local submodules are initialized, and the GPU host is reachable via the Linux `taro` alias. On a clean isolated taro worktree at f4a0d8b with CUDA 13.3 / RTX 5090 sm_120:

- `git diff --check HEAD~1..HEAD` passed locally.
- CMake configure passed with explicit CUDA 13.3 compiler/toolkit root.
- `cmake --build server/build-pr314 --target test_server_unit -j$(nproc)` passed.
- `./server/build-pr314/test_server_unit` passed: `1620 assertions, 0 failures`, `ALL PASSED`.
- `cmake --build server/build-pr314 --target dflash_server -j$(nproc)` passed and built `server/build-pr314/dflash_server`.
- GitHub Actions are green as well: Build, uv workspace, and cubic all passed.

I also updated the PR body with the current verification details so the stale Mac-session caveat is gone.

server/src/common/daemon_loop.cpp
```diff
@@ -349,7 +349,7 @@ int run_daemon(ModelBackend & backend, const DaemonLoopArgs & args) {
             req.snap_pos  = snap_pos;
             req.snap_slot = snap_slot;
 
-            auto result = backend.restore_and_generate(slot, req, io);
+            auto result = backend.restore_and_generate_with_empty_spec_fallback(slot, req, io);
```
howard0su, line 352: this change is free. we can use restore_and_generate as function name and default to enable this.

howard0su (approved):
LGTM

server/src/common/daemon_loop.cpp
OmarB97: Opened follow-up PR for this: https://github.com/Luce-Org/lucebox-hub/pull/319

It keeps `ModelBackend::generate` / `restore_and_generate` as the public default call names, moves backend-specific work behind `*_impl` overrides, and leaves the empty spec-decode retry enabled centrally by default.

Verified on `taro`: `test_server_unit` passed with 1620 assertions / 0 failures, and `dflash_server` builds.

PR #324, opened 2026-05-31
[codex] Retry dflash generations with no visible output

OmarB97:
## Summary

The cached dflash path could count an EOS/EOT-only spec-decode result as a successful completion. That produced no visible streamed or non-streamed text while still allowing prefix-cache confirmation/continuation to treat the decode as valid.

This adds an explicit `GenerateResult::empty_visible_output` signal, sets it for Qwen35 EOS/EOT-only spec-decode results, retries those results through AR decode in the common backend wrapper, and gates HTTP prefix-cache confirmation on visible emitted output rather than raw completion token count.

This upstream branch is based directly on `Luce-Org/main` and contains only the visible-output fix. The fork-side integration PR remains OmarB97/lucebox-hub#7.

## Validation

- Built `test_server_unit` on `taro` with CUDA 13.3
- `/home/omar/ai/lucebox-worktrees/visible-empty-dflash-retry/server/build/test_server_unit` (`1634 assertions, 0 failures`)
- Built and installed patched `dflash_server` to `/home/omar/ai/lucebox-hub/server/build/dflash_server` on `taro`
- Direct live smoke through `llama-swap`: `dflash` returned `OK` with visible content

### DIFF
--- a/server/src/common/daemon_loop.cpp
+++ b/server/src/common/daemon_loop.cpp
@@ -30,9 +30,9 @@ namespace dflash::common {
 // ── DaemonIO ────────────────────────────────────────────────────────────
 
 bool DaemonIO::should_cancel() const {
-    if (cancelled) return true;
+    if (cancelled.load(std::memory_order_relaxed)) return true;
     if (is_cancelled && is_cancelled()) {
-        cancelled = true;
+        cancelled.store(true, std::memory_order_relaxed);
         return true;
     }
     return false;
@@ -44,7 +44,7 @@ void DaemonIO::emit(int32_t v) const {
     // Call the token callback for non-sentinel tokens.
     if (on_token && v >= 0) {
         if (!on_token(v)) {
-            cancelled = true;
+            cancelled.store(true, std::memory_order_relaxed);
             return;
         }
     }
--- a/server/src/common/daemon_loop.cpp
+++ b/server/src/common/daemon_loop.cpp
@@ -29,7 +29,18 @@ namespace dflash::common {
 
 // ── DaemonIO ────────────────────────────────────────────────────────────
 
+bool DaemonIO::should_cancel() const {
+    if (cancelled) return true;
+    if (is_cancelled && is_cancelled()) {
+        cancelled = true;
+        return true;
+    }
+    return false;
+}
+
 void DaemonIO::emit(int32_t v) const {
+    if (should_cancel()) return;
+
     // Call the token callback for non-sentinel tokens.
     if (on_token && v >= 0) {
         if (!on_token(v)) {
--- a/server/src/common/dflash_spec_decode.cpp
+++ b/server/src/common/dflash_spec_decode.cpp
@@ -86,6 +86,7 @@ bool run_dflash_spec_decode(
 
     auto t_dec0 = std::chrono::steady_clock::now();
     while (n_generated < n_gen) {
+        if (io.should_cancel()) break;
         const int need_commit_budget = n_gen - n_generated;
 
         // ── Build noise input for draft ────────────────────────────────────
@@ -109,10 +110,12 @@ bool run_dflash_spec_decode(
         // ── Draft compute (local or remote) ───────────────────────────────
         const float * draft_hidden_host = nullptr;
         if (use_remote_draft) {
+            if (io.should_cancel()) break;
             if (!remote_draft->propose(committed, draft_ctx, noise_embed, remote_hidden)) {
                 std::fprintf(stderr, "dflash-spec remote draft propose failed\n");
                 return false;
             }
+            if (io.should_cancel()) break;
             draft_hidden_host = remote_hidden.data();
         } else {
             if (!build_draft_step(draft_sg, draft_weights, /*lm_head=*/nullptr, draft_backend,
@@ -138,6 +141,7 @@ bool run_dflash_spec_decode(
             ggml_backend_tensor_set(draft_sg.positions_k, pos_k.data(), 0,
                                     sizeof(int32_t) * pos_k.size());
             auto st = ggml_backend_graph_compute(draft_backend, draft_sg.gf);
+            if (io.should_cancel()) break;
             if (st != GGML_STATUS_SUCCESS) {
                 std::fprintf(stderr, "dflash-spec draft compute %d\n", (int)st);
                 return false;
@@ -174,6 +178,7 @@ bool run_dflash_spec_decode(
             std::fprintf(stderr, "dflash-spec snapshot_kv failed\n");
             return false;
         }
+        if (io.should_cancel()) break;
 
         int verify_last_tok = -1;
         if (!target.verify_batch(draft_tok, committed, verify_last_tok, &target_tok)) {
@@ -209,6 +214,7 @@ bool run_dflash_spec_decode(
             std::fprintf(stderr, "dflash-spec restore_kv failed\n");
             return false;
         }
+        if (io.should_cancel()) break;
 
         std::vector<int32_t> replay_tok((size_t)commit_n);
         for (int i = 0; i < commit_n; i++) {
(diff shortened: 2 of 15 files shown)


OmarB97:
Updated this PR with the dflash disconnect root-cause fix found during the 2026-05-31 Hermes hang investigation.

Root cause: the server only observed client disconnects when SSE writes happened. If Hermes/client disconnected while dflash was still in pre-header prompt handling, prefill, or first-token work, the backend kept CPU/GPU work alive and the single worker stayed unavailable.

Changes:
- added request-level socket watching before render/tokenize/enqueue
- made tokenizer encode cooperatively cancellable
- propagated DaemonIO cancellation through backend prefill/spec-decode/AR paths
- fixed the empty-visible retry test to call the retry wrapper

Verification on taro:
- /tmp/lucebox-dflash-cancel-test: cmake --build server/build --target test_server_unit -j 8 && ./test_server_unit
- result: 1641 assertions, 0 failures
- /tmp/lucebox-dflash-cancel-test: cmake --build server/build --target dflash_server -j 8

OmarB97:
Addressed the unresolved review issue from cubic about cancellation masking draft backend compute failures.

What changed:
- Added a shared completed-compute classifier so `ggml_backend_graph_compute()` failures win over pending cancellation.
- Applied it to the common DFlash spec-decode path and sibling qwen35/qwen35moe cancellation-after-compute sites.
- Added unit coverage for failure-vs-cancel ordering.

Verification:
- Clean taro clone `/tmp/lucebox-pr324-review-fix`: CMake configure/build for `test_server_unit` succeeded.
- `server/build/test_server_unit` passed: 1647 assertions, 0 failures.

Mirror tracking:
- The prior fork mirror PR OmarB97/lucebox-hub#8 had already merged before this review-fix commit, so I opened replacement mirror PR OmarB97/lucebox-hub#9 for the current head.

OmarB97:
## Reviewer notes (Hermes dispatch review)

Overall: Good scope and approach. The cooperative cancellation infrastructure, disconnect detection, and empty-visible-output retry all address real failure modes from the "sudden death" reports.

### Concerns (non-blocking)

**1. \`qwen35_empty_visible_output\` returns \`false\` for empty token vector** (\`qwen35_backend.cpp:43-49\`)

The function returns \`false\` when \`tokens.empty()\`, meaning \`empty_visible_output\` is never set when zero tokens are generated. The broader retry condition handles this via \`result.tokens.empty() \|\| result.empty_visible_output\`, so it is not broken in practice — but the flag is semantically wrong for the empty case. Consider returning \`true\` for empty tokens.

**2. Potential data race on \`DaemonIO::cancelled\`** (\`daemon_loop.cpp:32-39\`)

\`should_cancel()\` is \`const\` and writes to \`mutable bool cancelled\`. The \`cancelled\` field is also written from the generation thread (via \`on_token\` returning false) while \`should_cancel()\` itself may be called from the same thread or from the disconnect watcher thread. If two threads call \`should_cancel()\` concurrently after \`cancelled=false\`, both can observe the same value and both can write \`cancelled = true\` — classic data race on non-atomic bool. Consider making \`cancelled\` a \`std::atomic<bool>\` or adding a mutex.

**3. \`RequestDisconnectWatcher\` performance under load** (\`http_server.cpp:84-124\`)

Every request spawns a dedicated thread that polls the socket every 100ms. Under high concurrency (many concurrent SSE streams), this adds significant thread overhead. For short requests (count_tokens, small prompts), the thread may spin only once or twice before being joined. Consider a shared poller or edge-triggered epoll/kqueue approach for production deployments.

**4. \`\`POLLRDHUP\`\` fallback to 0 on macOS** (\`http_server.cpp:27-29\`)

On macOS (which lacks POLLRDHUP), the poll flags silently drop the clean-disconnect detection. The fallback to recv(MSG_PEEK) still works, but the semantics are different: POLLRDHUP fires immediately on FIN, while MSG_PEEK only detects it after the peer has actually sent 0 bytes. This means macOS has slightly slower disconnect detection. Consider documenting this or using kqueue-style detection on macOS.

**5. Test: \`test_tokenizer_encode_honors_cancel_callback\`** (\`test_server_unit.cpp:2469-2485\`)

The test uses an empty \`Tokenizer()\` with no vocabulary loaded. The callback always returns \`true\` (always cancelled). The test expects a \`TokenizationCancelled\` exception, but an empty tokenizer may return early with an empty vector before the cancellation check fires (depending on the fast path in \`encode\`). The test may pass because of the early check in \`encode\`, but it is fragile — it depends on the specific code path taken. Consider loading a minimal vocabulary or using a mock tokenizer.

### Suggestions

- The \`visible_output_seen\` gate on snapshot/continued/disk cache confirmation is a good change — it prevents caching responses where the client saw no visible output. However, ensure the thinking-only case (model emits only `\`\`\`<antThinking>` and `\`\`\`</antThinking>\`\`\` with no text after) is handled as intended: thinking tags DO set \`visible_output_seen = true\`, so a thinking-only response WILL still confirm the snapshot. This may or may not be desired.
- The \`classify_daemon_compute_result\` priority (failure > cancel) is correct and well-documented.
- The speculative decode cancellation points are well-placed — before draft compute, after verify, after restore_kv. This prevents wasted GPU time when the client has already disconnected.

OmarB97:
Addressed the Hermes dispatch-review follow-ups in `3f44fc8`.

What changed:
- `qwen35_empty_visible_output()` now treats an empty token vector as empty visible output, matching the common retry semantics.
- `DaemonIO::cancelled` is now an atomic latch with copy/assignment support, and cancellation writes now use explicit atomic load/store.
- Replaced per-request disconnect watcher threads with a shared disconnect poller for generation routes; `/v1/messages/count_tokens` avoids watcher registration on its short path.
- Documented the macOS/BSD `POLLRDHUP` fallback behavior next to the poll flags.
- Made `test_tokenizer_encode_honors_cancel_callback` cancel on the second poll so it exercises an inner tokenizer cancellation checkpoint rather than only the first guard.
- Documented that thinking delimiters intentionally count as visible stream output for cache confirmation.

Verification:
- Local hygiene: `git diff --check` passed.
- Local syntax-only check passed for `server/src/common/daemon_loop.cpp` and `server/src/server/http_server.cpp` with repo include paths.
- Clean taro scratch clone `/tmp/lucebox-pr324-codex-reviewfix`: configured CUDA explicitly with `/usr/local/cuda`, built `test_server_unit`, and ran it successfully: `1647 assertions, 0 failures`.
- GitHub after push: `uv workspace` and `cubic` checks are green; full build job is still running as of this comment.

OmarB97:
Audit update (2026-06-09): upstream has since landed the qwen35 slice of this work in 74c1483 (same retry-on-no-visible-output behavior, evolved logging) and its own cooperative-cancellation design (`DaemonIO.cancelled` + `InferenceObserver`) that supersedes the `should_cancel()` approach in f540b29.

What remains uniquely valuable here is the cross-backend generalization: gemma4, laguna, qwen3, and qwen35moe still have no `empty_visible_output` handling on upstream main. I'm going to rework this branch to (1) drop the superseded cancellation commits, and (2) re-express the cross-backend retry on top of upstream's current `model_backend.h` contract, then re-verify on real hardware before marking it ready for review again.

OmarB97:
Correction to my earlier audit comment after isolating the hunks: the empty-visible-output work in this PR only ever covered qwen35 (helper + both decode paths + tests) — which is exactly what landed in 74c1483 — and the gemma4/laguna/qwen3/qwen35moe changes here were entirely the should_cancel() cancellation design that upstream superseded with DaemonIO.cancelled + InferenceObserver. There is no unlanded content left, so closing.

Note for the record: other backends are still covered for fully-empty outputs by the backend-agnostic tokens.empty() clause in the retry condition; the only theoretical gap is all-EOS-token outputs on non-qwen35 backends, which would be new work if it ever shows up in practice.

PR #326, opened 2026-06-01
feat(server): soft-close thinking termination via logit-ratio peek

easel:
## Summary

Adds an operator-configurable **soft-close** dial that lets the AR
decode loop terminate `</think>` early when the close-token logit comes
within a configured probability ratio of the most-likely-token logit.
Default `0.0` keeps current behaviour byte-identical; any positive
value activates the third `close_kind="soft"` taxonomy entry that
`docs/specs/thinking-budget.md` §7 has reserved since the v2 work.

**Motivation.** Gemma 4 26B decodes at ~30 tok/s through up to 15 488
phase-1 thinking tokens (~8 min wall-clock / case) before the hard-cap
hook fires. Spot-checks of close-token logits in the late phase-1
window show `</think>` riding at 10-60 % of the chosen-token
probability for thousands of steps — i.e. the model is *near* ready
to close. A soft-ratio dial in the `[0.05, 0.5]` range can reclaim
30-50 % of those tokens at no quality loss. Sweep methodology and
recommended dial values land in a follow-up PR (out of scope here).

## What's in the PR

1. **Plan doc** (commit 1) at
   `docs/experiments/soft-close-thinking-termination-plan.md`, with a
   verbatim codex review and per-finding dispositions.
2. **Implementation** (commit 2):
   - `BudgetHook::soft_close_min_ratio` + `GenerateResult::soft_forced_close`.
   - Soft-close comparator + state machine in `qwen35_backend.cpp::do_ar_decode`.
   - CLI flag `--think-soft-close-min-ratio <F>` + startup banner line.
   - Per-request override via Anthropic envelope's
     `thinking.soft_close_min_ratio`, clamped to the server ceiling.
   - `close_kind="soft"` emitted in `finish_details` when the soft path
     fired (precedence: soft > hard > natural on tie — see plan §12).
   - Spec doc updated.

## Mechanism — zero-cost when disabled

The AR loop already materialises the full logits row to CPU each step
for the sampler. The comparator reads two scalars from that buffer
and runs `logit[close0] - logit[chosen] >= log(min_ratio)` — no graph
modification, no extra GPU work. When `min_ratio == 0` the outer
guard short-circuits before any work happens. Generation determinism
is byte-identical to pre-PR with the dial off.

For the math: `prob[i]/prob[j] = exp(logit[i] - logit[j])`, so
comparing `logit_diff >= log(min_ratio)` is identical to comparing
`prob_ratio >= min_ratio` but skips the softmax. Numerically stable
in fp32 for typical LLM logit ranges (codex confirmed §3.4).

## Scope (v1)

- **Qwen3.5/3.6 only.** Gemma 4 and Laguna's AR loops follow the same
  pattern (full logits already on CPU per step) but get their own PRs
  to keep the diff reviewable.
- **Pure AR.** Spec-decode's verify/accept inner loop reads only the
  argmax-of-target — soft-peek there would require graph extension.
  Spec-decode tails off into AR before the budget edge, so the soft
  trigger still fires correctly on the AR tail.
- **Single-token close peek.** Multi-token `</think>` sequences (Laguna)
  peek only the first id; the existing inject machinery handles the
  rest. Codex agreed this is the right engineering trade-off (review
  Q3).

## Codex review

Sent to the live Gemma 4 26B service via `lucebox codex`; verbatim
review and per-finding dispositions are recorded in plan §11. Codex
verdict: **PROCEED WITH CHANGES**. The single critical finding (Q5,
per-request clamp logic broken when server_default=0) is fixed: the
operator-disabled-server case now silently ignores per-request opt-in
attempts rather than enabling them via the clamp loophole.

## Tests

`server/test/test_server_unit.cpp` gains 12 new test functions
(~135 new assertions). The comparator is small and inline in
`model_backend.h::soft_close::should_fire`, called from a unit-test
mirror of the AR loop's close-state machine. No GPU required.

Standalone smoke-test of the comparator math (uncommitted) confirmed
135/135 assertions pass before committing.

## Test plan

- [ ] Unit tests pass in a fresh build: `cmake --build server/build --target test_server_unit && ctest -R server_unit`
- [ ] Existing thinking-budget integration tests in `luce-bench/tests/test_client_thinking_budget.py` pass unchanged (default dial=0 → behaviour byte-identical).
- [ ] Operator sets `--think-soft-close-min-ratio 0.1` in a smoke deploy, confirms server banner shows the new line, runs a Qwen3.6 thinking-enabled probe, and verifies a `finish_details.close_kind="soft"` appears for at least one case.
- [ ] Empirical sweep (follow-up PR) quantifies token savings vs quality across a sweep bracket of `min_ratio ∈ {0.05, 0.1, 0.2, 0.5}` on the existing coding-agent-loop probes.

## Follow-ups (NOT in this PR)

- Gemma 4 26B soft-close port (same mechanism, separate backend file).
- Laguna soft-close port (separate backend file).
- `dflash.think_soft_close_min_ratio` knob in the lucebox python CLI repo + autotune sweep bracket entry.
- Empirical sweep + recommended dial values per model.
- Spec-decode joint-peek (if Laguna multi-token false-positives warrant it).

### DIFF
--- a/docs/experiments/soft-close-thinking-termination-plan.md
+++ b/docs/experiments/soft-close-thinking-termination-plan.md
@@ -1,0 +1,774 @@
+# Soft-close: logit-ratio-driven early `</think>` termination
+
+Status: PLAN — pre-implementation. No code changes in this commit.
+
+Branch: `feat/soft-close-thinking-termination`
+Base: `Luce-Org/lucebox-hub:main` @ `8305b6c`
+Affected files (anticipated):
+- `server/src/common/model_backend.h` — extend `struct BudgetHook` and `struct GenerateResult`.
+- `server/src/qwen35/qwen35_backend.cpp` — soft-close peek inside the AR decode loop (`do_ar_decode`).
+- `server/src/server/http_server.cpp` — wire CLI/per-request soft ratio into `BudgetHook`; flip `close_kind` to `"soft"` when the soft path fired.
+- `server/src/server/http_server.h` — add `soft_close_min_ratio` to `ServerConfig` + per-request override field.
+- `server/src/server/server_main.cpp` — `--think-soft-close-min-ratio` CLI flag + startup banner.
+- `server/test/test_server_unit.cpp` — comparator + state-machine unit tests.
+- `docs/specs/thinking-budget.md` — note `close_kind="soft"` is now live and document the dial.
+
+Explicitly NOT touched (parallel sub-agent owns these on
+`fix/sse-emitter-content-mode-tool-parse`):
+- `server/src/server/sse_emitter.cpp`
+- `server/src/server/tool_parser.cpp`
+
+## 1. Problem statement
+
+The thinking-budget envelope (`docs/specs/thinking-budget.md`) today
+exposes two `close_kind` values:
+
+- `natural` — the model emitted `</think>` on its own.
+- `hard`    — the Level-2 hook injected `</think>` at the budget edge
+  because the model would otherwise burn the entire phase-1 budget.
+
+In practice, Gemma 4 26B decodes at ~30 tok/s through its full 15 488
+phase-1 cap (≈8 minutes wall-clock per case) on hard prompts whose
+reasoning the model has effectively finished much earlier. Sampled
+spot-checks show the close-token logit `logit[</think>]` riding very
+close to the argmax for hundreds or thousands of steps before the
+budget edge — i.e. the model is *near* ready to close, sampling just
+doesn't pick `</think>` because some content token has a marginally
+higher logit. Spec §7 already reserves a third `close_kind="soft"` value
+for "a future voluntary-close mechanism (logit-biasing the model toward
+`</think>` as the cap approaches, before forcing it)" — this PR turns
+that reservation on, with a different (cheaper, more legible) mechanism
+than logit biasing.
+
+## 2. Goal — bounded, opt-in, zero-cost-when-disabled
+
+Add a single configurable knob — `soft_close_min_ratio ∈ [0, 1]` — that,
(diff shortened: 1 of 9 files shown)


easel:
## Follow-up commit `f7e8d6f8`: probe/inject split + min_thinking_tokens floor

Empirical validation on qwen3.6-27b showed the dial inert at every sampled ratio (`prob_ratio < 1e-8` across 12,888 thinking-token steps). Root cause: `BudgetHook::close_token_ids` was used for both the soft-close peek probe AND the inject sequence — for qwen3.6's trained-hint sidecar that meant peeking the "Considering" lead-in (id 79939) instead of the `</think>` marker (id 248069).

This commit:

1. **Splits probe-vs-inject in BudgetHook**. New `soft_close_probe_ids` field (empty = legacy fallback). `server_main` detects the marker substring inside the hint and tokenizes it in isolation.
2. **Adds `--think-soft-close-min-tokens N`** false-positive floor (default 0 = disabled).
3. **Adds `--debug-thinking-logits`** trajectory diagnostic for tuning future curves.

After the fix, re-probe shows `</think>` reaches argmax (`max_diff = 0.000`, `prob_ratio = 1.0`) at 66-94% of natural reasoning length across 5 diverse prompts. **9.7 nat improvement** restores the mechanism to its designed regime.

5 new unit tests (test_server_unit: 1780 assertions, 0 failures). PR #331 (which had this fix on a bad base) is closed in favor of consolidation here.

easel:
Superseded by the 8-PR split. The soft-close mechanism (BudgetHook fields, qwen35+gemma4 maybe_soft_close lambdas, CLI flags, tests) is now in **PR #339**. Its prerequisite layer-split + c2_gate.h is in **PR #336**. Closing in favor of reviewing those two.

PR #345, opened 2026-06-05
feat(optimizations): Luce Spark — calibrated hot/cold expert residency

davide221:
Adds `optimizations/spark/`, the placement + caching **product layer** on top of the merged hot/cold MoE offload engine (`server/src/common/moe_hybrid_*`), **plus the single-graph hybrid-decode engine** (commit `31ce93c`) that makes offloaded laguna decode run at near all-GPU speed.

## What Spark is
A 33B-total MoE only fires ~8 of 256 experts/token, but naive hot/cold offload still hits the CPU tier ~36% of the time if it keeps the wrong experts resident. Spark:
1. **Calibrates** the hot set from the traffic you actually serve (replays real agent sessions, accumulates per-(layer,expert) routing → placement profile).
2. **Caches** the per-session tail in a bounded ring of spare GPU slots (LRU swap), driving cold-misses down in fixed VRAM.
3. **Single-graph decode (new):** runs the whole token in one fused graph so the offloaded path keeps near all-GPU throughput.

## Single-graph hybrid decode engine (commit 31ce93c)
The laguna hybrid-offload decode was rebuilt to run the whole token in **one fused graph** (`laguna_step_hybrid` + `build_laguna_moe_block_hybrid`): the routed FFN is served from the hot expert stack via per-layer residency LUTs (batched into one H2D / one D2H), instead of 40 separate per-layer graphs. That removes ~3.2 ms/token of per-layer host glue — the real cap on the multi-graph fused path. **Default-on** under offload (`DFLASH_LAGUNA_NO_SINGLE_GRAPH=1` to fall back). The bounded expert-cache swap is now **async on `cudaStreamPerThread` with a pinned cold store** (`ggml_backend_cuda_host_buffer_type`) so cold-tail re-fetch overlaps compute; shared storage, so qwen35moe benefits too.

### Measured (RTX 3090, single-token decode)
| Config | tok/s | vs all-GPU |
|---|---:|---:|
| laguna all-resident | 118 | 100% (96/96 bit-exact) |
| **laguna true-60% residency** | **97–101** | **~85–92%** (was 76) |
| qwen3.6-35B-A3B 60% offload (AR) | 98–106 | ~92–99% |

## Calibration results (RTX 3090, Laguna-XS.2 Q4_K_M, held-out Claude Code sessions)

| Config | tok/s | % all-GPU | VRAM |
|---|---:|---:|---:|
| All-GPU | 118 | 100% | 18.8 GiB |
| Uniform 60% | 66 | 56% | 10.6 GiB |
| **Spark calibrated 60%** | **81** | 69% | 10.6 GiB |
| **Spark + cache + single-graph** | **97–101** | **~85–92%** | 14.6 GiB |

Calibration: split by session (no leakage). Full tables in `RESULTS.md`.

## Honest notes
- The earlier "~0% cold-hit" figure was a **counter bug** (the fused branch never incremented the cold counter). Real cold-hit at true-60% is ~6–8%, **capacity-bound**: recency (LRU) is already at the Belady optimum, so eviction tuning can't beat it. The single-graph + async-pinned overlap *hides* that re-fetch under compute (that's the 76→~100 gain), it doesn't eliminate it.
- Token-level expert prediction (transition / router-score / pre-gate) caps at ~53% recall@8 — the experts that newly fire are driven by the *current* token's hidden evolution, invisible to any prior-token signal. Draft-as-predictor (MoE-SpeQ style) is the open lever, documented as a follow-up.

## Known follow-up (refactor debt)
`laguna_step_hybrid` / `build_laguna_moe_block_hybrid` currently duplicate ~150 lines of `build_laguna_graph` / `build_laguna_layer` / `build_laguna_moe_block_full` (path isolation while iterating). Unifying via a hybrid descriptor threaded through the existing builders is a clean follow-up to remove the drift risk.


howard0su:
I suggest to try 8GB which is most common VRAM for gamers.

davide221:
@cubic-dev-ai review

davide221:
Thanks @cubic-dev-ai. Triage of the engine-code findings:

- **P0 `step_graph_free` stale `hot_local_lut`/`valid_lut`** — valid, fixed in `44be2d2` (reset both on free). Only reachable on fused-gate-up archs under `DFLASH_LAGUNA_FUSE_FFN`; shipped targets (laguna-xs2, qwen35moe) use separate tensors so were unaffected, but it was a real latent use-after-free.
- **P0 `moe_hybrid_cache_swap_in` "no-op for fused gate-up"** — false positive. The full function handles both layouts (`if (st.fused_gate_up) {...} else {...}` + `copy_slice` for each); the flagged hunk was just my async/pinned change to `copy_slice`. Verified the cache works on laguna-xs2 (cold-hit drops with cache enabled).
- **P2s** (n_expert default, hot_count>0 spare gating, `--spark-vram` warm-reserve, unchecked `save_csv`, double profile save, `DFLASH_IGNORE_EOS` by-presence) — acknowledged; mostly pre-existing in the merged offload engine, low-impact. Tracking as cleanup alongside the documented refactor (unify `laguna_step_hybrid` with `build_laguna_graph`).

Issue #259, opened 2026-05-22
Dflash issue on V100 build


merbanan:
```
uv run --directory dflash python scripts/run.py --prompt "def fibonacci(n):"
[run] prompt 14 tokens, streaming up to 256 tokens, max_ctx=512
[cfg] seq_verify=0 fast_rollback=1 ddtree=1 budget=22 temp=1.00 chain_seed=1 fa_window=2048 draft_swa=0 draft_ctx_max=4096 draft_feature_mirror=0 peer_access=0 target_gpu=0 draft_gpu=0
[loader] eos_id=248046 eos_chat_id=-1
[target] target loaded: layers [0,64) output=1, 850 tensors on GPU 14.99 GiB, tok_embd 682 MiB CPU-only (q4_K)
[draft]  loaded
[prompt] 14 tokens
[prefill] token-seg ubatch=256
[prefill] 14 tokens in 0.10 s, last_tok=0
[migrate] 242.89 ms
[dbg sib step 0] N=23 accept=1 walked_sib=0
  walk: 0
[step 0] committed=14 last_tok=0 tree_N=23 accept=1 next=-1

[timing] per-step averages over 0 steps (ms):
  draft_build    0.27
  draft_copyfeat 0.07
  draft_set      0.02
  draft_compute  10.16
  draft_bridge   0.01
  draft_logits   9.47
  snapshot_ssm   0.00
  verify_build   1.68
  verify_set     0.25
  verify_compute 93.74
  verify_logits  0.00
  accept         0.04
  restore_ssm    0.00
  replay_build   0.00
  replay_set     0.00
  replay_compute 0.00
  replay_logits  0.00
  mirror_sync    0.00
  ----- sum     115.71

[dflash] generated 0 tokens in 0.116 s  ->  0.00 tok/s
[dflash] 0 draft steps, accepted=0/0 (0.0% per step), avg commit/step=0.00
[dflash] output tail: 248045 846 198 727 73111 1393 1590 248046 198 248045 74455 198 248068 198 0 
!ggml_cuda_init: found 1 CUDA devices (Total VRAM: 32501 MiB):
  Device 0: Tesla V100-PCIE-32GB, compute capability 7.0, VMM: yes, VRAM: 32501 MiB

[run] generated 1 tokens


uv run --directory dflash python scripts/bench_llm.py
[bench] target    = /media/per/work/tmp/lucebox-hub/dflash/models/Qwen3.6-27B-Q4_K_M.gguf
[...]

PR #346, opened 2026-06-05
fix(test): chain DFlash verify/replay produce valid logits at all positions

davide221:
## Summary

The chain (non-DDTree) DFlash decode path in `test_dflash` aborts after step 0 on Qwen3.6 and generates 0 tokens. Running the same prompt with `--ddtree` works, so the bug is specific to the chain verify/replay. This fixes it; chain mode now decodes correctly.

## Root cause

Two pre-existing bugs in `server/test/test_dflash.cpp`, both independent of the DDTree path:

1. **Causal-mask stride mismatch (the main bug).** The batched verify and the legacy replay call `build_causal_mask(...)` **without** `kv_pad_override`, so the mask buffer is strided by `align_up(win_len, kq_stride_pad)`, while the mask *tensor* is allocated with stride `align_up(max_ctx + n_tokens, kq_stride_pad)`. Only query row 0 lands at the correct offset; rows 1.. read an unwritten region, so attention (and therefore logits) is zeroed for every verify/replay position > 0. DDTree already passes the override; the chain path did not. Empirically, logits `maxabs` was `p0=26.5, p1..=0.0`; with the override, `p1..=26.x`.

2. **Broken GPU argmax read.** The chain verify read `sg.argmax_tokens`, whose CUDA `ggml_argmax` returns `-1` after position 0 even when logits are valid (the same defect fixed for DDTree in `1b3882d`). Switched to reading full `sg.logits` + CPU `argmax_f32` per position.

With (1) alone the verify is correct but the replay still corrupts `last_tok`; both are needed.

## Validation

RTX 3090 (sm_86) / CUDA 12, Qwen3.6-27B Q4_K_M target + 3.6 DFlash draft, on `main` (`f59f2a3`):

- **Before:** `[step 0] accept_n=2 bonus=-1`, then silent abort, **0 tokens**.
- **After:** 7 draft steps, **accepted=41/112 (36.6%/step)**, ~50 tok/s, coherent output identical to the DDTree/server path through the deterministic prefix (diverges only post-EOS under `temp=1.0` sampling).

## Notes

This is the chain test-harness path; the production server uses DDTree (unaffected). Issue #259 is a *different* (V100/Volta MMA) problem.

### DIFF
--- a/server/test/test_dflash.cpp
+++ b/server/test/test_dflash.cpp
@@ -3728,7 +3728,16 @@ int main(int argc, char ** argv) {
                 const int win_start_v = (verify_fa_window > 0 && committed > verify_fa_window)
                                             ? (committed - verify_fa_window) : 0;
                 const int win_len_v = committed + q_len - win_start_v;
-                build_causal_mask(mask_buf, win_len_v, q_len, committed, g_kq_stride_pad, win_start_v);
+                // Stride the mask to the SAME kv_pad the mask tensor was
+                // allocated with (align_up(max_ctx + n_tokens)). Without this
+                // override build_causal_mask strides rows by align_up(win_len),
+                // so only query row 0 lands at the right offset and rows 1.. read
+                // an unwritten region, giving NaN/zero attention (and logits)
+                // for every verify position > 0. That is also why the GPU
+                // argmax (sg.argmax_tokens) used to return -1 past position 0.
+                // DDTree already passes this override; chain did not.
+                build_causal_mask(mask_buf, win_len_v, q_len, committed, g_kq_stride_pad, win_start_v,
+                                  align_up(cache.max_ctx + q_len, g_kq_stride_pad));
             }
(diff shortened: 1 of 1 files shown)


server/test/test_dflash.cpp
```diff
@@ -3739,8 +3746,16 @@ int main(int argc, char ** argv) {
...
             tt_verify_compute += std::chrono::duration<double, std::micro>(T_verify_compute - T_verify_set).count();
 
-            ggml_backend_tensor_get(sg.argmax_tokens, target_tok.data(), 0,
-                                    sizeof(int32_t) * q_len);
+            // The GPU argmax shortcut (sg.argmax_tokens) returns -1 after
+            // position 0 for the multi-token batched verify even though the
+            // logits are valid (the same defect fixed for the DDTree path in
+            // 1b3882d). Read full logits and argmax on CPU per position so the
+            // accept loop and bonus see real target predictions, not -1.
+            ggml_backend_tensor_get(sg.logits, verify_logits_buf.data(), 0,
```
howard0su: this increases the size of memory to copy significantly. Do we have a way to optimize this? I remember some early tests shows this size is critical for performance.

PR #347, opened 2026-06-06
ci: run GPU tests on self-hosted lucebox3 (3090 + Radeon 8060S)

davide221:
## What

Adds two **self-hosted** CI jobs that run on `lucebox3`, so the GPU code paths are actually *executed* on hardware instead of only cross-compiled (GitHub-hosted `ubuntu-latest` runners have no GPU).

- **`gpu-tests`** — `runs-on: [self-hosted, gpu, sm86]`. Builds and runs `test_flash_attn_sparse` on the **RTX 3090** (sm_86) using the runner's local CUDA toolkit.
- **`gpu-tests-amd`** — `runs-on: [self-hosted, rocm, gfx1151]`. Compiles and runs `.github/ci/hip_smoke.cpp` (a self-contained HIP vector-add) on the **Strix Halo Radeon 8060S** (gfx1151) via `/opt/rocm/bin/hipcc`.

Both jobs `needs: [uv-workspace, build]` (they only run after the existing hosted jobs pass) and share one `concurrency` group so the single physical box runs at most one GPU job at a time.

## Why

The existing `build` job compiles the dflash CUDA kernels for `sm_86`, but hosted runners can't execute them. lucebox3 has a real 3090 and a Strix Halo iGPU, so these jobs give actual GPU execution coverage for both backends.

## Runner / infra notes

- Runner labels: `self-hosted, Linux, X64, gpu, cuda, sm86, rtx3090, strix-halo, gfx1151, rocm, hip`.
- AMD side is pinned on the box to **ROCm 6.4.4 + kernel 6.14** (ROCm 7.2 page-faults on gfx1151, a known upstream bug). `hipcc` is not on the runner's minimal PATH, hence the absolute path.
- This repo is public, so its Actions fork-PR policy was set to **require approval for all external contributors** before a fork PR can run on the self-hosted box.

## Test

Both smokes verified green directly on lucebox3: `test_flash_attn_sparse` passes on the 3090, and the HIP vector-add prints `PASS` on gfx1151.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,19 +154,26 @@ jobs:
         # safe and fast in CI. This is the execution the hosted jobs cannot do.
         run: ./server/build/test_flash_attn_sparse
 
-      # Optional model-backed end-to-end smoke, disabled by default: it needs
-      # the 16 GB Qwen3.6-27B GGUF + draft staged on the runner. To enable,
-      # stage the weights on lucebox3, set repo variable LUCEBOX_MODELS_DIR, and
-      # uncomment. Mark continue-on-error so a heavy/flaky run never blocks PRs.
+      # Optional model-backed end-to-end smoke (real spec-decode on the 3090),
+      # disabled by default because it builds dflash_server and lazy-loads the
+      # ~16 GB Qwen3.6-27B target + draft (~1-2 min). The weights are already
+      # staged at /opt/models on lucebox3 (override with repo var
+      # LUCEBOX_MODELS_DIR). Verified working on the runner. To enable, uncomment;
+      # continue-on-error keeps a heavy/slow run from ever blocking a PR.
       # - name: dflash end-to-end smoke (model-backed)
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -105,3 +105,99 @@ jobs:
...
+            -DCMAKE_BUILD_TYPE=Release
+          cmake --build build --target test_flash_attn_sparse -j"$(nproc)"
+
+      - name: Run flash-attn sparse kernel test on the 3090
+        # Self-contained kernel correctness test (no model weights), so it is
+        # safe and fast in CI. This is the execution the hosted jobs cannot do.
+        run: ./server/build/test_flash_attn_sparse
+
+      # Optional model-backed end-to-end smoke, disabled by default: it needs
+      # the 16 GB Qwen3.6-27B GGUF + draft staged on the runner. To enable,
```
howard0su: /opt/models already has the model and draft models.

PR #357, opened 2026-06-09
fix(pflash): adaptive anchor_radius eliminates 64K NIAH cliff

dusterbloom:
Re-carved from #274 (commit `1d0baa2`), with a DRY refactor + the unit test the original lacked.

At >=32K the needle text straddles multiple 32-token chunks and the fixed `anchor_radius=2` window (~160 tokens) drops the back half of the needle → truncated/hallucinated retrieval. Scales the window with `n_chunks`: <32K → (radius 2, hits 8); 32-64K → (4, 16); >=64K → (8, 32). Overridable via `PFLASH_COMPRESS_ANCHOR_RADIUS` / `PFLASH_COMPRESS_MAX_ANCHOR_HITS` (legacy `DFLASH_COMPRESS_*` still accepted).

Refactor: both call sites now use a pure `resolve_anchor_params()` helper (`server/src/qwen3/anchor_params.h`) — removes the duplicated block and makes the tier/override logic unit-tested (28 checks: tier boundaries, override precedence, -1 sentinels). Sub-32K behavior is byte-identical to the previous hardcoded defaults. Source commit validated at 49K (needle correctly retrieved; truncated before the fix).

4 files, +122/-4.


davide221:
I cant reproduce the cliff, can you give some more instruction about it?

PR #370, opened 2026-06-11
server: fall back to fresh prefill when a cached snapshot is longer than the prompt

Rhonstin:
## Problem

The prefix-cache restore path treats `prompt_len < snap_pos` as impossible and fails the request:

```cpp
} else if (prompt_len > 0 && prompt_len < snap_pos) {
    // Cached more than the request — should never happen in practice.
    result.error = "snapshot_longer_than_prompt";
    out_io.emit(-1);
    return result;
}
```

With agent clients (Letta, Hermes, OpenClaw, ...) this happens routinely: they summarize/edit their message history between turns, so a follow-up prompt is often *shorter* than the KV snapshot cached on the slot. In production this surfaces as agents silently hanging mid-conversation — the server returns an empty completion:

```
chat DONE ... ok=false in=42491 effective_in=42491 out=0 restore=true slot=14 prefix_len=38663 error=snapshot_longer_than_prompt
chat DONE ... ok=false in=37831 effective_in=37831 out=0 restore=true slot=25 prefix_len=37530 error=snapshot_longer_than_prompt
```

Note `prefix_len` (the matched boundary) is *smaller* than the prompt in these logs — the slot's actual snapshot `cur_pos` was larger, i.e. the entry table and the slot contents had diverged after history edits.

## Fix (two layers)

1. **Routing** (`http_server.cpp`): when the selected slot's snapshot covers more KV than the incoming prompt, treat the lookup as a cache miss instead of restoring a state that cannot be diff-prefilled. Applies to the inline, full-compress and disk hit paths.
2. **Backends** (`qwen35`, `gemma4`, `layer_split`): if such a request still reaches `restore_and_generate`, fall back to a fresh full prefill (mirroring the `generate()` path) instead of emitting `-1`.

## Validation

On an RTX 3090 (Qwen3.6-27B Q4_K_M + DFlash + PFlash): cold request, cache-hit continuation (diff prefill 1.0s vs 8.0s cold), and a shortened-history follow-up all return tokens; greedy outputs on normal cache hits are unchanged. Running in production since 2026-06-11 with the failures gone.

@Luce-Org/maintainers

### DIFF
--- a/server/src/common/layer_split_backend.cpp
+++ b/server/src/common/layer_split_backend.cpp
@@ -172,9 +172,12 @@ GenerateResult LayerSplitBackend::restore_and_generate_impl(
     }
     const int snap_pos = adapter_->snapshot_cur_pos(slot);
     if ((int)req.prompt.size() < snap_pos) {
-        result.error = "snapshot_longer_than_prompt";
-        io.emit(-1);
-        return result;
+        // Snapshot covers more KV than the new prompt (client edited or
+        // summarized its history). Fall back to a fresh full prefill.
+        std::fprintf(stderr,
+            "[pc] snapshot longer than prompt (snap=%d > prompt=%zu) — "
+            "fresh prefill fallback\n", snap_pos, req.prompt.size());
+        return run_from_state(req, io, /*base_pos=*/0, /*reset_state=*/true);
     }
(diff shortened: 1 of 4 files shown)


davide221:
@Rhonstin very good finding, thanks for the contribution

PR #372, opened 2026-06-11
compose: FlowKV aged-history compression + drafter residency fix — 1.72x vs disk-cache baseline at <=64K

dusterbloom:
## TL;DR

Baseline is **current main**, which already ships the [#364](https://github.com/Luce-Org/lucebox-hub/pull/364) scoped disk prefix cache (merged). This PR adds FlowKV aged-history compression on top of that baseline (RTX 3090 24GB):

| | main baseline (#364 disk cache) | this PR (compose) | delta |
|---|---:|---:|---:|
| 7-turn agentic session wall (N=3 mean) | 527.5s | **306.7s** | **1.72x** |
| worst-turn fresh prefill (26K tok) | 370 tok/s (73-77s) | 396 tok/s (66-73s) | parity+ |
| decode @63K context | 8.1-9.1 tok/s | 10.6-14.5 tok/s | +16-30% |
| tool-call validity | 16/21 | 18/21 | held |

Benchmark: `goldgate_fix` trace (real multi-turn agentic session, 34K-64K prompt tokens per turn), N=3 interleaved A/B on the same binary, same thermal window.

## Re-validation + q8_0 KV (2026-06-12)

The compose win was re-validated on the final PR head (incl. the deps bump below), same trace, same binary/lib/window, on both the current default KV type (tq3_0) and q8_0:

| arm | wall (N=2 mean) | worst-turn fresh prefill | decode @63K | tool-calls |
|---|---:|---:|---:|---:|
| main baseline, tq3_0 | 437.1s | 403 tok/s | 13.7-14.3 tok/s | 12/14 |
| **compose (this PR), tq3_0** | **257.6s** | 400 tok/s | 14.8-15.5 tok/s | 12/14 |
| main baseline, q8_0 | 270.3s | 664 tok/s | 22.1-23.2 tok/s | 10/14 |
| **compose (this PR), q8_0** | **192.2s** | 673 tok/s | 21.9-23.6 tok/s | 13/14 |

- compose wins **1.70x on tq3_0** and **1.41x on q8_0** (same-window N=2 means).
- q8_0 KV roughly doubles decode at 63K vs tq3_0 (llama-bench, this fork: tq3_0 14.43 vs q8_0 28.3-29.4 tok/s tg32 @ d63488). Compose + q8_0 vs the main baseline at today's default KV (tq3_0) — same binary, lib, and window — is **2.27x wall** end-to-end.
- Serving q8_0 at max-ctx 65536 on 24GB requires: (a) the llama.cpp deps bump in this PR ([llama.cpp-dflash-ggml#16](https://github.com/Luce-Org/llama.cpp-dflash-ggml/pull/16) — see Limitations), and (b) building with `-DDFLASH27B_FA_ALL_QUANTS=OFF` (see Limitations). Flags: `--cache-type-k q8_0 --cache-type-v q8_0`.
- The drafter-residency release in this PR is what buys the ~2GB that lets q8_0 KV fit at 65K on 24GB.

## Summary

Main's [#364](https://github.com/Luce-Org/lucebox-hub/pull/364) disk prefix cache (merged) made warm agentic turns cheap by restoring a stable token prefix from disk. The remaining cost on long sessions is the aged conversation history that still has to be prefilled fresh whenever the prefix diverges, and the per-turn growth beyond the cached boundary. This PR composes FlowKV aged-history compression with that cache: messages older than a hot window are compressed (drafter-scored, anchor-preserving) while the system prompt stays verbatim as the cache anchor, so the disk-cache key remains stable across turns. A unified gate keeps the three paths exclusive — turn-1 verbatim, FlowKV on continuations, whole-prompt pFlash otherwise — and with compression disabled the request path is byte-identical to main.

Two fixes found during benchmarking turned the compose from a wash into the 1.72x above:

- **Drafter residency.** The pflash scoring drafter (~2GB BF16) staying resident through the target's large prefill collapses prefill throughput 370 -> 121 tok/s on 24GB cards (allocator pressure at ~61K KV; verified by A/B, not capacity — the same prefill runs at full rate with the drafter released). The drafter is now released after its scoring pass and lazily reloaded (~2s). This is the `auto` residency default; `--draft-residency persistent` keeps the old behavior for >=32GB cards.
- **Admission ordering.** The ingress context-length check rejected prompt+max_tokens > max_ctx before compression could run. Oversized requests are now admitted when compression will run, and the hard limit is enforced on the post-compress effective size instead.

## Changes

- FlowKV aged-history compression composed with the #364 scoped disk cache behind a unified gate (`http_server.cpp`); `compress` off keeps main's behavior byte-identical.
- `auto` draft residency releases the pflash drafter after compress scoring (`placement/draft_residency.h`).
- Pure admission helper `should_reject_oversized()` + post-compress effective-size gate (`server/admission.h`).
- Skip-park guard: `--prefill-skip-park` downgraded on <32GB GPUs at max_ctx>65536 (VMM VA-fragmentation crash class) (`placement/skip_park_guard.h`).
- ee7 early-exit drafter, anchor-transitive cascade with expansion throttles, tail-capture guard for the chunk-boundary assert.
- deps: llama.cpp bumped to luce-dflash `574be613` — picks up [llama.cpp-dflash-ggml#16](https://github.com/Luce-Org/llama.cpp-dflash-ggml/pull/16) (cuMemSetAccess retry during VMM pool growth).
- Tests: 1926-assertion unit suite green; standalone suites for admission (7), skip-park guard (6), anchor cascade, early-exit score range, warm-path regression. ~55% of the diff is tests.

## Limitations

- **VMM pool-growth crash class — corrected root cause.** The long-standing `cuMemSetAccess device not ready` abort (>64K cold prompts, and reproducible at 65K with q8_0/f16 KV + decode draft resident, ~20GB load, no pFlash involved) is a transient NOT_READY during VMM pool growth that was masking the real condition. [llama.cpp-dflash-ggml#16](https://github.com/Luce-Org/llama.cpp-dflash-ggml/pull/16) (merged; included via this PR's deps bump) retries through the transient: the crash cell now either completes (verified 7/7 on the goldgate replay where the old pointer died turn-1) or reports an honest out-of-memory.
- **`FA_ALL_QUANTS` VRAM spike at high load.** `server/CMakeLists.txt` defaults `DFLASH27B_FA_ALL_QUANTS=ON`, which compiles ~3x more fattn kernel instances; with CUDA lazy module loading their device code faults in at the first FlashAttention call mid-prefill. At >=~20GB load (q8_0 KV at 65K + decode draft) this spike tips pool growth into OOM — reproduced deterministically with the GPU otherwise idle, and eliminated by building with `-DDFLASH27B_FA_ALL_QUANTS=OFF` (same-type KV pairs incl. tq3_0/q8_0/q4_0 keep their fast vec kernels; only mixed K/V type pairs need ON). Changing the default is left to a follow-up.
- Compression keeps ~93% on dense code (anchor-dominated) — known, separate lever.
- `GGML_CUDA_NO_VMM=1` as an environment variable is a no-op (compile-time option in this fork); scripts relying on it were never protected.

## History

1. `731561d1` compose FlowKV with #364 scoped cache; `0efdc33c` gate compression as fallback so compose can't regress main; `6a848058` unified gate (FlowKV reachable + scoped save preserved).
2. `cefa3caf` ee7 early-exit drafter + anchor-transitive cascade + tail-capture guard.
3. `3fc6882f` drafter auto-release after compress scoring (the 1.72x).
4. `2ae98c0f` compress-aware admission.
5. `637fbdaf` comment trim (-133 LOC, no logic changes).
6. `1c562eb4` skip-park footprint guard.
7. `e542e908` review fixes (request-scope override no longer drops server compress; full-cache-hit admission); `26e0ee33` -1 sentinel for full-cache served tokens.
8. `6a584981` deps bump to luce-dflash `574be613` (VMM pool-growth fix).

### DIFF
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -218,6 +218,7 @@ add_library(dflash_common STATIC
     src/draft/draft_gguf_loader.cpp
     src/draft/draft_safetensors_loader.cpp
     src/draft/draft_graph.cpp
+    src/qwen3/anchor_scan.cpp
     src/qwen3/qwen3_drafter.cpp
     src/qwen3/qwen3_loader.cpp
     src/qwen3/qwen3_graph.cpp
@@ -602,6 +603,31 @@ if(DFLASH27B_TESTS)
         target_link_libraries(test_anchor_params PRIVATE dflash_common)
         add_test(NAME anchor_params COMMAND test_anchor_params)
     endif()
+    if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test/test_drafter_early_exit_score_range.cpp")
+        add_executable(test_drafter_early_exit_score_range
+            test/test_drafter_early_exit_score_range.cpp)
+        target_include_directories(test_drafter_early_exit_score_range PRIVATE
+            ${CMAKE_CURRENT_SOURCE_DIR}/src/common)
+        add_test(NAME test_drafter_early_exit_score_range
+            COMMAND test_drafter_early_exit_score_range)
+    endif()
+    if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test/test_anchor_transitive.cpp")
+        add_executable(test_anchor_transitive
+            test/test_anchor_transitive.cpp
+            src/qwen3/anchor_scan.cpp)
+        target_include_directories(test_anchor_transitive PRIVATE
+            ${CMAKE_CURRENT_SOURCE_DIR}/src/qwen3)
+        add_test(NAME test_anchor_transitive
+            COMMAND test_anchor_transitive)
+    endif()
+    if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test/test_drafter_warm_path_regression.cpp")
+        add_executable(test_drafter_warm_path_regression
+            test/test_drafter_warm_path_regression.cpp)
+        target_include_directories(test_drafter_warm_path_regression PRIVATE
+            ${CMAKE_CURRENT_SOURCE_DIR}/src/common)
+        add_test(NAME test_drafter_warm_path_regression
+            COMMAND test_drafter_warm_path_regression)
+    endif()
     if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test/test_drafter_tail_capture_guard.cpp")
         # GREEN phase: built with TAIL_GUARD_USE_NEW_FORMULA — must pass after Bug #42 fix.
         add_executable(test_drafter_tail_capture_guard
@@ -614,8 +640,6 @@ if(DFLASH27B_TESTS)
         add_executable(test_drafter_tail_capture_guard_red
             test/test_drafter_tail_capture_guard.cpp)
         # No TAIL_GUARD_USE_NEW_FORMULA — uses old (buggy) guard, expected to FAIL.
-        add_test(NAME test_drafter_tail_capture_guard_red COMMAND test_drafter_tail_capture_guard_red)
-        set_tests_properties(test_drafter_tail_capture_guard_red PROPERTIES WILL_FAIL TRUE)
     endif()
     if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test/test_draft_vs_reference.cpp")
         add_executable(test_draft_vs_reference test/test_draft_vs_reference.cpp)
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -292,6 +292,7 @@ add_library(dflash_common STATIC
     src/server/sse_emitter.cpp
     src/server/prefix_cache.cpp
     src/server/disk_prefix_cache.cpp
+    src/server/freeze_history.cpp
     # ── Jinja chat-template engine (from llama.cpp common/jinja/) ──
     # Used by render_chat_template_jinja() to support --chat-template-file
     # in dflash_server. Mirrors llama.cpp's common_chat_template plumbing.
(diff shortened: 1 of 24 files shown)


dusterbloom:
Review disposition (all 8 cubic findings verified against code + tests before fixing):

| finding | verdict | action |
|---|---|---|
| http_server.cpp:2175 raw size on full-cache hits | confirmed | **fixed in e542e908** (`effective_prompt_overflows` helper, failing-test-first) |
| disk_prefix_cache.h:50 scope override drops compress | confirmed | **fixed in e542e908** (`apply_request_scope_override`, failing-test-first) |
| http_server.cpp:1904 JSON throw | refuted | `json::value(key, default)` is type-safe (returns default on type mismatch, no throw); the `.get<std::string>()` at :1897 is guarded by `is_string()` |
| anchor_scan.cpp:27 / :103 / anchor_scan.h:18 | confirmed in the utility library | not reachable from production (the shipping drafter uses its own inline scan; these functions are exercised by tests only). Fixes queued as a follow-up batch rather than blocking this PR |
| qwen35_backend.cpp:541 current-device VRAM query | partial | latent multi-GPU-only issue; single-GPU is the only shipped config. 1-line fix queued with the follow-up batch |
| freeze_history.h:10 unused include | confirmed | queued with follow-up batch |

Suite after fixes: 1939 assertions green; admission standalone 12/12.

dusterbloom:
cubic P2 (http_server.cpp:1798 sentinel conflation): a zero-length full-cache entry is not constructible today (admission floor + kept anchors guarantee non-empty compressed prompts), but the conflation class is now removed outright — sentinel is `-1`, gate treats `>=0` as a hit. Failing-test-first (`test_effective_overflows_zero_length_hit_is_a_hit`), suite 13/13 + 1939 asserts green. Fixed in 26e0ee33.

PR #374, opened 2026-06-12
qwen35: sampled-verify — speculative decoding with an active sampler

Rhonstin:
## What

Adds a **sampled-verify** mode to the qwen35 chain spec-decode path, enabled with `DFLASH_SAMPLED_VERIFY=1`. It implements the "active sampler during verification" item from the README's future additions: speculative decoding that works at **any temperature** while preserving the exact target sampling distribution.

Today the chain verify commits the target's argmax at each position, so spec decode is only distribution-correct at temperature 0 — sampling requests fall back to AR decode. On a 27B Q4_K_M / RTX 3090 that means 16 tok/s for any agent traffic that samples.

## How it works

Sample-and-match: at each chain position the verifier draws a token from the target's **full sampler chain** (temperature / top-k / top-p / penalties) over that position's verify logits, and accepts the draft token iff the sample matches it. On mismatch, the sampled token itself becomes the bonus token. Every committed token is therefore an exact sample from the target distribution — output distribution identical to AR sampling, no acceptance-rate correction terms needed.

Three spots must sample rather than take argmax — each one, when wrong, corrupts generation in a distinctive way that we hit and fixed during validation:

1. **First token after prefill** — was the prefill argmax; injects one greedy token per request (at temp 1, 150/150 generations opened with the same word). Now sampled from the prefill logits.
2. **Per-step seed** — the next draft step is seeded with the replay's last token, which the next verify commits as-is. Argmax here injects one greedy token per step (~1/16 of output) and locks long generations into repetition loops. Now sampled from the replay's last verify logits.
3. **The acceptance walk** — over per-position verify logits, exposed via a new `DFlashTarget::read_verify_logits()` hook (virtual, default `false`; the qwen35 target reads the step graph's logits tensor, so targets that don't implement it are unaffected).

## fa-window interaction

Sampled-verify is gated to `fa_window == 0`. With a finite `--fa-window`, the windowed **verify** pass starves the logit tail at long context: argmax survives, but sampling from the poisoned tail degrades quality. Decisive experiment on a 24K-token agent prompt: tool-call success **0/12** with a finite window vs **12/12** with full-attention verify. Prefill and AR decode always run full attention, so they're unaffected — this only constrains the verify pass.

## Results (RTX 3090, Qwen3.6-27B Q4_K_M + DFlash draft)

| Workload | AR sampling | sampled-verify |
|---|---|---|
| short prompts, temp 0.7–1.0 | 16 tok/s | 60–70 tok/s |
| 24K agent prompt w/ tools, temp 1.0 | 16 tok/s | ~31 tok/s |
| tool-call success (12 runs, temp 1.0) | 12/12 | 12/12 |

- Greedy path (`DFLASH_SAMPLED_VERIFY` unset) is bit-exact unchanged.
- Token histograms over 150 sampled generations match AR within noise.
- No degeneration over 1200-token continuous generations.
- `DFLASH_SV_DEBUG=1` traces the acceptance walk per position.

Default is **off**, so existing deployments see zero behavior change.

<a href="https://cubic.dev/pr/Luce-Org/lucebox-hub/pull/374?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/src/common/dflash_target.h
+++ b/server/src/common/dflash_target.h
@@ -33,6 +33,15 @@ struct DFlashTarget {
                               int & last_tok,
                               std::vector<int32_t> * all_argmax = nullptr) = 0;
 
+    // Read the full [n_tokens x vocab] f32 logits produced by the most
+    // recent verify_batch call. Used by sampled-verify (spec decode with
+    // temperature). Returns false when the implementation does not keep
+    // verify logits around.
+    virtual bool read_verify_logits(int n_tokens, std::vector<float> & out) {
+        (void)n_tokens; (void)out;
+        return false;
+    }
+
     // ── KV state management ─────────────────────────────────────────
 
     // Snapshot KV cache state before speculative verify, so it can be
--- a/server/src/qwen35/qwen35_backend.cpp
+++ b/server/src/qwen35/qwen35_backend.cpp
@@ -1348,16 +1348,38 @@ bool Qwen35Backend::do_spec_decode(int committed, int n_gen,
     // out-of-bounds tensor read.  cache_.last_tok is always correct.
     int32_t last_tok = cache_.last_tok;
 
+    // Sampled-verify: spec decode with an active sampler. Each chain
+    // position is verified against a token drawn from the target's own
+    // sampler chain instead of its argmax, so every committed token is an
+    // exact target sample — the output distribution is identical to AR
+    // sampling. Acceptance drops vs greedy but stays far above the AR
+    // floor. Opt in with DFLASH_SAMPLED_VERIFY=1; without it, sampling
+    // requests fall back to AR decode (zero behavior change by default).
+    static const bool kSampledVerify = []() {
+        const char * e = std::getenv("DFLASH_SAMPLED_VERIFY");
+        return e != nullptr && std::string(e) == "1";
+    }();
+    // Sampled-verify additionally requires full attention in the verify
+    // path. With a finite --fa-window the verify batch applies one
+    // window-start to the whole batch (unlike the AR step graph, which is
+    // hardcoded to full attention): the argmax stays robust, so greedy
+    // verification is unaffected, but the logit TAIL drifts at long
+    // context and top-k sampling draws degenerate tokens from it
+    // (reproduced at 24K: 0/12 tool calls with fa-window 2048, 4/4 with 0).
+    const bool sampled_verify = kSampledVerify &&
+        sampler_.needs_logit_processing() &&
+        cfg_.fa_window == 0;
+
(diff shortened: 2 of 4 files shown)


Rhonstin:
Amended after self-review: (1) the DFLASH_SAMPLED_VERIFY gate is now correctly opt-in (=1) as documented — the earlier revision had it opt-out; (2) the acceptance walk now seeds its penalty history with the step's seed token (draft_tok[0]), which is committed by the replay but not yet in out_tokens at walk time — without it the sampled distribution drifted from AR by one token whenever repetition penalties are active.

davide221:
@Rhonstin thanks for the contribution! Can you fix the conflict? I would like to know your precise testing setup to reproduce the performance improvements

Rhonstin:
Hi @davide221, thanks for the review! Conflict is resolved now.

## Testing setup to reproduce

**Hardware:**
- Single RTX 3090 24GB VRAM
- Intel i9-13900H host

**Models:**
- Target: `Qwen3.6-27B-Q4_K_M.gguf` (full 27B, Q4_K_M quantization)
- Draft: `Qwen3-0.6B-BF16.gguf` (0.6B PFlash drafter, BF16)

**Server config:**
- `dflash_server` (the custom ggml/CUDA fork in this repo)
- `--cache-type-k q4_0 --cache-type-v q4_0`
- `--max-ctx 114688`
- `--prefill-compression auto --prefill-threshold 16000` (PFlash speculative prefill compression)
- `--draft-residency request-scoped`

**Environment variables for sampled-verify:**
```
DFLASH_SAMPLED_VERIFY=1
DFLASH_TREE_VERIFY=1
DFLASH27B_CHUNKED=1
```

**Benchmark methodology:**
- `harness/benchmarks/generation_benchmark.py` with the Luce harness client
- Needle-in-haystack retrieval tests at various context depths (57K–136K tokens)
- Measured: speculative accept rate, retrieval accuracy (pass/fail on exact value extraction), generation throughput (tok/s)
- Compared baseline (greedy-only verify) vs sampled-verify mode across identical prompts

**Key observation:** Sampled-verify gives exact distribution matching at any temperature — the draft is accepted or rejected based on the target's actual sampler output, not just argmax. This matters for agent workloads where temperature > 0 is standard.

Happy to provide more detail or run additional benchmarks if needed.

PR #429, opened 2026-06-20
feat(kvflash): pager serialize/deserialize + critical-chunk pinning

dusterbloom:
## What
Snapshot/restore for the KVFlash pager: `serialize()` / `deserialize()` with a
fixed-layout header guard (rejects layout drift), plus critical-chunk pinning so
sink/anchor chunks survive eviction (with a deadlock floor that refuses pins that
would starve the eviction pool).

## Shape
- `server/src/common/kvflash_pager.h` — header-only addition; backend-agnostic.
- ~310 LOC. Test: `test_kvflash_pager.cpp` (serde round-trip + header-mismatch
  guard + pin deadlock guard).

Independent of the placement PR; required by the prefill-snapshot PR.

<a href="https://cubic.dev/pr/Luce-Org/lucebox-hub/pull/429?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #430, opened 2026-06-20
feat(qwen35moe): pooled chunked prefill + snapshot/restore over KVFlash

dusterbloom:
## What
Pooled chunked prefill for qwen35moe (Qwen3.6-35B-A3B) over KVFlash: when the
prompt exceeds the resident pool, prefill loops `hybrid_forward_batch` over
chunk-sized slices with live eviction instead of refusing. Plus pooled
snapshot/restore (save/restore the bounded pool across requests) and a
complexity-only refactor (dedup the two identical restore chunk loops, extract
`chunked_prefill`, inline a single-caller helper — net −25 LOC, behaviour-identical).

## Stacking
This is the tip of the KVFlash-MoE stack and depends on:
- #428 — pool-sized KV reservation (keeps experts hot / placement gate)
- #429 — pager serialize/deserialize (snapshot/restore serde)

Until those merge this PR's diff includes their commits; rebasing after they land
leaves only the prefill-snapshot + refactor commits.

## Tests
`test_kvflash_moe_paged.sh` — GPU silent-corruption gate: a sink fact in the
first (protected) chunk is recalled after the middle is evicted, and the greedy
(temp-0) answer is identical across two pool sizes. Green on RTX 3090 / Q3_K_M.

<a href="https://cubic.dev/pr/Luce-Org/lucebox-hub/pull/430?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dusterbloom:
Superseded by #466 (kvflash library — green) + a follow-up qwen35 integration PR. The pooled-chunked-prefill + snapshot/restore integration and all review fixes (P1 restore-cache check, P2 seed_n/deserialize/validation, P3 dead-code) are carried into the integration PR from the same clean-perf source.

Issue #44, opened 2026-04-27
Q5/Q6 support on RTX 3090?


fangyaodong:
Hi, I’m trying to understand the future direction a bit 🙂

Is Q5 or Q6 quantization on something like an RTX 3090 (24GB) something that is technically expected to work, or is it still uncertain / just an idea for now?

I’m currently deciding whether to upgrade my GPU (3090 vs higher VRAM options), and my main goal is to run models at Q5 or Q6 quality.

I’ve heard that Q6 is already very close to lossless in practice, so if something like that becomes possible on a 3090, it would definitely influence my decision.

So I just want to get a rough sense:

is this something likely to be achievable on 3090?
or should I not rely on that yet?

No need for exact timelines — just trying to understand the direction.

Also, really impressed by what you’re building here — it feels like this kind of optimization could significantly increase the value of cards like the 3090.

Thanks!

davide221:
Thanks for the support @fangyaodong. Our goal is to build as much as possible optimizations for consumer cards and we are starting heavily on the 3090. It is possible to run q5/q6 at good throughput on an RTX 3090, but it is not our main priority at the moment, qwen3.6 27b q4 has great performance and we want to optimize his throughput, for bigger quant we would need sparsity or activation prediction while we are currently focusing on speculative prefill/decode. But any contribution in that direction is really appreciated

fangyaodong:
Thanks for the clarification, really helpful!

tomByrer:
> lossless

@fangyaodong Seems you're new-ish to LLMs?  When they say a quant is 'lossy', it means it loses its conformity to the 'base'/parent model.  It might still be better than other models.  It might not lose anything for your use-case (eg  maybe most impact is on vision, or storytelling, but code is mostly untouched).  You can also 'tune' the model to better fit your needs, but I'm not sure if Luce supports those tuned models?

I have seen those charts from Unsloth; I'd say Q5 tends to be the sweet spot, but a Q4 might allow larger in-VRAM context, so YMMV.

davide221:
Yes lucebox dflash support custom models if they are fine-tuned from the model family that are supported by dflash (qwen, kimi, gpt-oss

PR #431, opened 2026-06-20
Add speed profiler

officialgr:
## Summary
Adds a report-only speed profiling workflow for Lucebox inference changes.

## Changes
- Adds `server/scripts/profile.py`
- Adds `.github/workflows/speed-profile.yml`
- Adds profiler documentation at `server/docs/speed-profile.md`

## Notes
- Workflow is report-only / non-blocking.
- Intended for the self-hosted RTX 3090 GPU runner.
- Produces `profile.json`, `profile.md`, and nsys artifacts when available.

<a href="https://cubic.dev/pr/Luce-Org/lucebox-hub/pull/431?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/speed-profile.md
+++ b/docs/specs/speed-profile.md
@@ -14,7 +14,9 @@ Three layers, coarse to fine:
 
 1. **Headline latency/throughput** — prefill time, model-side TTFT estimate, decode
    tok/s, ms/token, plus the speculative-decoding **acceptance length (AL)** and
-   accept %. (`AL` is how many tokens the target commits per draft+verify step; decode
+   accept %. Repeated timing passes report **mean ± stddev**, not a single-point
+   estimate, so reviewers can tell whether a small delta is real or just jitter.
+   (`AL` is how many tokens the target commits per draft+verify step; decode
    throughput ≈ `AL / step_time`.)
 2. **Per-step phase breakdown** — `draft_compute`, `draft_logits`, `verify_compute`,
    … from the engine's own timers. Tells you *which phase* dominates a step.
@@ -31,15 +33,26 @@ and they stay fixed so every run is comparable over time:
   per target pass. 22 is the `dflash_server` default. Bigger = a bigger bet (higher
   potential acceptance length) but more draft+verify cost; tuning it is a separate sweep,
   not the CI job.
-- **`--n-gen 48`** — tokens generated per prompt. Long enough to be past warmup, short
-  enough to keep the shared 3090 fast. (Very short n_gen under-counts spec-decode because
-  fixed startup costs aren't amortized.)
-- **`--reps 3`** — repeats the decode and takes the **median** tok/s to absorb run-to-run
-  GPU variance (thermals, clock boosting, scheduler jitter).
+- **`--n-gen 128`** — requested generated tokens per prompt. This is long enough
+  to amortize startup costs and reduce very-short-generation bias, while still
+  keeping the shared 3090 CI queue bounded. The argument is passed as `<n_gen>`
+  to both `test_dflash` and `test_generate`, where it controls the response length
+  generated for each benchmark prompt.
+- **`--reps 5`** — repeats each prompt enough times to expose run-to-run GPU
+  variance (thermals, clock boosting, scheduler jitter), then reports mean and
+  stddev for the headline metrics. Use `--reps 3` for a faster smoke profile when
+  needed, but PR comparisons should prefer 5+.
+- **`--noise-rsd-pct 0.05`** — report-only noise threshold. If any tracked
+  headline metric has relative stddev above 5%, the markdown calls the profile
+  **NOISY** and tells reviewers to treat small deltas as below the profiler
+  detection threshold.
 
 **Rule:** keep these consistent. A delta vs the baseline is only a valid regression
 signal if both runs used the same config — if you ever change a parameter, re-seed the
-baseline (you cannot compare across configs).
+baseline (you cannot compare across configs). When baseline and current 1σ intervals
+overlap and the delta is smaller than `--regress-pct`, the report marks that row as
+**noisy / overlap** instead of inviting reviewers to chase a ghost regression. All of
+these states are warnings only; the profiler remains report-only.
 
 ## Losslessness gate (and why a bit-exact compare is too strict on its own)
 
@@ -74,6 +87,14 @@ one-at-a-time). Classifying a FAIL as bug-vs-FP needs the **logit gap** at the f
 mismatch (near-tie = FP, clear gap = bug) — a follow-up the binaries don't emit yet. CI
 surfaces a FAIL as a non-blocking `::warning::` for triage; it stays report-only.
 
+## CI settings
+
+The `Speed Profile` workflow uses the same profiler defaults as the local recipe:
+`--n-gen 128`, `--reps 5`, and `--noise-rsd-pct 0.05`. Runner owners can
+temporarily override those values with repo variables `LUCEBOX_SPEED_N_GEN`,
+`LUCEBOX_SPEED_REPS`, and `LUCEBOX_SPEED_NOISE_RSD_PCT`, but PR-to-PR comparisons
+should keep them fixed.
+
 ## Run it locally
 
 ```bash
@@ -82,7 +103,7 @@ python3 scripts/profile.py \
   --target /opt/models/Qwen3.6-27B-Q4_K_M.gguf \
   --draft  /opt/models/draft/dflash-draft-3.6-q4_k_m.gguf \
   --tokenizer Qwen/Qwen3.6-27B \
-  --n-gen 48 --budget 22 --reps 3 \
+  --n-gen 128 --budget 22 --reps 5 --noise-rsd-pct 0.05 \
   --nsys --check-lossless \
   --baseline scripts/speed-baseline.json --regress-pct 0.10 \
   --out-json profile.json --out-md profile.md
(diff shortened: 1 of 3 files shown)


davide221:
@officialgr thanks for the profiler work, can you fix the CI error?

officialgr:
@davide221 Fixed — the job was running profile.py with the system python3, which doesn't have transformers (used for tokenization). It now installs it into an isolated venv before running.

officialgr:
The failure was caused by the speed-profile workflow looking for the wrong speculative draft model path. The workflow now defaults to the repo-documented Qwen3.6 draft GGUF path and skips cleanly with diagnostics if runner-local weights are missing.

davide221:
Nice progress, it now runs end-to-end on the 3090 and is already useful. Two things to look at in your own passing run before this is mergeable:

1. The lossless gate reports FAIL (2/3 prompts diverge) but the job is still green. Is that a real correctness bug or too-strict a check?
2. The nsys section prints 0.0 everywhere with an empty table

A couple of things would make this much more valuable: a baseline + threshold so it can actually flag a regression and a thought on whether we should also profile dflash_server (the shipping path), since test_dflash can't see the prefix cache / batching.

officialgr:
@davide221 
Ran it end-to-end on the 3090 at the shipping defaults (Qwen3.6-27B, budget 22, n_gen 48).

**1. Lossless — gate is no longer "too strict."** Added an AR-vs-AR determinism control (reuses the existing AR passes, no extra GPU cost): a prompt only FAILs when AR is self-deterministic *and* spec-decode still differs; intrinsically-nondeterministic prompts become *inconclusive*. On my run it flags a divergence at token #44 (2/3 prompts). I checked whether it's a tail artifact by re-running at n_gen 96 — it stayed at #44, so it's reproducible/structural, not a length effect. I'm deliberately not over-calling it: it's ruled out as run-to-run noise, but it could still be batched-verify FP (verify scores draft tokens as a batch vs AR one-at-a-time). The definitive bug-vs-FP test is the logit gap at #44 (near-tie = FP, clear gap = bug) — the binaries don't emit per-token logits yet, so I'm flagging it for the engine team rather than claiming a bug.

**2. nsys — fixed, now fully populated.** Root cause was report-name drift across nsys versions (`cuda_gpu_kern_sum` ↔ legacy `gpukernsum`) plus a swallowed error. It now tries both name sets and, if stats still won't parse, prints the nsys error + version instead of silent zeros. On Qwen3.6: 568.8 ms GPU kernel time, 674.6 launches/token, top kernel `mul_mat_q` (187 ms / 2880 launches), 75.7 ms/token in host↔device copies.

**3. Baseline + threshold — added.** `--baseline` + `--regress-pct` (default 10%, wide enough for run-to-run variance), direction-aware (tok/s & AL up = good; ms/token & TTFT down = good). The report flags `⚠️ REGRESSION` and CI emits a non-blocking warning. Verified the delta path on a self-vs-self run (`✅ within threshold`). Seeded `server/scripts/speed-baseline.json` from the green run.

**4. dflash_server — recommendation (no code, since you asked for "a thought").** Keep the binaries as the per-PR signal (deterministic, low-noise, isolates compute). The shipping path's prefix-cache + batching is a different question (throughput under load), so I'd add it as a separate, higher-variance Phase 2 — replay a request trace at fixed concurrency; TTFT/p50/p99 + cache-hit rate — on a schedule/label, not every PR. Written up in `docs/specs/speed-profile.md`.

**Final note on `n_gen`.** `n_gen=48` is good for CI speed, but it is too short to give the most representative throughput signal — in my runs it stays around ~0.98 and is more sensitive to setup/early-step overhead. The `128–256` range works much better and gives more stable speed numbers. I’d recommend moving the default to `128` for a balanced PR profile, or `256` if we want the profile to favor benchmark quality over runtime.

dusterbloom:
@officialgr  Great work getting the profiler reliable and report-only. 

1 thing I’d love to see before we add more features: let’s make the profiler honest about its own noise. Right now a single run produces one latency number, which makes it hard to tell whether a 3% delta in a PR is real or just run-to-run jitter.

Here a couple of things that could help avoid chasing ghost / thermal variance that could be added:

   • More than 3 repeated runs per prompt (e.g., 3–5, configurable) and report the mean + stddev.
   • A clear “noisy / below detection threshold” state when the relative stddev is above some sensible threshold, or when the before/after overlap is too large to call a regression.
   • Keep it report-only for now - warnings, not blockers.
   
Keep up the good work. Let me know what you think.

officialgr:
@dusterbloom 
Thanks — agreed with this direction, and I think the profiler now covers the noise concerns while staying report-only.

What changed:

- The profiler now uses a longer generation sample by default: `--n-gen 128`. This controls the number of generated response tokens per benchmark prompt. Based on testing, shorter generations below 128 were too sensitive to startup/thermal jitter, while 128 is a good compromise between CI runtime and measurement quality.
- The profiler now runs repeated timing passes per prompt by default (`--reps 5`), with `--reps` still configurable for smoke runs or future tuning.
- Instead of reporting one latency/throughput number, it records per-run samples and reports mean ± stddev for the headline metrics, including decode tok/s, ms/token, TTFT estimate, and acceptance length.
- The markdown report now has a dedicated “Noise / detection threshold” section. If the relative stddev exceeds the configured threshold (`--noise-rsd-pct`, default 5%), it marks the profile as NOISY and explicitly says to treat small deltas as below the profiler detection threshold.
- Baseline comparisons now include stddev-aware overlap handling. If the current and baseline 1σ ranges overlap and the delta is smaller than the regression threshold, the row is marked as “noisy / overlap” rather than encouraging us to chase a likely ghost regression.
- This is still report-only. The workflow only emits GitHub warning annotations for noisy profiles / regressions / losslessness issues; it does not block the PR.
- CI was updated to use the same defaults: `--n-gen 128`, `--reps 5`, and `--noise-rsd-pct 0.05`, with repo-variable overrides available for temporary smoke runs.

So the profiler should now be more honest about whether a small delta is meaningful or just run-to-run / thermal variance. The `--n-gen 128` change helps reduce startup-noise sensitivity, the repeated runs quantify remaining jitter, and the noisy/overlap labels keep the whole thing report-only instead of turning measurement noise into a blocker.

docs/specs/speed-profile.md
```diff
@@ -0,0 +1,110 @@
+# Lucebox speed profiler (MVP)
+
+A small, CI-runnable profiler for the inference engine. It measures forward-pass
+speed on one GPU and produces a report that shows **where the time goes**, so a
+reviewer can see at a glance whether a PR moved the needle and where the next
+optimization margin is.
+
+It runs the engine **binaries directly** (no HTTP) so the numbers reflect compute,
```
howard0su: which binary?

docs/specs/speed-profile.md
```diff
@@ -0,0 +1,110 @@
...
+
+## CI settings
+
+The `Speed Profile` workflow uses the same profiler defaults as the local recipe:
+`--n-gen 128`, `--reps 5`, and `--noise-rsd-pct 0.05`. Runner owners can
+temporarily override those values with repo variables `LUCEBOX_SPEED_N_GEN`,
+`LUCEBOX_SPEED_REPS`, and `LUCEBOX_SPEED_NOISE_RSD_PCT`, but PR-to-PR comparisons
+should keep them fixed.
+
+## Run it locally
```
howard0su, line 101: how to compare the result with baseline?

server/scripts/profile.py
```diff
@@ -0,0 +1,822 @@
...
+              "def sum_product(numbers: List[int]) -> Tuple[int, int]:\n"
+              '    """Return a tuple of (sum, product) of all integers. Empty -> (0, 1).\n'
+              "    >>> sum_product([1, 2, 3, 4])\n    (10, 24)\n"
+              '    """\n    s = 0\n    p = 1\n    for n in')},
+]
+
+# --------------------------------------------------------------------------------------
+# stdout parsers (matched to the exact print formats in test_dflash.cpp / test_generate.cpp)
+# --------------------------------------------------------------------------------------
+RE_PREFILL      = re.compile(r"\[prefill\]\s+(\d+)\s+tokens?\s+in\s+([\d.]+)\s*s")
```
howard0su, line 73: this is error prone. it is fine for the first version. I would suggest to add json output support to test_dflash binary so we don't do the text parse here.

officialgr:
Thanks for the review, @howard0su — good calls, addressed all three:

1. Which binary: there are two — `test_dflash` (the speculative / DFlash decode
path) and `test_generate` (the autoregressive baseline), both built from server/
as CMake targets (`build/test_dflash`, `build/test_generate`, overridable via
`--df-bin`/`--ar-bin` or `DFLASH_BIN`/`DFLASH_BIN_AR`). I've named them explicitly
in the doc.

2. Comparing against a baseline: save a run's JSON (`--out-json
scripts/speed-baseline.json`) and commit it, then on later runs pass `--baseline
scripts/speed-baseline.json --regress-pct 0.10`. The report renders a "Delta vs
baseline" table (baseline ± σ, now ± σ, Δ, Δ%) and flags a row as a regression
only when the move exceeds `--regress-pct` and the 1σ intervals don't overlap.
Added a "Comparing against a baseline" section to the doc spelling out the
round-trip.

3. Text parsing: agreed, it's coupled to the exact print formats in
test_dflash.cpp / test_generate.cpp and is brittle. Keeping it for this first
version as you suggested; I'll open a follow-up to add a `--json` output mode to
the binaries and switch the profiler to consume that (keeping the text path as a
fallback for one release).

PR #445, opened 2026-06-24
fix(kvflash): pooled prefix snapshot at chunk boundary (≥128K restore)

dusterbloom:
Wave-1 KVFlash deep-context. **New commit in this PR:** `feee0a1d` — save the pooled prefix snapshot at a chunk-aligned boundary so ≥128K restore works (was a crash).

**Stacking / merge order:** depends on #428 (pool reservation+sizing), #429 (pager serde), and #430 (pooled chunked prefill core). Those appear as lower commits in this branch's diff; review/merge them first — this branch rebases to drop them once they land, leaving only the snapshot-boundary fix.

**Validation:** rebased clean onto current main; the composed deep-context stack passed 64K NIAH (needle exact, restore engaged) on the integration branch. No standalone agentic gate yet.

<a href="https://cubic.dev/pr/Luce-Org/lucebox-hub/pull/445?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dusterbloom:
Superseded by #466 (kvflash pager + QK residency library — green, mergeable), which extracts the reusable model-agnostic foundation consumed by qwen35/qwen35moe, gemma4, and laguna.

The integration portion of this PR (pooled snapshot-at-chunk-boundary + the e97baee6 correctness fix) will land in a follow-up qwen35 integration PR that stacks on #466. Nothing is lost — both are built from the same clean-perf source.

PR #446, opened 2026-06-24
feat(kvflash): snapshot ledger + QK-pool restore-seed + consume-restored-KV

dusterbloom:
Wave-1 KVFlash deep-context — the warm-prefill payoff.

New commits: snapshot ledger + QK-pool restore-seed (f457afbc, lib+tests); consume-restored-KV on pooled restore instead of re-prefill (cfbf8ede, env KVFLASH_RESTORE_CONSUME, default-on); drafter-history reconstruction from ledger (14b295e1).

Stacks on: #445 (cache-persistence), then #430 core, #429 pager, #428 reservation. Lower commits in this diff are those predecessors; review/merge in order — this rebases to drop them as they land.

Reviewer caveat (correctness): KVFLASH_RESTORE_CONSUME=1 output diverges from the no-consume AR path between ~35K and ~60K tokens. Root cause: chunk-to-block placement permutation changes the FlashAttention key-reduction order, FP non-associativity perturbs the logits, and a near-tie greedy argmax flips. See bench/abc_cache_harness/phase3_divergence_rootcause.md. It is an accepted reuse-vs-recompute flip, NOT a correctness bug — but output is NOT bit-identical at long ctx and there is NO agentic gate yet.

The ~0.1s warm-turn speedup is synthetic NIAH single-needle, not a published result. A real-CLI multi-turn output-equivalence capture is the open follow-up before defaulting this on for any widely-read claim.

<a href="https://cubic.dev/pr/Luce-Org/lucebox-hub/pull/446?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dusterbloom:
Superseded by #466 (kvflash library — green), which lands the pager serialize/deserialize + ledger + QK-pool primitives this PR depended on. The consume-restored-KV wiring lands in the follow-up qwen35 integration PR stacking on #466.

PR #447, opened 2026-06-24
feat(dflash): config-driven drafter converter + spec-decode fixes + rope_parameters Q4 converter

dusterbloom:
Wave-2 base. The config-driven drafter converter + spec-decode perf/correctness fixes, plus a new corrected converter.

New in this PR (commit 166a4097): server/scripts/convert_modal_dflash_to_gguf.py — resolves rope_theta from the nested rope_parameters scope then top-level (legacy), and FAILS LOUD (sys.exit) on a missing theta instead of silently baking a 1M default. This fixes the root cause of the drafter RoPE mismatch: Qwen3.6 drafters publish their true 10M under rope_parameters, which the old top-level-only pick() missed, baking 1M and collapsing long-ctx spec-decode acceptance (gemma4 is genuinely 1M so the bug was invisible there). Q4_K_M output, 12 pure-python tests. Original convert_dflash_to_gguf.py untouched.

Depends on the pooled-prefill core (#430).

<a href="https://cubic.dev/pr/Luce-Org/lucebox-hub/pull/447?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/include/dflash27b.h
+++ b/server/include/dflash27b.h
@@ -36,6 +36,10 @@ extern "C" {
 #define DFLASH27B_DRAFT_BLOCK_SIZE     16
 #define DFLASH27B_DRAFT_N_TARGET_LAYERS 5  // fc projects 5*hidden -> hidden
 #define DFLASH27B_DRAFT_MASK_TOKEN_ID  248070
+// Maximum capture layers the target-side array can hold. Newer DFlash
+// drafters (Qwen3.6) use 8 taps; this ceiling is generous enough for any
+// plausible drafter without wasting stack space.
+#define DFLASH_MAX_CAPTURE_LAYERS      16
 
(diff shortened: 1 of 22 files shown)


dusterbloom:
Closing this stale omnibus after portfolio review. The branch is now conflicting and combines converter, loader-validation, and runtime/spec-decode concerns. The durable pieces will be preserved and proposed as focused PRs on current main rather than asking reviewers to untangle this stack.

PR #452, opened 2026-06-25
feat(server): prefix-aware inline prefix-cache eviction

cheese-cakee:
## Summary

The inline prefix cache evicts by pure LRU. On branched multi-turn workloads that can drop a short, broadly-shared ancestor prefix (system prompt + first turn, reused by many later branches) while keeping a long, conversation-specific leaf snapshot that nothing else reuses. The next branch needing the ancestor then re-prefills it. This change makes eviction prefix-aware: prefer evicting the oldest leaf so shared ancestors stay resident. It degrades to plain LRU when no ancestor structure exists.

This is a trade, not a strict win over LRU. The policy keeps the frozen shallow ancestors plus the current deepest entry; LRU keeps the N most-recent. It wins when later branches reuse an early shared root (the agentic system-prompt pattern) and can lose when a branch reuses a recent but non-current prefix that LRU would still hold. Linear conversations are unaffected: both keep the deepest entry.

## Changes

- `server/src/server/prefix_cache.h` / `.cpp`: new pure helper `select_inline_evict_victim(ids_lru)`; inline `LruEntry` now carries its prefix tokens (`ids`); `prepare_inline_snap` uses the helper instead of always reserving `entries_.front()`.
- `server/test/test_server_unit.cpp`: five model-free unit tests for the policy.

## How it works

Inline entries are kept in LRU order (front = oldest; `move_to_end` on hit). At capacity, instead of evicting the oldest entry unconditionally, the helper scans oldest-first and returns the first entry whose tokens are not a strict prefix of any other live entry (a "leaf"). The longest entry is always a leaf, so a victim always exists; when no entry is an ancestor of another the helper returns index 0 and behavior is identical to LRU. All inline-cache mutation is on the single worker thread, so no locking changes are needed. The reserve/confirm/abort protocol and the PR #370 stale-slot cleanup are unchanged.

## Performance

Correctness is unit-tested (below). The policy only differs from LRU when an ancestor/descendant relation exists among cached prefixes; otherwise it is identical to LRU. When it does differ it keeps the shared ancestor instead of a recent leaf, which is a trade: it raises the inline-cache hit rate / lowers TTFT on branched multi-turn sessions that reuse an early shared root, and can lower it when a session reuses a recent but non-current prefix that LRU would have kept. The net effect is workload-dependent, so this is not claimed as a universal win.

This branch has not been performance-validated end-to-end: an attempted A/B used prompts shorter than the snapshot chunk size, so the inline cache never engaged. The change is shipped on unit-test validation plus the no-change-vs-LRU guarantee when no ancestor structure exists. A representative end-to-end A/B (shared system prompt + first turn, then more-than-capacity distinct continuations) is the suggested validation for maintainers with a multi-turn agentic trace.

## Limitations

- Helps only when the shared ancestor is itself a cached entry (the common multi-turn-progression case). It does not synthesize a shared-prefix entry that was never snapshotted; snapshotting at additional turn boundaries needs a cross-backend change to `GenerateRequest` (one `snap_slot`/`snap_pos` per generate today) and is left as a follow-up.
- Bounded by the inline slot pool and by snapshots existing only at chat-turn boundaries (no token-level sharing).

## Verification

```
cmake --build server/build --target test_server_unit -j32 && ./server/build/test_server_unit
```

1990 assertions, 0 failures. New cases:

- empty / single entry returns index 0 (LRU fallback).
- oldest-first chain `[s]`, `[s,a]`, `[s,a,b]`: evicts the deepest leaf, keeps the shared roots.
- unrelated prefixes: evicts index 0 (LRU).
- branch `[s]`, `[s,a]`, `[s,b]`: spares the shared root `[s]`, evicts the oldest leaf.

### DIFF
--- a/server/src/server/prefix_cache.cpp
+++ b/server/src/server/prefix_cache.cpp
@@ -140,6 +140,38 @@ PrefixHash hash_prefix(const int32_t * ids, int count) {
     return h;
 }
 
+// ─── Prefix-aware eviction ──────────────────────────────────────────────
+
+static bool is_strict_prefix(const std::vector<int32_t> & a,
+                             const std::vector<int32_t> & b) {
+    // True iff `a` is a strict (shorter) prefix of `b`.
+    if (a.size() >= b.size()) return false;
+    return std::equal(a.begin(), a.end(), b.begin());
+}
+
+int select_inline_evict_victim(const std::vector<const std::vector<int32_t> *> & ids_lru) {
+    const int n = (int)ids_lru.size();
+    if (n <= 0) return 0;
+    // Oldest-first scan: evict the first entry that is not a strict prefix of any
+    // other entry (a leaf). Shared ancestor prefixes are thereby kept resident.
+    for (int i = 0; i < n; i++) {
+        bool is_ancestor = false;
+        for (int j = 0; j < n; j++) {
+            if (j == i) continue;
+            if (is_strict_prefix(*ids_lru[i], *ids_lru[j])) { is_ancestor = true; break; }
+        }
+        if (!is_ancestor) return i;  // oldest leaf
+    }
+    return 0;  // unreachable (the longest entry is always a leaf); pure-LRU fallback
+}
+
+int select_inline_evict_victim(const std::vector<std::vector<int32_t>> & ids_lru) {
+    std::vector<const std::vector<int32_t> *> ptrs;
+    ptrs.reserve(ids_lru.size());
+    for (const auto & v : ids_lru) ptrs.push_back(&v);
+    return select_inline_evict_victim(ptrs);
+}
+
 // ─── PrefixCache ────────────────────────────────────────────────────────
 
 PrefixCache::PrefixCache(int cap, const Tokenizer & tokenizer)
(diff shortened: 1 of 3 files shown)


cheese-cakee:
Validated the eviction policy end to end on an RTX 3090 (sm_86, power-capped 200W) with Qwen3.6-27B Q4_K_M. Built two binaries, this branch and its parent (the LRU baseline at 7d5f083), and ran the same workload against both: a shared system prompt, three requests that populate and then pressure a 2-slot inline cache, then a fourth request that returns to the shared root.

Workload (cap 2 slots): R1 `[sys, U_a]`, R2 `[sys, U_a, A_a, U_b]`, R3 `[sys, U_a, A_a, U_b, A_b, U_d]`, then R4 `[sys, U_a]`. R1 commits the ancestor prefix, R2 commits a deeper prefix that has the ancestor as a strict prefix, R3 at capacity forces an eviction. Measured TTFT (first streamed token) on R4, steady state:

| policy | R4 TTFT | server cache line |
|--------|---------|-------------------|
| LRU (parent) | ~2044 ms (miss) | restore=false slot=-1 |
| prefix-aware (this PR) | ~1159 ms (hit) | restore=true slot=0 prefix_len=217 |

About 43 percent lower TTFT (around 885 ms saved) on shared-root reuse under pressure. The mechanism is visible in the logs. LRU evicts the shared ancestor, so R4 is a full re-prefill. This PR keeps it:

```
[pc] prefix-aware evict: victim idx=1 (len=308) kept oldest ancestor (len=217)
```

so R4 restores 192 KV positions and prefills only the 49-token delta.

One note on reproduction: the inline snapshot only commits when the shared prefix crosses a prefill chunk boundary, so I set `DFLASH27B_PREFILL_UBATCH=64` to demonstrate this with a compact 217-token prefix. At the default ubatch of 512 the same win applies to shared system prompts longer than 512 tokens, which is typical for tool-heavy agentic system prompts. As called out in the PR, linear conversations and recent-but-not-current prefixes are unaffected or can lose; this is a win specifically on early-shared-root reuse.

davide221:
lgtm

PR #466, opened 2026-06-30
feat(kvflash): pager serialize/deserialize + QK residency library

dusterbloom:
## What

Model-agnostic KV-paging primitives in `server/src/common/`, already consumed by the **qwen35/qwen35moe, gemma4, and laguna** backends. This is the reusable foundation extracted from the kvflash snapshot-chain work — lands independently of any backend integration.

## Why split it out

The full kvflash snapshot chain (~4000 LOC) couples tightly into `qwen35_backend.cpp`. The **library** (pager + QK scoring) does not — it touches only `common/` headers + tests. Landing it standalone gives every backend a shared, tested KV-paging + residency layer without the integration-review burden.

## Contents
- `kvflash_pager.h` — `serialize(max_chunks)`/`deserialize` v1 format with ledger section (was_resident + qk_score + dtype) for pooled-prefix snapshot/restore; critical-chunk pinning + deadlock guard; `floor_to_chunk`; `identity_prefix_covers` (preserved for existing consumers).
- `kvflash_qk.h` — cosine QK residency scoring with seeded-fallback; **overflow-safe `seeded_n`** (limit defaults to 0, not `n_chunks`, preventing OOB reads when a caller omits the length).

## Verification
- Builds against main's `llama.cpp` submodule (**no fork dependency**).
- `test_kvflash_pager` PASS: serde + pinning + partial-serialize + ledger round-trip.
- `test_kvflash_qk` PASS: pure-math scoring incl. seeded-sentinel fallback.
- Full `dflash_common` (main's server code) recompiles — `identity_prefix_covers` keeps laguna/gemma4 working.
- CPU-only; no GPU required.

## Scope
5 files, +713/−18. No server-backend files, no bench/docs noise. Supersedes the library portions of #429/#445/#430/#446 (those integration slices remain tracked separately).

Follow-up PR (qwen35 pooled snapshot-chain integration) will stack on top of this.

<a href="https://cubic.dev/pr/Luce-Org/lucebox-hub/pull/466?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #469, opened 2026-06-30
fix(qwen35): route non-tree GDN capture through persist buffer (src[7])

dusterbloom:
## Summary

Route Qwen35 non-tree GDN capture through the persistent intermediate buffer (`src[7]`) instead of copying from the GDN result tensor embedded intermediate region.

This preserves the current main/#465 pure-chain prefill optimization: `ggml_gated_delta_net_set_skip_intermediate()` is still applied only when `can_skip_gdn_intermediate` is true. Capture paths keep intermediates available through the explicit persist buffer.

## Why

Fast rollback / chain verify needs per-token SSM intermediate states for rollback. Tree verify already had a persist path via `_tree_persist`; non-tree capture was still using the legacy result-region copy path. Routing non-tree capture through `src[7]` avoids allocator lifetime issues and keeps the rollback buffer populated directly.

The unsupported fallback now fails loudly if capture is requested with a non-F32/F16 intermediate buffer, rather than leaving rollback state stale.

## Relationship

- Rebased on current `main` including #465.
- Keeps #465 skip-intermediate behavior intact.
- Compatible with the reworked ggml PR24 / hub #473 shape, where in-place GDN is explicit and pure-AR-only rather than a global GDN default change.
- Independent of #466 KVFlash library.

## Validation

- `env CCACHE_DIR=/tmp/lucebox-ccache CCACHE_TEMPDIR=/tmp/lucebox-ccache-tmp cmake --build server/build --target test_server_unit -j 8`
- `server/build/test_server_unit`
- Result: `2022 assertions, 0 failures`

### DIFF
--- a/server/src/qwen35/qwen35_target_graph.cpp
+++ b/server/src/qwen35/qwen35_target_graph.cpp
@@ -886,7 +886,16 @@ static ggml_tensor * build_delta_net_block(
     // tree_persist writes directly to the intermediate buffer. It only supports
     // F32/F16 output; for Q8_0 intermediates, fall back to the legacy ggml_cpy
     // path which handles F32→Q8_0 quantization automatically.
-    ggml_tensor * persist_inter = (parent_ids && cap && cap->ssm_intermediate_states
+    // persist_inter: when capture is requested, route the kernel's per-token
+    // intermediate-state writes DIRECTLY into the persistent cache buffer via
+    // src[7], avoiding the legacy result-region cpy. Works for BOTH tree and
+    // non-tree (chain-verify) capture — the kernel checks src[7] regardless of
+    // tree mode, and write_inter is forced true whenever src[7] is non-null.
+    // This is required for non-tree capture once the GDN op defaults to K=1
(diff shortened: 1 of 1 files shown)


davide221:
Validated on lucebox2 (RTX 3090 CUDA, stability profile, Qwen3.6-27B-Q4_K_M + dflash-draft-3.6-q8_0, greedy):
- Dense AR and spec-decode outputs **byte-identical** to main (6/6 hashes), spec accept stats identical (326/1440, avg_commit 4.27) -> the src[7] persist-buffer capture routing is behaviorally equivalent on the chain path
- The red 'GPU tests (gfx1151/ROCm)' check is an **infra flake**, not this PR: that job only runs rocminfo + a HIP vector-add (never builds PR code), and its 14-min failure is the known lucebox3 KFD-wedge signature. Re-run once the runner is rebooted.
- LGTM to merge from the CUDA side.

davide221:
Screenshot 2026-07-02 alle 18 35 06

PR #470, opened 2026-06-30
fix(server): recover single-tool JSON argument responses

dusterbloom:
## Summary

Recover whole-response bare JSON argument objects as a tool call when exactly one tool is declared.

This is intentionally conservative: standalone JSON object only, exactly one declared tool only, rejects prose-wrapped JSON and multi-tool ambiguity, and routes through the existing parser/emitter path.

## Why

The agentic replay produced valid JSON arguments without an explicit tool wrapper, which caused otherwise valid tool turns to be counted as invalid. This preserves correctness without changing multi-tool or prose parsing behavior.

## Validation

- `env CCACHE_DIR=/tmp/lucebox-ccache CCACHE_TEMPDIR=/tmp/lucebox-ccache-tmp cmake --build server/build-cuda126-sm86 --target test_server_unit -j 8`
- `server/build-cuda126-sm86/test_server_unit`
- Result: `2034 assertions, 0 failures`

<a href="https://cubic.dev/pr/Luce-Org/lucebox-hub/pull/470?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/src/server/sse_emitter.cpp
+++ b/server/src/server/sse_emitter.cpp
@@ -23,6 +23,17 @@ static bool has_request_tools(const json & tools) {
     return tools.is_array() && !tools.empty();
 }
 
+static bool has_single_request_tool(const json & tools) {
+    return tools.is_array() && tools.size() == 1 && tools[0].is_object();
+}
+
+static bool starts_with_potential_bare_json_tool(const std::string & text,
+                                                 const json & tools) {
+    if (!has_single_request_tool(tools)) return false;
+    size_t first = text.find_first_not_of(" \t\n\r");
+    return first != std::string::npos && text[first] == '{';
+}
+
 static bool find_tool_start(const std::string & text, size_t & pos) {
     size_t idx = text.find('<');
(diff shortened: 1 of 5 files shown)


davide221:
Validated on lucebox2:
- Unit suite on this branch: **ALL PASSED**
- Live tool-call smoke through dflash_server (Qwen3.6-35B-A3B): single declared tool, bare-JSON-arguments response correctly recovered -> tool_calls=1, arguments parsed, finish_reason=tool_calls
- The red 'Speed profile' check is an **infra flake**, not this PR: the runner OOM'd loading the 15.3GB target model (cudaMalloc failed during target load) — a server-side parsing PR cannot cause that. Re-run should clear it.
- LGTM to merge.

PR #479, opened 2026-07-02
perf(laguna): spec-decode verify path — bonus fold, fused domino head, draft-graph replay, AUTO width, fused QK/shexp weights

davide221:
## What

Speculative-decode verification-path optimization for the Laguna target, profiled and validated end-to-end on RTX 3090 (laguna-xs2 Q4_K_M + DFlash v23 domino drafter, greedy):

| Workload | main (old AUTO) | this PR | plain AR |
|---|---|---|---|
| HumanEval | 172.0 | **192.6** | 181 |
| GSM8K-style | 164.8 | **181.6** | 179.6 |

Spec decode goes from 8% behind AR to **+6.4% ahead on HumanEval**, with **byte-identical outputs** enforced on every change by a 19-output sha256 harness (quicksort-chat + 10 HumanEval + 8 GSM8K prompts, greedy 192-512 tokens).

## Commits

**1. `perf(laguna): verify-path optimizations — bonus fold, fused domino, draft padding, AUTO width`**
- **AUTO verify width fix (the single biggest win):** the old `(int)EWMA + 2` drifted to width 4-5 exactly on high-acceptance workloads; measured marginal verify position costs ~2.3ms while its marginal commit is <0.3 tokens beyond width 3 (HE: w3 187.8 / w4 172.8 / w5 162.1 tok/s). AUTO is now `round(EWMA)+1` capped at `DFLASH_LAGUNA_VERIFY_WIDTH_MAX` (default 3).
- **Bonus-fold (chain greedy):** the bonus token becomes the next iteration's seed instead of running a dedicated 1-token target forward for it (same contract the DDTree path already uses via `next_token`). One target forward per step instead of two; EOS-on-bonus skips the forward entirely. Sampled-verify path untouched.
- **Fused domino head:** the GRU correction loop was ~30 blocking syncs + a full-vocab logits D2H/H2D round-trip per step; now one GPU graph (lm_head projection + unrolled GRU + in-graph argmax→get_rows feedback) on the draft backend stream, replayed by the CUDA-graph cache, with async token readbacks and a single synchronize. Needs a one-time F16 GPU embedding table (392 MiB for the 100k vocab; CUDA get_rows can't read k-quants). Kill switch `DFLASH_LAGUNA_FUSED_DOMINO=0`.
- **Draft-graph CUDA-graph replay:** the draft forward never replayed because (a) its context length grows every step (topology change) and (b) it built its metadata in a fresh 256MB malloc each step, defeating the pointer-keyed graph cache. Now: draft ctx padded to a 64-aligned allocation with -inf pad-key masks (`DFLASH_LAGUNA_DRAFT_PAD=0` to disable; guarded off when SWA windowing is active at the current ctx), plus a persistent `meta_arena` in `StepGraph` (benefits all arches using `build_draft_step`).

**2. `perf(laguna): fused adjacent QK and shexp gate/up weights for decode-width matmuls`**
- The loader lays `attn_q|attn_k` and `ffn_gate_shexp|ffn_up_shexp` adjacently in the weight buffer and binds a fused tensor over each pair — zero extra VRAM, the split tensors remain valid views for all other paths.
- At decode widths the attention builder runs ONE matmul for Q|K, ONE rms_norm+mul with a fused per-head norm weight, ONE rope over all heads (Q and K use identical rope params), then splits with views; the shared expert becomes one matmul + `ggml_swiglu`.
- **Bit-exactness boundary (established with a standalone bitwise test):** MMVQ (n_tokens ≤ 8) is bit-identical under weight row-concatenation — every output row is the same independent dot product. MMQ (prefill) partitions work by total row count and is NOT bit-stable (~1e-5), so the fused weights engage only for `n_tokens <= 8`; prefill keeps the split weights and is unchanged. Kill switch `DFLASH_LAGUNA_FUSED_QK=0`.

**3. `deps: bump llama.cpp`** to `perf/luce-verify-kernels` (companion PR in Luce-Org/llama.cpp: CUDA-graph cache stats, q8_1 quantize memo, MMVQ ncols knob, richer op-error diagnostics — all architecture-agnostic).

## How general is this? (portability beyond Laguna)

- **Architecture-agnostic already:** everything in the llama.cpp companion PR; the persistent `StepGraph` metadata arena (common draft code, all DFlash arches benefit immediately).
- **General technique, Laguna-first implementation (straightforward follow-up ports):**
  - *Bonus-fold* — any chain-verify spec loop with the separate bonus forward (qwen35, qwen35moe, gemma4 share the pattern).
  - *AUTO width cap* — the "marginal verify position vs marginal commit" economics apply to every MMVQ-width verify; each arch's width heuristic can adopt the capped-EWMA form.
  - *Draft-ctx padding* — lives in common `dflash_draft_graph.cpp` behind an opt-in flag; other arches can enable after validating their draft SWA configs.
  - *Fused adjacent weights* — applies to any GGUF arch with same-type q/k or gate/up pairs (qwen35moe has the same shexp pair; the loader pattern and the MMVQ-only gating carry over 1:1).
- **Laguna-specific:** the fused domino head (Laguna's drafter feature), though the single-graph + in-graph-argmax-feedback pattern is reusable for future draft heads.

## Validation

- Byte-identity: all 19 harness outputs sha256-identical to the pre-change reference on every commit (harness: `capture_outputs.sh`, greedy, temperature 0); the fused-weight commit additionally verified with a standalone concat-vs-split bitwise matmul test at the exact laguna shapes.
- nsys profiling before/after each stage (verify exec 9.07ms at w=3 fully CUDA-graph replayed; per-key replay rates 90-99.8% via the new `GGML_CUDA_GRAPH_STATS`).
- Negative results documented to save future effort: MMQ routing at verify widths (-8%), f16 KV cache (neutral), `GGML_CUDA_FORCE_MMQ` (no-op for this path), DDTree at budgets 2/4/8 with this drafter (strictly worse).

https://claude.ai/code/session_01NGcrzzuJsdxU5tLgztaQnK

<a href="https://cubic.dev/pr/Luce-Org/lucebox-hub/pull/479?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/src/common/dflash_draft_graph.cpp
+++ b/server/src/common/dflash_draft_graph.cpp
@@ -24,6 +24,10 @@ static bool draft_has_swa_layers(const DraftWeights & dw) {
 
 // Build draft graph at a given ctx_len into sg. Does NOT touch sg.alloc.
 // mirror_view: if true, uses a view into mirror->target_feat at slot0.
+// ctx_alloc: allocation/topology size of the ctx dimension (>= ctx_len).
+// When ctx_alloc > 0 and differs from the legacy behavior, a full-layer pad
+// mask input is created so the graph topology stays stable while ctx_len
+// grows (CUDA-graph replay for the draft forward).
 static bool build_draft_graph_internal(
     StepGraph & sg,
     const DraftWeights & dw,
@@ -31,11 +35,14 @@ static bool build_draft_graph_internal(
     int ctx_len,
     const DraftFeatureMirror * mirror,
     int mirror_slot0,
-    bool mirror_view) {
+    bool mirror_view,
+    bool pad_masked = false) {
 
+    const size_t arena_sz = 32u * 1024 * 1024;
+    if (sg.meta_arena.size() < arena_sz) sg.meta_arena.resize(arena_sz);
     ggml_init_params ip{};
-    ip.mem_size   = 256 * 1024 * 1024;
-    ip.mem_buffer = nullptr;
+    ip.mem_size   = sg.meta_arena.size();
+    ip.mem_buffer = sg.meta_arena.data();
     ip.no_alloc   = true;
     sg.ctx = ggml_init(ip);
     if (!sg.ctx) return false;
@@ -86,6 +93,18 @@ static bool build_draft_graph_internal(
         ggml_set_input(sg.attn_mask);
     }
 
+    bool any_full_layer = false;
+    for (int i = 0; i < dw.n_layer; i++)
+        if (!dw.layers[i].is_swa) { any_full_layer = true; break; }
+    sg.pad_mask_full = nullptr;
+    if (pad_masked && any_full_layer) {
+        const int total_k = ctx_len + q_len;
+        const int kv_pad = mask_align_up(total_k, MASK_KV_PAD);
+        sg.pad_mask_full = ggml_new_tensor_2d(sg.ctx, GGML_TYPE_F16, kv_pad, q_len);
+        ggml_set_name(sg.pad_mask_full, "pad_mask_full");
+        ggml_set_input(sg.pad_mask_full);
+    }
+
     sg.gf = ggml_new_graph_custom(sg.ctx, 4096, false);
 
     DraftGraphInputs gi{};
@@ -96,6 +115,7 @@ static bool build_draft_graph_internal(
     gi.positions_k       = sg.positions_k;
     gi.lm_head           = lm_head;
     gi.causal_mask_swa   = sg.attn_mask;
+    gi.pad_mask_full     = sg.pad_mask_full;
     DraftGraphOutputs go = build_draft_graph(sg.ctx, dw, gi);
     sg.hidden_states = go.hidden_states;
     sg.logits = go.logits;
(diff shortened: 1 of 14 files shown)


davide221:
All four review findings addressed in d03e7939:

1. **P1 pad-mask null-deref (all-SWA drafts)**: the full-layer pad-mask write is now guarded; all-SWA drafts are covered by the (pad-aware) causal SWA mask alone. Real config-dependent crash — good catch: our v23 draft has mixed layers so it never fired locally.
2. **P2 fused-domino teardown leak**: the f16 embedding-table buffer/context and the dedicated backend instance are now released in ~LagunaDFlashTarget.
3. **P2 persistent meta arena**: step_graph_destroy now clears meta_arena (shrink_to_fit) and resets the reserve size, so park/unpark no longer pins ~32 MiB host per graph.
4. **P2 fused-domino vocab guard**: same compatibility check as the legacy path, one-shot warning + clean fallback.

Validation after the fixes: unit suite 2022 assertions / 0 failures, spec-decode 512-token greedy output hash unchanged (byte-identical).

PR #455, opened 2026-06-26
perf(laguna): align MoE router and verify paths

davide221:
## What This PR Achieves

This is the Laguna autoregressive parity patch with llama.cpp. It keeps PR #455 scoped to the AR target path and removes the later speculative/DDTree experiments from the PR diff.

Main changes:
- Aligns Laguna MoE routing with llama.cpp's sigmoid -> top-k -> get_rows pattern so ggml-cuda can use the top-k MoE fusion path.
- Adds a small shared `moe_router_graph` helper for the fused router node order.
- Auto-enables Laguna head-major KV layout for the plain AR decode path when it is safe to do so.
- Adds head-major KV cache/snapshot/attention views and 256-aligned KV capacity for the Laguna FA path.
- Keeps Laguna KV cache type selection on the existing `DFLASH_KV_TYPE`, `DFLASH_KV_K_TYPE`, and `DFLASH_KV_V_TYPE` controls.
- Bumps the llama.cpp submodule to the companion ggml/CUDA change used by the parity path.

Cleaned scope:
- `5 files changed`
- `261 insertions`
- `90 deletions`
- one clean commit authored by `davide221` with `Codex` as co-author

## Validation

Checked locally:

```bash
git diff --check 26ca0dcb6a198628b6cab2b0a2ffbb901baaa564 b8a9f46643674a662524fffa402d54f1593ef4a2
```

Built on the Laguna CUDA bench box:

```bash
cmake -B /home/lucebox/lucebox-hub-pr455-clean-code/server/build-pr455-clean-cuda \
  -S /home/lucebox/lucebox-hub-pr455-clean-code/server \
  -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_CUDA_ARCHITECTURES=86

cmake --build /home/lucebox/lucebox-hub-pr455-clean-code/server/build-pr455-clean-cuda \
  --target bench_laguna_generate dflash_server -j 8
```

The clean build produced both `bench_laguna_generate` and `dflash_server`. BSA was disabled in this copied verification tree because only the llama.cpp submodule was synced; that does not affect this Laguna AR compile check.

## Not Included

- DDTree behavior changes.
- GPU speculative accept / verify plumbing.
- Speculative fallback/gating/profiling experiments.
- Domino-head experiments.
- SWA-window experiments.
- The later experimental MoE down+combine kernel, which was not byte-identical.

Companion ggml/CUDA PR:
https://github.com/Luce-Org/llama.cpp-dflash-ggml/pull/23

PR #482, opened 2026-07-02
perf(laguna): DSpark Markov head + spec-verify stack - 206 to 249 tok/s (RTX 3090)

davide221:
## Summary

Laguna spec-decode performance stack: **206 → 249.6 tok/s on HumanEval** (laguna-xs2 Q4_K_M, RTX 3090, greedy), with every step gated on output-hash identity or measured acceptance parity.

| stage | tok/s (HE, w=best) |
|---|---|
| main @ 737cd47f, v24 drafter f16 | 206.0 (w3) |
| + DSpark Markov head (#460/#461, rebased) | 208.8 (w4) |
| + fused Markov head (one graph, no host round-trip) | 210.8 (w4) |
| + ggml dense MMVQ→MMQ crossover (lucebox-ggml#30) | 236.6 (w6) |
| + q8_0 draft (acceptance-neutral) | 241.3 (w6) |
| + persistent verify step-graph | 244.1 (w6) |
| + q4-mix draft (acceptance-neutral) | **249.6 (w6)** |

## Commits

1. **deps**: ggml spec-verify kernel stack (Luce-Org/lucebox-ggml#30) - dense MMVQ→MMQ crossover default, 16-token MoE `MUL_MAT_ID` fast path.
2. **draft: DSpark head loading** - incorporates #461 rebased onto main (post-#479). Old draft GGUFs are unaffected; the `dflash.dspark.*` tensors are optional. (The draft LoRA variants from #460 were initially bundled here and later removed in 55e53f0; #460 stays open for a future pass.)
3. **laguna: fused DSpark Markov head** for chain drafting and DDTree candidate generation - single graph on the draft stream (lm_head + unrolled Markov argmax→get_rows chain), replacing the per-candidate graph rebuild and host logits round-trip. Chain accept +2.2pts at w3; DDTree commit 3.83→4.43 at budget 12. Kill switches: `DFLASH_LAGUNA_FUSED_DSPARK=0`, `DFLASH_LAGUNA_DSPARK_TREE=0`.
4. **laguna: persistent verify step-graph** - the verify graph is structurally identical across steps within a (width, mask-window) key (kv position enters only through input data), so the built graph is cached and each step reduces to input updates + compute. Byte-identical outputs; `DFLASH_LAGUNA_PERSIST_VERIFY=0` kill switch.
5. **scripts: draft requantize tool** - `quantize_dflash_draft.py` with `q8_0` and `q4-mix` schemes; q4-mix keeps the `dflash.*` heads at q8_0, which is what preserves acceptance (64.0% vs 63.5% f16 at w6).

## Production recipe

```
dflash_server laguna-xs2-Q4_K_M.gguf \
  --draft laguna-v24-...-q4mix.gguf --verify-width 6
```

No env vars required; the width-6 optimum follows from the cheaper verify slope (2.0 → 1.13 ms/width) plus the Markov head's deep-candidate corrections (depth-4 marginal acceptance 38% → 55%).

## Correctness methodology

- Chain w6 output hashes identical across: persistent-graph on/off, ggml mmid extension, and the final published tree vs the development branch (0/19 mismatches at every gate).
- Acceptance/commit tracked at every stage; q8_0 and q4-mix drafts measured acceptance-neutral.
- Verify-width (2-8) and DDTree-budget (7-24) sweeps; nsys graph-node profiles before/after kernel routing changes.

## Notes

- Supersedes #461 (rebased as commit 2). #460 (draft LoRA variants) is deferred: its code was removed from this PR in 55e53f0 and the PR stays open for later.
- DDTree on Laguna remains slower than the tuned chain at greedy (measured: equal commit at equal verify tokens, tree pays ~1.5ms extra overhead); the Markov-corrected candidates and the 16-token verify path mainly benefit sampled/degraded-acceptance regimes.
- Temperature-mode serving currently drops to 98.8 tok/s (sampled-verify: full-logits D2H + CPU sampling + no bonus-fold); a follow-up port of the #478 GPU sampler to the Laguna path is the identified fix (~135-145 tok/s projected).

<a href="https://cubic.dev/pr/Luce-Org/lucebox-hub/pull/482?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -258,6 +258,7 @@ add_library(dflash_common STATIC
     src/laguna/laguna_dflash_target.cpp
     src/common/backend_ipc.cpp
     src/common/domino_head.cpp
+    src/common/dspark_head.cpp
     src/common/target_shard_ipc.cpp
     src/common/target_shard_ipc_daemon.cpp
     src/common/dflash_feature_ring.cpp
--- a/server/scripts/convert_dflash_to_gguf.py
+++ b/server/scripts/convert_dflash_to_gguf.py
@@ -244,6 +244,17 @@ DOMINO_TENSOR_MAP = {
 }
 
 
+DSPARK_TENSOR_MAP = {
+    "dspark_markov_head.markov_w1.weight": ("dflash.dspark.markov.w1", gguf.GGMLQuantizationType.F16),
+    "dspark_markov_head.markov_w2.weight": ("dflash.dspark.markov.w2", gguf.GGMLQuantizationType.F16),
+}
+
+DSPARK_CONFIDENCE_TENSOR_MAP = {
+    "dspark_confidence_head.weight": ("dflash.dspark.confidence.weight", gguf.GGMLQuantizationType.F16),
+    "dspark_confidence_head.bias": ("dflash.dspark.confidence.bias", gguf.GGMLQuantizationType.F32),
+}
+
+
 def add_domino_aux_heads(writer, arch: str, aux_path: Path | None):
     if aux_path is None:
         return
@@ -265,6 +276,9 @@ def add_domino_aux_heads(writer, arch: str, aux_path: Path | None):
         print(f"[error] Domino aux heads file is not a tensor dict: {aux_path}", file=sys.stderr)
         sys.exit(1)
 
+    if not any(k in state for k in DOMINO_TENSOR_MAP):
+        return
+
     missing = [k for k in DOMINO_TENSOR_MAP if k not in state]
     if missing:
         print(f"[warn] incomplete Domino aux heads; missing {missing}; skipping Domino tensors")
@@ -282,6 +296,79 @@ def add_domino_aux_heads(writer, arch: str, aux_path: Path | None):
     writer.add_uint32(f"{arch}.dflash.domino.vocab_size", vocab)
 
     for st_name, (gguf_name, raw_dtype) in DOMINO_TENSOR_MAP.items():
+        t = state[st_name]
+        if hasattr(t, "detach"):
+            t = t.detach().cpu()
+        arr = t.float().numpy()
+        if raw_dtype == gguf.GGMLQuantizationType.F16:
+            arr = arr.astype("<f2")
+        else:
+            arr = arr.astype("<f4")
+        writer.add_tensor(gguf_name, arr, raw_dtype=raw_dtype)
+        print(f"[tensor] {gguf_name:50s} aux ->{raw_dtype.name:4s} {tuple(arr.shape)}")
+
+
+def add_dspark_aux_heads(writer, arch: str, aux_path: Path | None):
+    if aux_path is None:
+        return
+    if not aux_path.exists():
+        return
+
+    try:
+        import torch
+    except ImportError as exc:
+        print(f"[error] --aux-heads requires torch: {exc}", file=sys.stderr)
+        sys.exit(1)
+
+    state = torch.load(aux_path, map_location="cpu")
+    if isinstance(state, dict) and "state_dict" in state and isinstance(state["state_dict"], dict):
+        state = state["state_dict"]
+    if not isinstance(state, dict):
+        print(f"[error] DSpark aux heads file is not a tensor dict: {aux_path}", file=sys.stderr)
+        sys.exit(1)
+
+    missing = [k for k in DSPARK_TENSOR_MAP if k not in state]
+    if missing:
+        return
+
+    print(f"[info] reading DSpark aux heads from {aux_path}")
+    w1 = state["dspark_markov_head.markov_w1.weight"]
+    w2 = state["dspark_markov_head.markov_w2.weight"]
+    vocab = int(w1.shape[0])
+    rank = int(w1.shape[1])
+    if tuple(w2.shape) != (vocab, rank):
+        print(f"[error] DSpark markov_w2 shape {tuple(w2.shape)} != {(vocab, rank)}", file=sys.stderr)
+        sys.exit(1)
+
+    writer.add_uint32(f"{arch}.dflash.dspark.enabled", 1)
+    writer.add_uint32(f"{arch}.dflash.dspark.markov_rank", rank)
+    writer.add_uint32(f"{arch}.dflash.dspark.vocab_size", vocab)
+
+    for st_name, (gguf_name, raw_dtype) in DSPARK_TENSOR_MAP.items():
+        t = state[st_name]
+        if hasattr(t, "detach"):
+            t = t.detach().cpu()
+        arr = t.float().numpy().astype("<f2")
+        writer.add_tensor(gguf_name, arr, raw_dtype=raw_dtype)
+        print(f"[tensor] {gguf_name:50s} aux ->{raw_dtype.name:4s} {tuple(arr.shape)}")
+
+    conf_missing = [k for k in DSPARK_CONFIDENCE_TENSOR_MAP if k not in state]
+    if conf_missing:
+        print(f"[warn] incomplete DSpark confidence head; missing {conf_missing}; Markov head will still load")
+        return
+
+    conf_w = state["dspark_confidence_head.weight"]
+    conf_b = state["dspark_confidence_head.bias"]
+    confidence_dim = int(conf_w.shape[1])
+    if int(conf_w.shape[0]) != 1 or tuple(conf_b.shape) != (1,):
+        print(
+            f"[error] DSpark confidence shapes weight={tuple(conf_w.shape)} bias={tuple(conf_b.shape)}",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+    writer.add_uint32(f"{arch}.dflash.dspark.confidence_dim", confidence_dim)
+    writer.add_uint32(f"{arch}.dflash.dspark.confidence.enabled", 1)
+    for st_name, (gguf_name, raw_dtype) in DSPARK_CONFIDENCE_TENSOR_MAP.items():
         t = state[st_name]
         if hasattr(t, "detach"):
             t = t.detach().cpu()
@@ -400,6 +487,7 @@ def main():
     if not args.no_aux_heads:
         aux_path = args.aux_heads if args.aux_heads is not None else args.safetensors.parent / "dflash_aux_heads.pt"
     add_domino_aux_heads(writer, ARCH, aux_path)
+    add_dspark_aux_heads(writer, ARCH, aux_path)
 
     print(f"[info] writing {args.out_gguf}")
     writer.write_header_to_file()
--- a/server/scripts/quantize_dflash_draft.py
+++ b/server/scripts/quantize_dflash_draft.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python3
+"""Requantize a DFlash draft GGUF (f16) for faster spec-decode drafting.
+
+Schemes:
+  q8_0    - all 2D matmul weights to q8_0 (~47% smaller, acceptance-neutral)
+  q4-mix  - backbone (blk.*) weights to q4_0, dflash.* head weights kept at
+            q8_0 (~67% smaller; the head split protects the Markov/projection
+            bias precision that near-tie corrections depend on).
+
+Norms, biases and non-f16 tensors are copied as-is. Metadata is preserved,
+so the output loads anywhere the input does.
+
+Measured on Laguna-XS.2 + v24 drafter (RTX 3090, HumanEval, verify width 6):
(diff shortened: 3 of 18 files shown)


davide221:
Pushed 8c02575 addressing all 7 cubic findings:

- **draft_gguf_loader**: bounds check now uses the overflow-safe `gguf_tensor_in_file()` with the `gguf_bounds_error()` diagnostic (P1), and tensor type / byte-size validation happens before the host buffer resize (P2).
- **laguna_verify_batch**: the persistent verify slot is keyed on backend/weights/cache identity (P1), and reuse is restricted to the kv_idx input-data path; the NO_KVPAD / PAD_CPY fallbacks rebuild every step since they bake `kv_start` into the graph structure (P1). The gallocr is recreated on backend change.
- **server_main**: `--draft-lora` is rejected for non-laguna targets and for layer-split placement instead of being silently ignored (P2).
- **LagunaBackend**: dead `active_draft_lora_` member removed (P3).
- **quantize_dflash_draft.py**: metadata array parts go through `np.asarray()` before `.item()` (P1). The original expression actually works with the in-tree gguf-py (parts are numpy memmap slices; verified against the v24 drafts, which carry an INT32 `target_layer_ids` array), so this is hardening rather than a crash fix.

Validation on RTX 3090 (lucebox, CUDA 13.2, sm_86):

- clean build, no new warnings in touched files; ctest 12/12 pass
- `quantize_dflash_draft.py --scheme q4-mix` output is byte-identical (`cmp`) to the previously generated v24 q4mix draft
- `--draft-lora` on a gemma4 target fails fast with a clear error; on a laguna target the flag passes the gate and a bad adapter path yields the loader error
- chain w6 output hash identical between `DFLASH_LAGUNA_PERSIST_VERIFY=0` and `=1` under the new reuse gate

Environment note, unrelated to this PR: fresh builds of both this branch and its merge-base 737cd47f produce deterministic gibberish laguna output on lucebox itself (accept 16.7%, avg_commit 1.00, same hash across CUDA 12.0 and 13.2 toolchains and persist on/off), while a Jun 23 binary from the laguna-ddtree lineage serves the same GGUFs correctly at 176 tok/s. The problem pre-dates this PR (window: hub 0788827c..737cd47f and/or ggml 574be613..ac06e543, possibly interacting with the reinstalled box stack: kernel 6.17, driver 595). Deserves a separate issue; this PR's numbers come from the cloud 3090 per GATE_RESULTS.

davide221:
Follow-up: the "broken laguna output on lucebox" from my previous comment is root-caused and fixed in ac531b0. It was not box-specific at all.

**Root cause** (a `main` regression from #455, not introduced by this PR): `resolve_laguna_kv_types()` feeds laguna's KV dtype through `resolve_kv_types()` whenever any `DFLASH27B_KV_*` env var is present. `dflash_server` auto-sets `DFLASH27B_KV_TQ3=1` when `max_ctx > 6144` (a qwen-27B VRAM heuristic), and laguna's default `max_ctx` is 8192. So every laguna `dflash_server` run since #455 silently used a **TQ3_0 ternary KV cache** instead of Q8_0 and produced degenerate output on every decode path (accept 16.7%, avg_commit 1.00). Runs with smaller ctx or explicit `--cache-type` flags (e.g. the pod gate harness) never triggered it, which made it look machine-specific.

**Fix**: laguna now honors only the explicit per-axis `DFLASH27B_KV_K` / `_KV_V` overrides (`--cache-type-k/v`), logs them when applied, and ignores the legacy F16/Q4/TQ3 shorthands.

With this, the PR stack validates end to end on real hardware (lucebox RTX 3090, CUDA 13.2, default env):

- laguna-xs2 Q4_K_M + v24 q4mix draft, `--verify-width 6`: coherent output at **272 tok/s, accept 71.1%, avg_commit 4.27** (single HE-style prompt; PR headline was 249.6)
- output hash byte-identical with `DFLASH_LAGUNA_PERSIST_VERIFY=0` vs default
- explicit `q4_0` override still honored and logged
- ctest 12/12

Bisect trail for the record: main@737cd47f (PR base) already reproduced the garbage; commit bisect converged on #455; file-level grafts isolated `laguna_backend.cpp`; hunk bisect isolated `resolve_laguna_kv_types()`; the trigger is `server_main`'s `setenv("DFLASH27B_KV_TQ3", "1", 0)` at `max_ctx > 6144`. Note TQ3_0 KV specifically garbles laguna while explicit q4_0/q8_0 work, so the tq3 fattn path deserves a separate look if laguna+TQ3 is ever wanted.

davide221:
ae89446 addresses the follow-up cubic finding on the KV override path: the supported-pair check is extracted from `resolve_kv_types()` into a shared `dflash::validate_kv_pair_or_abort()`, and the laguna per-axis override path now calls it. An unsupported combination (e.g. `DFLASH27B_KV_K=tq3_0 DFLASH27B_KV_V=q4_1`) aborts at startup with the supported-pairs listing instead of failing later inside a CUDA kernel; the listing stays single-sourced with the qwen path.

Verified on the 3090: invalid pair aborts with the listing, valid `q4_0/q4_0` override still honored + logged + coherent, default q8_0 path untouched, ctest 12/12.

davide221:
b9e1146 adds two things from a kernel-optimization pass toward 300 tok/s on the 3090:

- **Pinned async input staging**: per-step graph inputs now upload via `tensor_set_async` from slot-owned pinned buffers instead of ~6 pageable copy+sync pairs per step. Byte-identical output; +1-2 tok/s at w6, grows with context (mask uploads scale with kv_pad).
- **`DFLASH_LAGUNA_STEP_PROF=1`**: per-step phase breakdown (draft / heads / verify / other) printed with the spec summary.

Measured state at w6 (282.6 tok/s, laguna-xs2 + v24 q4mix): step = 15.2 ms of which verify graph = 11.8 ms (11.6 ms pure GPU at 98% occupancy of its window, confirmed via `--cuda-graph-trace=node`), draft 2.2, DSpark 0.75, glue 0.4. CUDA-graph replay already covers verify (93%) and the draft chain (84%).

Inside the verify GPU time, top items per step: MoE expert matmuls ~3.8 ms (`mul_mat_vec_q_moe` at ~20% effective bandwidth - it re-reads expert weights once per token at batch 7), dense MMQ+fixup+quantize ~3.7 ms, rope ~0.76 ms (80 launches of 12 us), KV `set_rows` quant ~0.6 ms, `quantize_mmq_q8_1` ~0.4 ms (261 launches).

Negative result worth recording: routing K-quant mmid to the MMQ path at batch 7 (crossover 8 -> 2) was a clear regression, verify 11.8 -> 17.5 ms - MMQ tile machinery across 256 experts costs more than MMVQ's redundant reads at 512-dim experts. Reverted; the winning move for the MoE 3.8 ms is an expert-grouped MMVQ variant (read each selected expert once, apply to all its tokens), which is a real kernel project, est. 1.5-2 ms/step.

Related: Luce-Org/lucebox-ggml#31 fixes the mmid CUDA-graph gate to mirror the dispatch (MMQ path is sync-free), which keeps tree-verify widths replayable; no pin bump needed for this PR.

Realistic roadmap from 283: expert-grouped mmid MMVQ (+~1.7 ms) -> ~305-315; rope qk fusion + KV-quant batching (+~0.8 ms) -> ~320-330. Beyond that acceptance becomes the binding constraint.

davide221:
Completing the kernel-optimization ledger from the last comment, after exhausting the bit-exact space (all tested on the 3090 at w6, byte-identity gated on output hashes):

Already banked by this PR stack (verified active): q/k rope fused at graph level (#479 fused-qk), q8_1 activation memoization, CUDA-graph replay on verify (93%) and draft chain, MMQ dense crossover at width>3.

New negative results, for the record:
- `LUCE_MMVQ_MAX_NCOLS=8` (dense MMVQ at w6, retested now that graphs replay): verify 11.8 -> 14.0 ms, 240 tok/s, different bytes. The MMQ crossover tuning stands.
- mmid MMQ at batch 7 (previous comment): 17.5 ms verify. MMVQ stands.
- Recomputed with unique-expert accounting (56 draws over 256 experts = ~50 unique): the MoE mmvq kernels already run at ~975 GB/s effective, i.e. at the memory roofline. The earlier "20% bandwidth" read was wrong; there is no expert-grouping win at this expert count. Kernel idea withdrawn.
- ggml's rope+set_rows fused shader can't take laguna's KV writes (q8_0 dst + i32 ids vs f16/i64 gate), and the fused-qk rope output feeds Q/FA anyway.
- The multi-stream `concurrent_events` scaffolding in ggml-cuda has no populator (dormant upstream infra).

Where this leaves the step (15.2 ms, 283 tok/s): verify 11.8 ms of which MoE ~3.8 (roofline, frozen), dense MMQ stack ~3.7 (the only real kernel headroom, but any change alters rounding), small ops ~1.7 (dual K/V set_rows fusion is the last bit-exact item, worth only ~0.3 ms).

Conclusion: under strict output-byte identity the ceiling is ~290 tok/s with this drafter. The two honest paths to 300+: (a) a small-batch dense matmul kernel for sm_86 (accepting rounding changes, gated on acceptance parity like this PR's other kernel swaps), or (b) acceptance: avg_commit 4.5 at the current step time clears 300 with zero numeric risk - a v27 drafter tuned for w6 acceptance is the cheapest 300.

davide221:
55e53f0 removes the draft LoRA variant support that had been folded in from #460 (deferred by request): the `--draft-lora` flag, the `extra_body.draft_lora` request selector, the GGUF LoRA adapter loader/merger in `draft_gguf_loader.cpp`, and the `DraftLoraSpec` plumbing (-481 lines). The DSpark Markov head from #461 stays, as does the single-"base" draft-variant container so a variant mechanism can return without re-plumbing.

Validation on the 3090: output hashes byte-identical to b9e1146 (`d6bb1a78` / `468cd6ce`), 282 tok/s at w6 with 71.1% accept, DSpark head active, `--draft-lora` now rejected as an unknown option, ctest 12/12. PR description updated accordingly - this PR now supersedes only #461; #460 remains open for a future pass.

Note: this also obsoletes two of the earlier cubic fixes by deletion (the LoRA loader bounds/resize hardening and the `--draft-lora` arch gate went away with the code they guarded). The shared `validate_kv_pair_or_abort()` and everything else stands.

davide221:
55d7b69 lands the QKV projection fusion discussed in the perf thread: attn_v joins the #479 load-time weight fusion in the 20 layers where it shares the q/k quant type (Q6_K attn_v layers keep qk+v). One projection matmul per fused layer at decode widths; prefill keeps split weights, same kill switches as qk.

Safety gate (paired same-session A/B vs `LUCE_QK_FUSE_LAYERS=0`, since fused vs split changes MMQ partial-sum order): he +1.4%, math +6.6%, agent +3.0% tok/s with GSM/MATH gold accuracy identical (7/10, 5/10) and all expectation checks passing; ctest 12/12. Single-prompt w6: code 288 tok/s, math **310 tok/s**.

Benchmarking caveat worth recording: this 3090's suite baselines from earlier sustained runs were depressed by VRAM heat (the known 0x20 SW-slowdown behavior), which made naive cross-session deltas look like +20-30%. All claims above are from paired runs on an equally-warm card; treat absolute suite numbers as session-relative.

davide221:
7005de6 bumps the ggml pin to lucebox-ggml `perf/dual-set-rows` (PR incoming there): the per-layer K/V cache appends now run as ONE fused kernel launch instead of two, **bit-identical outputs** (each grid half executes the exact `k_set_rows_quant` element math).

Final benchmark battery on the complete stack (QKV fusion + dual set_rows + pinned staging), RTX 3090, w6, v24 q4mix:

| bench | result | quality |
|---|---|---|
| code prompt (warm) | 288-290 tok/s, accept 72.3%, commit 4.34 | text stable across all safe changes |
| math prompt | **317 tok/s**, accept 78.9%, commit 4.73 | correct (80 km/h) |
| HE suite | 216.5 | 10/10 checks |
| GSM suite | 220.7 | gold 7/10 (unchanged) |
| MATH suite | 243.5 | gold 5/10 (unchanged) |
| Agent suite | 119.8 | 6/6 checks |
| AR (target-only) | 197.6 tok/s | - |
| qwen3.6-27B | coherent | generic ggml changes regression-checked |
| ctest | 12/12 | - |

Byte-identity chain: dual set_rows verified hash-identical to the QKV-fusion build on code/math/long prompts; QKV fusion itself was gated on paired A/B acceptance parity plus unchanged suite accuracy (its numeric deltas even left the code-prompt text byte-identical).

Issue #457, opened 2026-06-27
gfx1100 (RX 7900 XTX, ROCm 6.4.1): dflash_server ~2x slower than llama.cpp, DDTree speculation never engages, SIGSEGV at long context


L0GYKAL:
## Environment
- **GPU:** AMD Radeon RX 7900 XTX (**gfx1100**, RDNA3), 24 GB
- **ROCm:** 6.4.1 (rocm/dev-ubuntu-22.04:6.4.1 builder + runtime)
- **lucebox-hub:** commit `96100834eca003c2bc7cf76eb943b95fcd6d06a0` (+ pinned llama.cpp submodule `30c9d7da…`)
- **Build:** `DFLASH27B_GPU_BACKEND=hip`, `DFLASH27B_HIP_ARCHITECTURES=gfx1100`, `SM80_EQUIV=OFF` (Phase-1 HIP, matches the default in `Dockerfile.rocm`)
- **Model:** unsloth `Qwen3.6-27B-Q4_K_M.gguf` (target) + `Lucebox/Qwen3.6-27B-DFlash-GGUF/dflash-draft-3.6-q4_k_m.gguf` (drafter)
- Built/run in rootless podman with `--device /dev/kfd --device /dev/dri --group-add keep-groups`.

## Summary
On gfx1100, `dflash_server` is **~2× slower than upstream llama.cpp** for the same model on the same GPU, its **DDTree speculative decode never engages** (silently falls back to autoregressive decode), and it **SIGSEGVs at long context**. Filing in case gfx1100 is intended to be supported — or to confirm the HIP speculative path is gfx1151-only.

## Benchmarks — Qwen3.6-27B Q4_K_M, f16 KV cache, full GPU offload
| Engine | Decode | Prompt eval | 56k-token prefill |
|---|---|---|---|
| llama.cpp (Vulkan/RADV, b64739ea3) | **36.4 tok/s** | 67 tok/s | prefill 641 tok/s, decode 32 tok/s, stable (20.8 GB) |
| `dflash_server` (HIP, Phase-1) | **17.6 tok/s** | — | **SIGSEGV during prefill** |

Same 300-token generation prompt, `temperature` matched, both with f16 KV.

## 1. DDTree speculation never engages on gfx1100
Started with the matched drafter and DDTree on:
```
dflash_server /models/Qwen3.6-27B-Q4_K_M.gguf \
  --draft /models/dflash-draft-3.6-q4_k_m.gguf \
[...]

PR #434, opened 2026-06-22
Custom topk-logprobs kernel + Remove a redundant Device to Host logits copy.

pramodith:
## What does the PR do?

The extraction of top-k vocab indices from the logit scores of the draft model currently happens on the cpu. This involves a costly transfer of the entire logit scores from the GPU to CPU i.e. Device to Host.  **In this PR we move the top-k computation to the GPU entirely.**

This PR also adds a verify-argmax fastpath that skips the 
 logits transfer from Device to Host and re-uses the argmax indices stored as an attribute. It can be toggled via DFLASH_GPU_VERIFY_ARGMAX.  It re-uses the same CPU fallback on any out-of-range index.

A comment stated that previous builds had issues with GPU based verify-argmax, this PR adds a check to detect failures and resort to using the prior CPU method for running the top-k computation in this scenario.

The GPU kernel and verify-argmax is enabled by default but can be disabled via the `DFLASH_GPU_DRAFT_TOPK` and `DFLASH_GPU_VERIFY_ARGMAX` flags.

## Files edited

* Minor bug fixes to `bench_llm.py` to use the right tokenizer + correct the naming conventions of datasets being used.
* `geometric_draft_topk_cuda.cu/h`: Creates a new cuda kernel used by the draft model to identify the topk (where k is in the range 1-8) vocab indices from the logit scores.
* `qwen35_dflash_target.cpp`: Invokes the new `.cu` kernel and falls back to the cpu execution of topk incase there are errors.
* `test_dflash.cpp`: Invokes the `.cu` kernel if the `DFLASH_GPU_DRAFT_TOPK` flag is enabled.  Uses the fast-path verification of draft tokens if `DFLASH_GPU_VERIFY_ARGMAX=1`, if it is set to 2 it identifies/reports any mismatches between cpu and gpu based verification.
* `test_draft_topk_cuda.cpp`: Correctness tests for the cuda kernel by comparing it against the previously used cpu function.
* `CMakeLists.txt`: Register the new cuda kernel.

## Results
**Results in an ~27% increase in tok/s. (105.63-82.80)/(82.8)**
Numbers can be reproduced by running bench_llm.py after re-building with the changes.
The benchmark was run on an _RTX 3090_.

### Results before
```
[bench] ==== HumanEval (n=10, n_gen=256) ====
  [01/10] n_tok=  92  AR= 33.96  DFlash=  98.69  AL= 6.74
  [02/10] n_tok= 146  AR= 33.93  DFlash=  76.72  AL= 5.33
[...]

PR #488, opened 2026-07-05
perf(hip): compile DFlash GPU draft top-K kernel for HIP

cheese-cakee:
## Summary

The DFlash GPU draft top-K + log-prob kernel (`geometric_draft_topk_cuda.cu`, added in #434) was compiled only on the CUDA backend. On HIP/ROCm builds `DFLASH27B_HAVE_DRAFT_TOPK_CUDA` was never defined, so `Qwen35DFlashTarget::project_hidden_to_topk` always took the CPU fallback: a full `vocab x n_tokens` device-to-host logits copy plus a CPU heap extract on **every** draft step. On a bandwidth-bound card that per-step D2H is a plausible contributor to the gfx1100 "~2x slower than llama.cpp" report in #457.

This compiles the exact same kernel for HIP. The body is pure arithmetic (no tensor cores, no CUDA-only intrinsics), so it builds unchanged through the existing `hip_compat` shim, the same way `flashprefill_kernels.hip.cu` is built.

## Changes

- `server/src/common/geometric_draft_topk.hip.cu` (new): a thin TU that `#include`s the shared kernel body and is compiled with `LANGUAGE HIP` + `-I hip_compat`. On CUDA builds the `.cu` is still added directly; on HIP only this wrapper is.
- `server/hip_compat/cuda_runtime.h`: add the three `cuda*` -> `hip*` mappings the kernel needs and the shim did not yet cover (`cudaPointerAttributes`, `cudaPointerGetAttributes`, `cudaMemoryTypeDevice`).
- Rename the guard macro `DFLASH27B_HAVE_DRAFT_TOPK_CUDA` -> backend-neutral `DFLASH27B_HAVE_DRAFT_TOPK`, defined on both backends; update the two call sites (`qwen35_dflash_target.cpp`, `test_dflash.cpp`).
- `server/CMakeLists.txt`: register `test_draft_topk_cuda` on the HIP backend too (GPU-vs-CPU parity test), mirroring the existing HIP `test_flashprefill_kernels` wiring.
- `README.md`: correct the note that both flags are CUDA-only.

## How it works

`DFLASH_GPU_DRAFT_TOPK` now takes effect on HIP: the kernel runs top-K + online-logsumexp directly on the logits device buffer and copies back only the `n_positions x K` results, eliminating the per-step full-vocab D2H and CPU heap extract. Compiling one shared body for both backends keeps the GPU result bit-for-bit identical to the CUDA build and to the CPU reference it is validated against; the automatic CPU fallback on any device failure (out-of-range index, allocation failure, etc.) is unchanged, so the change is output-neutral by construction.

Note on the sibling flag: `DFLASH_GPU_VERIFY_ARGMAX` already reads an in-graph `ggml_argmax` node in `verify_tree` and was never gated on the CUDA-only macro, so it needs no port and already runs on HIP. The README is corrected accordingly.

## Performance

Not yet measured. The change removes a `vocab x n_tokens` (Qwen3.5 vocab ~152k) D2H plus a CPU heap top-K per draft step; the win scales with how bandwidth-bound the card is. A/B recipe on an AMD card: baseline `DFLASH_GPU_DRAFT_TOPK=0`, treatment `DFLASH_GPU_DRAFT_TOPK=1`, `python server/scripts/bench_llm.py --bench HumanEval` (temp 0), comparing decode tok/s and acceptance length.

## Limitations

- The `hipPointerAttribute_t.type` member is spelled `type` on ROCm >= 5.0. All AMD targets in scope (ROCm 6.4.x on the gfx1151 CI runner, 7.x per the discrete-RDNA3 guidance in #457) use `type`; older ROCm that only exposes `.memoryType` is not supported.
- This is the qwen35/DFlash draft path only. It does not touch the laguna tree-verify HIP top_k (#436) or the CUDA fused sampler (#478), which are different code paths.

## Verification

- Static: no stale `DFLASH27B_HAVE_DRAFT_TOPK_CUDA` references remain; guard macro resolves on both backends; new HIP TU and test registration mirror the shipped `flashprefill` HIP pattern.
- **Not build-validated locally this PR:** no GPU toolchain on the authoring machine, and the CUDA validation box was unreachable while preparing this. The CUDA arm change is a macro rename plus a CMake comment, so a clean CUDA rebuild should be a no-op; the HIP arm (compile + the GPU-vs-CPU parity test `test_draft_topk_cuda`, and `DFLASH_GPU_VERIFY_ARGMAX=2` dual-path mismatch mode) needs gfx1100/gfx1151. Requesting the `gpu-tests-amd` runner build (it compiles the tree) or a maintainer proxy to confirm the HIP compile and parity before merge.

<a href="https://cubic.dev/pr/Luce-Org/lucebox-hub/pull/488?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


cheese-cakee:
Rebased onto current `main` (`453427d`) to resolve the conflicts, and validated the HIP path on real AMD hardware.

**On-HW validation (Radeon 8060S, gfx1151, wave32, ROCm 6.4.4):** built `test_draft_topk_cuda` with `LANGUAGE HIP` (arch `gfx1151`, Release) and ran it on the GPU. The `.hip.cu` kernel now actually executes (previously the test skipped for lack of a device), and every case matches the CPU reference:

```
[PASS] n=15 vocab=151936 K=8 temp=1.00  id_mismatch=0 tie_swap=0 max_lp_err=3.166e-04
[PASS] n=1  vocab=151936 K=8 temp=1.00  id_mismatch=0 tie_swap=0 max_lp_err=1.488e-04
[PASS] n=15 vocab=151936 K=8 temp=0.70  id_mismatch=0 tie_swap=0 max_lp_err=9.727e-05
[PASS] n=15 vocab=151936 K=8 temp=2.00  id_mismatch=0 tie_swap=0 max_lp_err=3.815e-05
[PASS] n=7  vocab=1024   K=8 temp=1.00  id_mismatch=0 tie_swap=0 max_lp_err=9.537e-07
[PASS] n=32 vocab=4096   K=8 temp=1.00  id_mismatch=0 tie_swap=0 max_lp_err=1.335e-05
[PASS] n=3  vocab=257    K=8 temp=1.00  id_mismatch=0 tie_swap=0 max_lp_err=9.537e-07
[PASS] n=1  vocab=151936 K=1 temp=1.00  id_mismatch=0 tie_swap=0 max_lp_err=1.659e-04
[PASS] n=15 vocab=151936 K=4 temp=1.00  id_mismatch=0 tie_swap=0 max_lp_err=2.728e-04
[PASS] fallback contract: K=64 (>kMaxK) returned false
ALL PASS: 10/10 cases
```

All `id_mismatch=0` and `max_lp_err <= 3.17e-04` (tolerance `2e-3`), including the Qwen3.5 vocab (151936) K=8 decode shape, and the CPU-fallback contract for K beyond the kernel range holds. The change stays output-neutral by construction (automatic CPU fallback on any failure).

server/src/common/geometric_draft_topk.hip.cu
```diff
@@ -0,0 +1,14 @@
+// HIP translation unit for the DFlash GPU draft top-K + log-prob kernel.
```
howard0su: why we need this file?

howard0su (changes requested):
suggest to cleanup. there are lots of CU files are shared between HIP and CUDA. We should not create hip.cu unless no other approach. This new pattern can confuse people.

server/src/common/geometric_draft_topk.hip.cu
cheese-cakee: Removed the wrapper. The kernel body in `geometric_draft_topk_cuda.cu` is now compiled directly with `LANGUAGE HIP` (the same shared-.cu pattern as `deepseek4_hc_cuda.cu` above), so the separate `.hip.cu` TU is gone. Renamed the guard macro `DFLASH27B_HAVE_DRAFT_TOPK_CUDA` to `DFLASH27B_HAVE_DRAFT_TOPK` so the consumers are backend-neutral, and rebased onto current main. Re-ran the GPU-vs-CPU parity test on gfx1151 (Radeon 8060S, wave32): 10/10 pass.

server/CMakeLists.txt
```diff
@@ -433,8 +446,9 @@ elseif(DFLASH27B_GPU_BACKEND STREQUAL "cuda")
         src/flashprefill.cpp
         src/common/geometric_draft_topk_cuda.cu)
     # PUBLIC so consumers (e.g. the test_dflash executable) also see the macro
-    # and take the GPU draft top-K path instead of the CPU fallback.
-    target_compile_definitions(dflash_common PUBLIC DFLASH27B_HAVE_DRAFT_TOPK_CUDA=1)
+    # and take the GPU draft top-K path instead of the CPU fallback. Same macro
+    # name as the HIP branch above (backend-neutral).
+    target_compile_definitions(dflash_common PUBLIC DFLASH27B_HAVE_DRAFT_TOPK=1)
```
howard0su, line 451: shall we have a separate PR to remove this knob as it was default already? DFLASH27B_HAVE_DRAFT_TOPK

howard0su (approved):

server/CMakeLists.txt
cheese-cakee: in my opinion better as a separate PR so this one stays scoped to the HIP port

PR #353, opened 2026-06-09
feat(dflash): add DeepSeek V4 Flash backend

howard0su:
Implement full DS4 Flash model backend for AR-only decode:

- deepseek4_internal.h: data structures (layer, weights, cache, config)
- deepseek4_loader.cpp: GGUF loader with all DS4 metadata/tensor binding
- deepseek4_graph.cpp: ggml compute graph (MLA attention, KV compression with ratio-4/ratio-128, indexer selective attention, MoE with sqrt(softplus) routing, hash routing, HC residual streams)
- deepseek4_backend.cpp: ModelBackend subclass with hybrid hot/cold expert placement (DFLASH_DS4_HYBRID=1)
- deepseek4_daemon.cpp: daemon entry point

Integration:
- Register 'deepseek4' arch in backend_factory.cpp
- Add to CMakeLists.txt (include path + sources)

Tests:
- test_deepseek4_unit.cpp: CPU-only unit tests with synthetic weights (compressor pooling, MoE routing, RMSNorm, grouped output shape, hash routing lookup)
- deepseek4-vectors/: official API test vectors ported from ds4 project (greedy decode logprob fixtures for integration testing)

PR #489, opened 2026-07-05
feat(dflash): add direct DeepSeek4 pure-UMA runtime path

weicj:
## Summary

This PR brings the current DeepSeek4 backend line from #353  to `main`, and adds a direct `DeepSeek4Backend` runtime path for non-layer-split DeepSeek4 requests in the C++ server.

Before this change, `arch=deepseek4` always went through `DeepSeek4LayerSplitAdapter` + `LayerSplitBackend`, even for ordinary single-backend placements such as pure HIP / UMA. This PR keeps the existing layer-split route when the configured device is actually layer-split, and routes non-layer-split placements to the native DeepSeek4 backend instead.

It also adds decode-side full-step graph reuse and fuller full-step timing telemetry so the direct runtime can be validated and profiled on large UMA devices.

## Changes

- Route non-layer-split `deepseek4` requests to `DeepSeek4Backend`.
- Keep the existing `DeepSeek4LayerSplitAdapter` + `LayerSplitBackend` path for `device.is_layer_split()`.
- Reuse the direct full-step decode graph metadata / allocator state on GPU decode steps.
- Expose full-step `build / alloc / set / compute / read` timings in DeepSeek4 runtime telemetry.
- Leave layer-split placement behavior unchanged.

## Validation

- Runtime validation in a pure UMA `gfx1151` environment with `max_ctx=1048576`
- Large-model startup succeeded with:
  - model buffer: `81687.7 MB`
  - KV cache at `1M ctx`: `6897.8 MB`
- Native `1M ctx` runtime checks with the default `chunk=512`:

| Prompt / Completion | Prefill walltime | Prefill tok/s | Decode walltime | Decode tok/s |
| --- | ---: | ---: | ---: | ---: |
| `256 / 32` | ~~`1228.3 ms`~~ | ~~`208.42`~~| ~~`2001.8 ms`~~ |~~`16.00`~~|
| `4096 / 128` | ~~`18429.1 ms`~~ | ~~`222.26`~~ | ~~`8288.5 ms`~~ | ~~`15.40`~~|
| `8192 / 256` | ~~`42513.7 ms`~~ | ~~`192.69`~~ | ~~`17407.2 ms`~~ | ~~`14.70`~~|
| `16384 / 512` | ~~`108775.9 ms`~~ | ~~`150.62`~~ | ~~`38418.0 ms`~~ | ~~`13.30`~~ |

*above listed data is obselette due to lack of HC-correctness guard*

<a href="https://cubic.dev/pr/Luce-Org/lucebox-hub/pull/489?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -258,7 +258,6 @@ add_library(dflash_common STATIC
     src/deepseek4/deepseek4_daemon.cpp
     src/deepseek4/deepseek4_layer_split_adapter.cpp
     src/deepseek4/deepseek4_target_shard_ipc_daemon.cpp
-    src/deepseek4/deepseek4_expert_ipc_daemon.cpp
     src/flashprefill_q8.cpp
     src/kv_cache.cpp
     src/kv_quant.cpp
@@ -276,7 +275,6 @@ add_library(dflash_common STATIC
     src/common/target_shard_ipc_daemon.cpp
     src/common/dflash_feature_ring.cpp
     src/common/dflash_capture.cpp
-    src/common/expert_ipc.cpp
     src/common/dflash_draft_ipc.cpp
     src/common/dflash_draft_ipc_daemon.cpp
     src/common/pflash_drafter_ipc.cpp
@@ -284,13 +282,6 @@ add_library(dflash_common STATIC
     src/common/dflash_spec_decode.cpp
     src/common/layer_split_backend.cpp
     src/common/layer_split_runtime.cpp
-    src/common/expert_split_plan.cpp
-    src/common/expert_split_state.cpp
-    src/common/expert_split_runtime.cpp
-    src/common/expert_split_compute_runtime.cpp
-    src/common/expert_split_materialization.cpp
-    src/common/expert_split_target_config.cpp
-    src/common/gguf_tensor_data.cpp
     src/qwen35/graph_builders.cpp
     src/qwen35moe/qwen35moe_ffn.cpp
     src/qwen35moe/qwen35moe_backend.cpp
@@ -305,7 +296,6 @@ add_library(dflash_common STATIC
     src/common/moe_hybrid_stream.cpp
     src/common/moe_expert_compute.cpp
     src/common/moe_expert_compute_cpu.cpp
-    src/common/moe_expert_compute_local_gpu.cpp
     src/common/moe_expert_compute_ipc.cpp
     src/common/moe_hybrid_swap_manager.cpp
     src/common/moe_routing_collector.cpp
@@ -324,7 +314,6 @@ add_library(dflash_common STATIC
     src/common/daemon_loop.cpp
     src/common/gguf_inspect.cpp
     src/common/backend_factory.cpp
-    src/common/host_memory_mode.cpp
     src/placement/placement_config.cpp
     src/common/layer_split_utils.cpp
     src/common/ddtree.cpp
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -591,11 +591,6 @@ find_package(OpenMP)
 if(OpenMP_CXX_FOUND)
     target_link_libraries(dflash_common PRIVATE OpenMP::OpenMP_CXX)
 endif()
-if(DFLASH27B_GPU_BACKEND STREQUAL "cuda")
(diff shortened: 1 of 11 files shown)


weicj:
This update stabilizes the PR489 DeepSeek4 direct pure-UMA runtime path.

The previous implementation introduced the direct DeepSeek4 pure-UMA path, but the HC backend graph path was still incomplete and unsafe as a default runtime path. This revision keeps the incomplete batch HC backend graph opt-in only, and uses the verified quality-stable CPU-side/eager HC path by default.

### Changes

- Use GGUF tokenizer metadata for EOS detection instead of hardcoded token ids.
- Make the incomplete batch HC backend graph path opt-in via `DFLASH_DS4_ENABLE_BACKEND_BATCH_HC_GRAPH=1`.
- Keep the quality-stable HC CPU-side/eager runtime path as the default behavior.
- Preserve the performance-oriented HC prefill CPU path, including f16 x f32 matvec, token parallelism, and AVX2/F16C/FMA support.

### Validation

Strict exactseq validation, default runtime, `ctx=1M`, `chunk=512`:

- `bad_files=[]`
- Stable hello256 prefill: `142.5~143.9 tok/s`
- Stable hello256 decode: `12.45~12.6 tok/s`

Replay performance, default runtime, `ctx=1M`, `chunk=512`:

| Case | Prompt Tokens | Completion Tokens | Prefill tok/s | Decode tok/s |
|---|---:|---:|---:|---:|
| 256/32 | 257 | 32 | 130.04 | 11.7 |
| 4k/128 | 4097 | 128 | 146.27 | 10.8 |
| 8k/256 | 8193 | 256 | 90.36 | 10.1 |

davide221:
@weicj can you verify if this can work with the new rocmfpx support? #502

weicj:
Verified with #502 (ROCmFPx). Combined build passes on lucebox5. Perf:

256/32 (ctx1m): prefill=144.36 tok/s, decode=12.1 tok/s
4k/128 (ctx1m): prefill=143.94 tok/s, decode=11.0 tok/s

also cleaned up the `accidental expert_split_target_config` dependency to keep this PR purely on the UMA/HC path (that's from the unfinished dsv4f expert_split route experiement)

weicj:
### 2026-07-10 Update

- Shared decode scalar inputs across cached per-layer decode attention graphs. Per-step values such as RoPE positions, raw KV row, compressor rows, and compressor positions are now uploaded once through compact i32/i64 bundles instead of being set repeatedly for every layer.
- The same shared-input path is wired into both the hybrid decode path and the HC-complete `deepseek4_step_layer_range` path used by the PR #502 compatible runtime flow.
- Request reset now clears compressed-cache counters and HC state so each new request starts from a clean cache topology.
- Hash-routing table handling now preserves the original GGUF table width while consuming only the effective `n_expert_used` prefix, keeping routing logic compatible with default and experimental widths.
- Added UTF-8 boundary filtering in the HTTP output path so byte-fallback tokens split across decode steps are emitted only after they form valid UTF-8. This fixes replacement-character artifacts without changing model compute, routing, EOS handling, or the default HC path.
- Added `DFLASH_DS4_TOPK` as an experimental routing override. This is not used by the default path; the default still follows GGUF `expert_used_count=6`.


## Validation

Quality checks passed on 256/32, 4k/128, and 8k/256 validation runs. All performance numbers below use the default path, `chunk=512`, `ctx1m`, and no TOPK override:

| Case | Prefill | Decode |
| --- | ---: | ---: |
| 256/32 | `178.43 tok/s` | `13.6 tok/s` |
| 4k/128 | `144.13 tok/s` | `11.6 tok/s` |
| 8k/256 | `123.83 tok/s` | `10.9 tok/s` |

`DFLASH_DS4_TOPK=4` was also tested as an experimental mode. It improved throughput to roughly `188.18 / 14.2` on 256/32, `152.22 / 12.2` on 4k/128, and `129.41 / 11.4` on 8k/256, but output quality was not clean enough for the default path due to obvious degradation with repetitive/self-descriptive responses.

PR #492, opened 2026-07-05
feat(server): native tool-prefix caching for multi-turn agents

davidmroth:
## Summary

This revision implements tool-prefix reuse in the production C++
`dflash_server` HTTP path. It replaces the legacy Python tool-split prototype
and its separate `SNAPSHOT_THIN` / `RESTORE_CHAIN` protocol.

The root issue was cache-tier arbitration: a generation can save one snapshot,
and the exact full-prompt cache always took that slot. Agent prompts grow every
turn, so the exact key missed every turn and the reusable inline system/tool
snapshot was never created.

The production behavior is now:

- requests carrying tools prefer the native inline system/tool-boundary
  snapshot, even when the exact prefill cache is enabled;
- ordinary requests retain exact full-prompt cache priority;
- restores use a complete backend-owned snapshot, including attention KV,
  recurrent/convolution state, and last-token state where required;
- the exact token-prefix key includes the rendered tool schema, so changing any
  tool field forces a miss;
- failed, cancelled, invalid, or reused snapshot slots cannot publish stale
  cache metadata;
- `usage.timings` reports `cache_hit`, `cached_prefix_tokens`,
  `prefilled_tokens`, and `effective_prompt_tokens` for direct validation.

No tool-specific flag or alternate server is required. The CUDA and ROCm
container entrypoints now preserve the native HTTP-server default of 32 prefix
slots instead of silently passing `--prefix-cache-slots 0`. Operators can set
`DFLASH_PREFIX_CACHE_SLOTS=N`, including `0` as an explicit opt-out.

## Supported-model benchmark

Hardware and software:

- AMD Ryzen AI MAX+ 395 / Radeon 8060S (`gfx1151`)
- ROCm 7.2.2, target device `hip:0`
- supported Qwen3.6-27B Q4_K_M target, no draft model, greedy AR decode
- `--max-ctx 32768 --prefix-cache-slots 4 --prefill-cache-slots 2`
- 24 tools, 8 parameters per tool, approximately 15K prompt tokens
- three growing warm turns; `max_tokens=4` to isolate prefill

| Model | Cold prefill | Median warm prefill | Cold / warm | Changed-tool control |
|---|---:|---:|---:|---:|
| Qwen3.6-27B Q4_K_M | 50,350.9 ms | **1,041.9 ms** | **48.33x** | 51,176.8 ms, miss |

Warm turns restored 14,848 tokens and prefilled only 218–256 new tokens. All
nine correctness and performance checks passed with identical non-empty
outputs. This is a prefill result, not a claim about decode throughput.

The final container-default patch was also exercised through the production
entrypoint with this supported 27B target. With no cache environment variable,
startup reported `prefix_cache = 32 slots`; explicit `0`, `4`, and exact-prefill
overrides are covered by a committed entrypoint regression test.

## Internal compatibility A/B

Qwen3.5-9B is not an advertised Lucebox target. Its GGUF declares the same
`qwen35` architecture and was used only as a faster internal test vehicle for
an exact `main`/PR A/B of the generic cache plumbing.

| Build | Cold prefill | Median warm prefill | Cold / warm | Warm restore |
|---|---:|---:|---:|---:|
| `main` (`8a16e99e`) | 16,837.6 ms | 17,195.4 ms | 0.98x | none; each turn was cold |
| This PR (`eab5fb6c`) | 16,971.3 ms | **367.1 ms** | **46.23x** | 14,848 tokens |

Changing one tool description produced a full 17,402.2 ms cache miss, proving
that incompatible tool KV was not reused.

## Reproduce

Run the benchmark included in this PR against a production server:

```bash
python3 server/scripts/benchmark_tool_prefix_cache.py \
  --url http://127.0.0.1:8080 \
  --json-out /tmp/tool-prefix-pr.json
```

For an unmodified `main` build, which predates the cache telemetry fields:

```bash
python3 server/scripts/benchmark_tool_prefix_cache.py \
  --url http://127.0.0.1:8080 \
  --timing-only --min-speedup 0.1 \
  --json-out /tmp/tool-prefix-main.json
```

## Validation

- Fresh HIP/ROCm build of `dflash_server`, `test_server_unit`, and
  `test_feature_gate`: pass
- Server unit suite: 2,091 assertions, 0 failures
- Entry-point default/disable/override regression: pass on Linux/ROCm
- Default benchmark on Qwen3.6-27B: all 9 performance, output,
  token-accounting, cache-hit, and mutation checks pass
- Extended five-warm-turn run: pass, including four-slot wraparound and
  preservation of the active restore slot
- `python3 -m py_compile server/scripts/benchmark_tool_prefix_cache.py`: pass
- `bash -n` on the production entrypoint and regression script: pass
- `git diff --check`: pass

The two historical Cubic P1 findings referred to the removed legacy Python
streaming server. The native request lifecycle now owns reservation
confirmation/abort and prompt lifetime in one path.


davide221:
@davidmroth thanks for the huge contribution, we will review it soon. Meanwhile can you fix cubic P1s?

easel:
I was able to get the branch built and the mechanical tests pass, but I cannot reproduce the claimed performance gains.

What I verified:

- `python3 server/scripts/test_prefix_cache_slot_depth.py`: PASS
- `python3 server/scripts/test_tool_split_adapters.py`: PASS
- PR HTTP/multi-turn prefix-cache scripts: PASS mechanically; logs show cache activity, but not the claimed speedup.
- `lucebench` agent + longctx against origin/main vs this PR:
  - origin/main: agent `1/4`, longctx `6/6`, total wall roughly `413.1s`
  - PR #492: agent `1/4`, longctx `6/6`, total wall roughly `412.9s`
  - net result: no meaningful performance delta.
- Agentic replay with Claude CLI judge, 2 cases / 43 turns:
  - origin/main: `2/2` pass, final prefill p50 `117.2s`, largest final prefill `212.2s`
  - PR #492: `2/2` pass, final prefill p50 `111.8s`, largest final prefill `200.6s`
  - this is a small/noisy improvement, not close to the PR’s claimed `~20x` / `~44s -> ~4-5s` improvement.
- On the longer agentic case, logs showed `snap_slot=-1` and `restore=false`, so the run was effectively doing cold growing-prefill work rather than demonstrating warm chain restore.

I also tried to trace where the benchmark claim comes from. The PR body claims:

- `~8.5K-token tool session prefill`: `~11s -> ~0.5s`
- same session wall clock: `~44s -> ~4-5s`
- agent-after-tool incremental turn: `~5s -> ~3.6s`

The closest in-repo source I found is `server/docs/tool-split-goal.md`, which references an external benchmark:

`model-runner-v4/scripts/benchmark-tool-split.py`

That script is not included in this repo, and I could not find it locally. The checked-in test plan validates cache mechanics, but it does not reproduce the advertised benchmark.

So my recommendation is: do not merge this yet as a performance PR. Either:

1. include the benchmark harness / exact commands / environment needed to reproduce the claim, with origin/main vs PR results, or
2. narrow the PR description to say this adds tool-split/prefix-cache plumbing that passes functional tests, but does not yet demonstrate the advertised end-to-end performance gain.

Right now the branch appears to “work” functionally, but the central performance claim is not reproducible from the PR.

davidmroth:
Thank you for the feedback!

The core idea stands strong: _**we separate the huge static tool parameters from the chat portion of the prompt — so PFlash now runs on clean chat turns instead of being skipped for tool calls, plus a speedup from agents never recomputing the tool KV cache**_.

This PR effectively allows agent harnesses to benefit from PFlash, which they cannot today.

I'll share full benchmarks in a day or so.

**Test Environment:**

| Component | Details                  |
|-----------|--------------------------|
| Harness     | Hermes Agent             |
| GPUs      | Dual RTX 3090 (24GB)     |

> I was able to get the branch built and the mechanical tests pass, but I cannot reproduce the claimed performance gains.
> 
> What I verified:
> 
> * `python3 server/scripts/test_prefix_cache_slot_depth.py`: PASS
> * `python3 server/scripts/test_tool_split_adapters.py`: PASS
> * PR HTTP/multi-turn prefix-cache scripts: PASS mechanically; logs show cache activity, but not the claimed speedup.
> * `lucebench` agent + longctx against origin/main vs this PR:
>   
>   * origin/main: agent `1/4`, longctx `6/6`, total wall roughly `413.1s`
>   * PR [feat(server): Anchored Prefix Caching for multi-turn agent workloads #492](https://github.com/Luce-Org/lucebox/pull/492): agent `1/4`, longctx `6/6`, total wall roughly `412.9s`
>   * net result: no meaningful performance delta.
> * Agentic replay with Claude CLI judge, 2 cases / 43 turns:
>   
>   * origin/main: `2/2` pass, final prefill p50 `117.2s`, largest final prefill `212.2s`
>   * PR [feat(server): Anchored Prefix Caching for multi-turn agent workloads #492](https://github.com/Luce-Org/lucebox/pull/492): `2/2` pass, final prefill p50 `111.8s`, largest final prefill `200.6s`
>   * this is a small/noisy improvement, not close to the PR’s claimed `~20x` / `~44s -> ~4-5s` improvement.
> * On the longer agentic case, logs showed `snap_slot=-1` and `restore=false`, so the run was effectively doing cold growing-prefill work rather than demonstrating warm chain restore.
> 
> I also tried to trace where the benchmark claim comes from. The PR body claims:
> 
> * `~8.5K-token tool session prefill`: `~11s -> ~0.5s`
> * same session wall clock: `~44s -> ~4-5s`
> * agent-after-tool incremental turn: `~5s -> ~3.6s`
> 
> The closest in-repo source I found is `server/docs/tool-split-goal.md`, which references an external benchmark:
> 
> `model-runner-v4/scripts/benchmark-tool-split.py`
> 
> That script is not included in this repo, and I could not find it locally. The checked-in test plan validates cache mechanics, but it does not reproduce the advertised benchmark.
> 
> So my recommendation is: do not merge this yet as a performance PR. Either:
> 
> 1. include the benchmark harness / exact commands / environment needed to reproduce the claim, with origin/main vs PR results, or
> 2. narrow the PR description to say this adds tool-split/prefix-cache plumbing that passes functional tests, but does not yet demonstrate the advertised end-to-end performance gain.
> 
> Right now the branch appears to “work” functionally, but the central performance claim is not reproducible from the PR.

davide221:
@easel I addressed the reproducibility blocker by replacing the non-shipping legacy Python path with the native C++ server path and adding a self-contained benchmark. On the same ROCm host/config, current main remained cold at 17,195.4 ms median warm prefill; this PR restored 14,848 tokens and measured 366.2 ms median warm prefill (47.0x vs main). The benchmark also checks identical outputs, exact token accounting, repeated hits, and a changed-tool forced miss. Exact environment, commands, and scoped results are now in the PR body.

davide221:
Addressed all three new Cubic findings in eab5fb6c: non-positive prefill timings now invalidate the run; output stability compares only identical-tool cold/warm requests; and usage.timings now exposes effective_prompt_tokens so accounting remains valid after FlowKV/PFlash prompt rewriting. Rebuilt on ROCm, 2,089 assertions pass, and the updated benchmark passes all nine gates.

PR #501, opened 2026-07-08
vendor: grouped-expert MUL_MAT_ID for small MoE verify batches (lagunaxs2.1 porting)

davide221:
## What

Patches the vendored ggml (`server/deps/llama.cpp`).

Opt-in grouped-expert path for `MUL_MAT_ID` on small token batches (2-16), targeting speculative-verify workloads on quantized MoE models. Enable with `DFLASH_MMID_GROUPED=1`.

## Why

Consecutive draft tokens route to heavily overlapping expert sets, but `mul_mat_vec_q_moe` reads each (token, slot) pair's expert weights independently. Measured on a 256-expert top-8 MoE (Laguna XS 2.1):

| verify rows | expert reads today | distinct experts (union) | duplication |
|---|---|---|---|
| 8 | 64 | 39.4 | 1.62x |
| 15 | 120 | 58.9 | 2.04x |

## How

- A tiny single-block prep kernel rank-sorts the batch's (token, slot) pairs by routed expert (`[TAG_MMID_GROUPED]` in `mmvq.cu`). Capture-safe: ids never leave the device, all launch shapes are static.
- The main kernel is `mul_mat_vec_q_moe` with one change: warp `w` processes sorted pair `blockIdx.y*8 + w` instead of `(slot = blockIdx.y, token = w)`. Same-expert pairs are adjacent after sorting, so the warps of a block share an expert and duplicate weight reads hit L1/L2 instead of DRAM.
- Launch shape, per-warp structure, `vec_dot` sequence and warp reduction are identical to `mul_mat_vec_q_moe` → **bit-exact outputs**, CUDA-graph capture unaffected.
- `DFLASH_MMID_GROUPED_TYPES` bitmask: `1` = Q4_K (default), `2` = Q6_K (also lifts Q6_K's MUL_MAT_ID MMVQ ceiling to 16, keeping CUDA graphs on for those batches), `4` = Q4_0/Q8_0/Q5_K.
- Model-agnostic: any MoE with `n_expert_used <= 16` (asserted).

## Results

RTX 3090 (sm_86), Laguna XS 2.1 Q4_K_M + DFlash drafter, greedy HumanEval prompts. Outputs byte-identical to baseline in all cases (verified per-prompt, identical accept trajectories):

| config | baseline tok/s | grouped tok/s | gain |
|---|---|---|---|
| verify width 8 | 212.0 | 219.9 | +3.7% |
| verify width 15 | ~157 | 173.3 | ~+10% |

Gains grow with verify width (more overlap to dedup), which lowers the marginal cost of wide/hedged verify configurations.

## Notes

- Off by default; zero impact when the env is unset (single branch check in `ggml_cuda_mul_mat_vec_q`).
- Prep adds one ~µs kernel launch per MUL_MAT_ID call; included in all numbers above.
- Kernel design history: a register-blocked variant (tokens accumulated in registers per warp) lost 28% to spills; a warp-per-group variant lost 7% to occupancy collapse from early-exiting warps. The shipped design keeps the tuned kernel's exact occupancy profile and takes the dedup purely from cache locality.

---

## Update: opt-in per-token adaptive expert count (second commit)

On top of the (bit-exact, default-off) grouped kernel, `DFLASH_ADAPTIVE_K_TAU=<0..1>` enables per-token expert-count gating inside the grouped prep kernel: each verify token keeps its leading experts until cumulative router mass ≥ τ, the rest are sentineled to `-1` (skipped, exact zero contribution) and kept weights are renormalized in place. Near-lossless, not bit-exact — off by default.

- Builder side: `server/src/common/mmid_adaptive_k.h` tags the router ids tensor with the combine-weights tensor via `ids->extra`. Wired for **laguna** (DFlash capture layers `1,13,25,33,39` dense by default), **qwen35moe** and **gemma4** (`DFLASH_ADAPTIVE_K_DENSE` for exclusions). deepseek4 (DSpark routing) not wired.
- Prefill and 1-token decode untouched; gating applies only to 2-16-token batches on the grouped path (`DFLASH_MMID_GROUPED_TYPES=7` recommended so all routed quant types participate).

Measured (RTX 3090, Laguna XS 2.1 Q4_K_M + DFlash, HumanEval-10 greedy):

| config | tok/s | expected_pass |
|---|---|---|
| baseline | 212.0 | 10/10 |
| grouped only (bit-exact) | ~220 | 10/10 |
| grouped + τ=0.80 | **224.5** | **10/10** |

τ sweep: 0.90 ≈ no gain, **0.80 best**, 0.70 starts costing accept length. τ is a per-model constant: laguna is validated, qwen35moe/gemma4 carry the mechanism but need their own 30-minute sweep before enabling.

**Width note:** with both optimizations on, the verify-width optimum shifts from 8 to **6** (HE-10: τ=0.80 w6 = 226.9 tok/s vs w8 = 224.5; true pre-update best was w5 = 217.1). The width curve is a flat plateau across w5-w8.

---

## Update: drafter-confidence adaptive verify width (third commit)

The width plateau above exists because a fixed verify width averages two regimes: on confident steps deep slots commit often (w8 worth paying), on unsure steps they almost never do (w4 is enough). Every MoE verify row routes to its own top-k experts (~1ms/row of expert weight reads for a Q4 target on a 3090), so rows are the dominant marginal cost. `DFLASH_ADAPTIVE_WIDTH_THETA=<0..1>` trims the batch per step: keep candidate slot j while the drafter's own reach-mass (product of top-1 probabilities through slot j) stays above θ, floored at `DFLASH_ADAPTIVE_WIDTH_MIN` (default 4) rows. Output-exactness is preserved — only the number of speculated slots changes; every committed token is still target-verified.

- `ggml-cuda`: new `ggml_backend_cuda_topk_rows` (`topk-rows.cu`) — top-k (k ≤ 8) + softmax probabilities per row of the draft-head logits, ~KB readback instead of the full-vocab logits round-trip. Also accelerates `project_hidden_to_topk` (draft-tree candidates) at unit temperature.
- `common`: `adaptive_verify_width.h` (model-agnostic helper — any family loop with per-slot drafter top-1 probs can call it) + a `DFlashTarget::project_hidden_to_tokens_topk` hook that returns candidate probs alongside draft tokens; the fused domino head exposes its corrected logits for the same extraction.
- `laguna` (reference integration): candidate extraction rides the existing draft-token projection (no extra matmul), per-width verify-graph slots so each row count's CUDA graph captures once and replays, trim applied right before `verify_batch`.

Measured (RTX 3090, Laguna XS 2.1 Q4_K_M + official DFlash drafter, HumanEval-10 greedy, 512 tok, same-sweep control):

| config | tok/s | commit/step | avg rows | expected_pass |
|---|---|---|---|---|
| grouped + τ=0.80, fixed w6 (prev best) | 226.4 | 3.58 | 6.00 | 10/10 |
| + θ=0.20, min 4, `--verify-width 8` | **236.0** | **3.79** | **5.35** | **10/10** |

θ sweep: 0.15/0.25/0.30 and base widths 10 all worse than θ=0.20 w8; min 5 hurts. Mechanism confirmed by the row stats: confident steps stretch to 8 rows (commit 3.58 → 3.79) while the average row count lands *below* the fixed-w6 control. Cumulative PR gain vs the true pre-PR baseline (w5, 217.1): **+8.7%**.

Off by default; θ unset = existing fixed-width behavior. qwen35moe/gemma4 carry the helper + hook but need their loop wiring and a θ sweep before enabling.

---

## Update: rebased onto main @ 5e302cb + two more commits

**`fix(draft)` — Laguna XS 2.1 drafter support (4a4d519).** Runtime validation on a main build (all previous numbers were measured on a lucebox working tree) exposed that main's draft loader ignores the XS 2.1 drafter's attention gates, aux hidden norms and context-KV input norms — that support only existed as uncommitted edits on the lucebox checkout. Without the feature norms the fc projection overflows in f16 and every draft hidden state is NaN, so **spec decode was broken on any main build** (`laguna_verify_batch: embed failed` from an all-NaN argmax). This commit ports the loader + draft-graph support; main can now run the official drafter.

**`server:` defaults consolidation (1238629).** Collapses the env sprawl: the two exact optimizations are now on by default (grouped MUL_MAT_ID on CUDA, adaptive verify width θ=0.20 with base width 8 for greedy chains), envs demoted to kill-switches/debug overrides. The one output-changing knob is a CLI flag: `--adaptive-experts [tau]` (default 0.80). A stock server invocation gets the validated exact stack with zero configuration.

Measured on the rebased branch, main build, RTX 3090, HE-10 512-tok, 10/10 pass both:

| config | tok/s (harness mean) | commit/step | avg rows |
|---|---|---|---|
| default (no flags, no envs) | 224.6 | 3.74 | 5.32 |
| `--adaptive-experts` | 228.3 | — | — |

Warm per-request decode with `--adaptive-experts`: 248-258 tok/s (server-side). Numbers are a hot-GPU session on the .105 box (known VRAM-throttle drift); same-protocol working-tree measurements earlier the same day: fixed-w6 control 226.4 → full stack 236.0.


server/deps/llama.cpp/ggml/src/ggml-cuda/mmvq.cu
```diff
@@ -233,8 +233,67 @@ static constexpr __host__ __device__ int get_mmvq_mmid_max_batch_rdna4(ggml_type
...
+        return e && e[0] == '1' && e[1] == '\0';
+    }();
+    return on;
+}
+
+static bool mmid_grouped_type_ok(ggml_type type) {
+    // bit0 = Q4_K, bit1 = Q6_K, bit2 = Q4_0/Q8_0/Q5_K. Default Q4_K only:
+    // Q6_K keeps its tuned MMQ route above 5 tokens unless explicitly enabled.
+    static const int mask = []() {
+        const char * e = std::getenv("DFLASH_MMID_GROUPED_TYPES");
```
howard0su, line 293: I suggest we don't keep experiement env if we believe this doesn't bring regression. too many env is so hard to maintain and keep track on them.

howard0su (approved):

PR #503, opened 2026-07-08
feat(ds4): add fused HIP decode for ROCmFPX

davide221:
## Summary

Add the production DeepSeek V4 Flash decode path on top of the ROCmFPX types
merged in #502:

- cached, monolithic HIP decode graph covering HC, attention, MoE, and logits;
- gfx1151 single-column ROCmFP2/FP3 fixed-K kernels and ROCmFP4 loop
  specialization, restricted to the exact profiled shapes;
- explicit `--ds4-fused-decode` and `--ds4-expert-top-k N` serving options;
- exact tokenwise prefill and cache-reset/ring-wrap correctness fixes;
- race-free persistent HC worker pool for the layered/hybrid fallback.

The branch is rebuilt directly on current `main`; it no longer carries the
already-merged #502 changes or the old submodule-era conflicts.

## Runtime behavior

`--ds4-fused-decode` now requests a monolithic HIP model load because the fused
graph must reference every expert tensor directly. If that allocation fails,
the backend logs the fallback and disables fused execution for hybrid expert
placement instead of claiming that the fused graph is active.

`--ds4-expert-top-k 4` keeps and renormalizes four routed experts. It is an
explicit approximate-inference policy; `0` retains the model's configured
expert count.

No performance feature in this PR requires a serving-time environment
variable. `DFLASH_DS4_TIMING` remains optional instrumentation only.

## Correctness and maintainability

- New sequences reset both cache storage and compressed-row topology.
- Decode after raw-SWA ring wrap reads the correct physical rows; unit coverage
  exercises pre-wrap, boundary, and wrapped cases.
- Compressed-KV writes are graph dependencies, and cached graph keys include
  the flush topology.
- Fused timing is included in phase-level telemetry.
- Default MoE reduction keeps the proven `permute + cont + sum_rows` path.
- No implicit system prompt or unsupported chunked-prefill mode is introduced.
- Dead full-step graph-cache code was removed.
- The HC worker sequence and stop predicates are synchronized with the same
  mutex used by the condition-variable wait, preventing lost wakeups.

## Validation

Hardware and model:

- Strix Halo `gfx1151`, ROCm 7.2.2, 128 GiB RAM / 96 GiB ROCm aperture;
- performance profile with 2900 MHz SCLK observed throughout the run;
- `DeepSeek-V4-Flash-ROCMFP2-STRIX.gguf`, SHA-256
  `8fa6c30d9badd8e72f83c62952fe24b2e8cd647e103aff086bece328cb9c9208`.

Configuration:

```text
--target-device hip:0
--max-ctx 8192
--prefix-cache-slots 0
--prefill-cache-slots 0
--ds4-fused-decode
--ds4-expert-top-k 4
```

Two consecutive GSM+Math runs produced identical generated text and scores:

| Suite | Accuracy per run | Mean decode | Output tokens |
|---|---:|---:|---:|
| GSM | 5/5 | 26.39 tok/s | 632 |
| Math | 5/5 | 25.83 tok/s | 963 |
| Combined | 10/10 | 26.11 tok/s unweighted; 25.89 token-weighted | 1,595 |

The same checked prompt measured 16.10 tok/s when the old 8K initialization
silently selected hybrid/layered execution and 26.56 tok/s after the monolithic
fused path was made reachable.

The strict trace gate compared 64 full-vocabulary F32 logits rows byte for byte
with zero mismatches. The final model-loading fix changes path selection only;
it does not change graph arithmetic. A fresh ROCm build passes all 20 CTest
targets, including ROCmFP reference/tail tests, RMSNorm, HC, DeepSeek4, and
server tests.

Top-k 4 still requires workload-level quality validation beyond this small
regression suite.

### DIFF
--- a/server/src/deepseek4/deepseek4_backend.cpp
+++ b/server/src/deepseek4/deepseek4_backend.cpp
@@ -69,6 +69,10 @@ static void add_step_tel(DeepSeek4StepTelemetry & dst, const DeepSeek4StepTeleme
     dst.output_us += src.output_us;
     dst.sample_us += src.sample_us;
     dst.emit_us += src.emit_us;
+    dst.full_graph_build_us += src.full_graph_build_us;
+    dst.full_graph_set_us += src.full_graph_set_us;
+    dst.full_graph_compute_us += src.full_graph_compute_us;
+    dst.full_graph_read_us += src.full_graph_read_us;
     dst.hot_selected += src.hot_selected;
     dst.cold_selected += src.cold_selected;
 }
@@ -92,6 +96,7 @@ static void log_step_tel(const char * phase,
         "ffn_hot_graph_build=%llu ffn_hot_graph_hit=%llu ffn_cold_graph_build=%llu ffn_cold_graph_hit=%llu "
         "hc_pre=%.1fms hc_pre_build=%.1fms hc_pre_input=%.1fms hc_pre_compute=%.1fms "
         "hc_post=%.1fms output=%.1fms sample=%.1fms emit=%.1fms "
+        "full_build=%.1fms full_set=%.1fms full_compute=%.1fms full_read=%.1fms "
         "hot_sel=%d cold_sel=%d\n",
         phase, tokens, steps, wall_s, tok_s,
         ms(t.total_us), ms(t.embed_us), ms(t.attn_build_us), ms(t.attn_compute_us), ms(t.attn_read_us),
@@ -107,6 +112,8 @@ static void log_step_tel(const char * phase,
         ms(t.hc_pre_compute_us),
         ms(t.hc_post_attn_us + t.hc_post_ffn_us),
         ms(t.output_us), ms(t.sample_us), ms(t.emit_us),
+        ms(t.full_graph_build_us), ms(t.full_graph_set_us), ms(t.full_graph_compute_us),
+        ms(t.full_graph_read_us),
         t.hot_selected, t.cold_selected);
 }
 
@@ -342,6 +349,15 @@ bool DeepSeek4Backend::init() {
         }
     }
 
+    if (cfg_.expert_top_k < 0 || cfg_.expert_top_k > w_.n_expert_used) {
+        std::fprintf(stderr,
+                     "[deepseek4] expert top-k must be in [0,%d], got %d\n",
+                     w_.n_expert_used, cfg_.expert_top_k);
+        return false;
+    }
+    w_.routed_expert_top_k = cfg_.expert_top_k;
+    w_.fused_decode = cfg_.fused_decode;
+
     const int max_ctx = cfg_.max_ctx > 0 ? cfg_.max_ctx : 8192;
     if (!create_deepseek4_cache(backend_, w_, max_ctx, cache_)) {
         std::fprintf(stderr, "[deepseek4] failed to allocate KV cache (ctx=%d)\n", max_ctx);
@@ -353,8 +369,13 @@ bool DeepSeek4Backend::init() {
         // The DeepSeek4Backend single-GPU path now runs all experts locally.
     }
 
-    std::fprintf(stderr, "[deepseek4] initialized: %d layers, ctx=%d, %d experts (%d used)%s\n",
-                 w_.n_layer, max_ctx, w_.n_expert, w_.n_expert_used,
+    const int active_experts =
+        w_.routed_expert_top_k > 0 ? w_.routed_expert_top_k : w_.n_expert_used;
+    std::fprintf(stderr,
+                 "[deepseek4] initialized: %d layers, ctx=%d, %d experts "
+                 "(%d/%d routed), fused_decode=%s%s\n",
+                 w_.n_layer, max_ctx, w_.n_expert, active_experts, w_.n_expert_used,
+                 w_.fused_decode ? "on" : "off",
                  moe_hybrid_ ? " [hybrid]" : "");
     return true;
 }
@@ -533,11 +554,20 @@ bool DeepSeek4Backend::unpark(const std::string & what) {
 int DeepSeek4Backend::do_prefill(const std::vector<int32_t> & tokens,
                                   const DaemonIO & io,
                                   int kv_offset) {
-    // Hybrid currently implements HC for single-token steps only; keep prefill
-    // token-by-token so the first sampled token is seeded from the correct HC state.
-    const int chunk = moe_hybrid_ ? 1 : (cfg_.chunk > 0 ? cfg_.chunk : 512);
+    // The current batched graph updates only the chunk's final compressor
+    // window. Until batched compressor state is sequentially updated, process
+    // prefill one token at a time to preserve long-prompt correctness.
+    constexpr int chunk = 1;
     const int n_total = (int)tokens.size();
     int pos = kv_offset;
+    // New sequence: clear the cache buffer so compressor state double-buffers
+    // and compressed-KV rows start from zeros, exactly like a fresh server.
+    // Without this, the first flush windows of a request pool over the
+    // previous request's leftover state rows and outputs from the 2nd/3rd
+    // request on can drift by a token or two.
+    if (kv_offset == 0) {
+        reset_deepseek4_cache(cache_);
+    }
     last_logits_.clear();
     const bool timing = env_flag_enabled("DFLASH_DS4_TIMING");
     const auto phase_t0 = Clock::now();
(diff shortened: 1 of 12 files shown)


davide221:
Validation update (2026-07-09): validated on lucebox2 against DeepSeek-V4-Flash ROCmFP2. The env-only expert routing knob DFLASH_DS4_TOPK=4 reduces routed experts from 6 to 4 and renormalizes retained weights. AR validation (GSM n=5 plus Math n=5) reached 26.19 tok/s GSM and 25.16 tok/s Math, both 5/5, versus baseline 23.54 and 22.24 tok/s. DSpark fused speculative decode reached 20.1–20.3 tok/s warm in repeated q=4 runs, with approximately 105–106 ms verify steps. Adaptive width, causal batched verify, fused verify, and interior-boundary compressor emission are wired and producing coherent output. The neuron-magnitude mask preserved the small-screen scores but produced no speedup; the block-sparse follow-up was not quality-safe and is kept out of the validated path. These measurements come from the hub-494 experiment worktree; the final k=4 patch and larger DSpark runtime follow-on still require explicit commit/push if they are to become part of the PR diff.

server/src/deepseek4/deepseek4_graph.cpp
```diff
@@ -1725,20 +2091,26 @@ static ggml_tensor * build_moe_ffn(
...
-        routed_out = ggml_cont(ctx, ggml_permute(ctx, routed_out, 1, 0, 2, 3));
-        routed_out = ggml_sum_rows(ctx, routed_out);
-        routed_out = ggml_reshape_2d(ctx, routed_out, n_embd, n_tokens);
+        if (ds4_ffn_fused_combine_enabled()) {
+            routed_out = ggml_laguna_moe_combine(ctx, down_e, routing.weights);
+        } else {
+            ggml_tensor * weights_3d = ggml_reshape_3d(ctx, routing.weights, 1, n_used, n_tokens);
+            routed_out = ggml_mul(ctx, down_e, weights_3d);
+            ggml_tensor * sum_shape = ggml_new_tensor_3d(ctx, GGML_TYPE_F32, n_embd, 1, n_tokens);
+            routed_out = ggml_repeat_back(ctx, routed_out, sum_shape);
```
DeanoC: **Default-path reduction changed: `sum_rows` → `repeat_back`.** This fires whenever `FFN_FUSED_COMBINE` is off, i.e. the default, yet it isn't in the "byte-identical default" list. Two things to confirm before merge:

1. **Backend support** — `GGML_OP_REPEAT_BACK` (f32) must be implemented on the vendored ROCmFPX/HIP backend. The old `cont(permute)+sum_rows` path was known-good; if `repeat_back` isn't wired up here, default decode fails at compute time. Given the PR's premise is ROCmFPX byte-safety, this is the top risk.
2. **Bit-identity** — both sum the same `n_used` values per `(n_embd, n_tokens)` output, so it's mathematically equal, but kernel summation order/rounding can differ. If you validated logit/argmax equality vs the #494 tip, please note it.

Same change also at the decode variant in `build_cached_decode_ffn_graph` (~line 3739).

server/src/server/chat_template.cpp
```diff
@@ -327,6 +327,11 @@ std::string render_chat_template(
             system_content += msg.content;
             has_system = true;
         }
+        if (!has_system) {
+            system_content =
```
DeanoC: Default system prompt is correctly emitted (via `result += system_content` below, regardless of `has_system`), but two small things:

- **Missing separator with tools:** when `has_tools && !has_system`, the `if (has_system) result += "\n\n";` above is skipped, so `tools_json` is concatenated directly onto this default prompt. Setting `has_system = true` inside this block fixes the separator and is more honest (there now *is* system content).
- **Behavior change:** every DeepSeek4 request without a system message now silently gets this prompt. That's product-visible for API callers who deliberately send none. Please confirm it matches the ds4 reference server's default.

server/src/deepseek4/deepseek4_graph.cpp
```diff
@@ -2678,23 +3360,71 @@ bool deepseek4_step(
 
     const int n_embd = w.n_embd;
     const int n_layer = w.n_layer;
+    const size_t ctx_size = ds4_full_step_meta_size(n_tokens);
+    const bool reuse_full_step_decode =
```
DeanoC: **Possibly-dead code.** `do_prefill`/`do_decode` now route the non-hybrid case to `deepseek4_step_layer_range`, and `deepseek4_step` early-returns to `deepseek4_step_hybrid` for the hybrid case. That leaves this full-step decode-graph reuse machinery (`reuse_full_step_decode` / `full_step_cache`) reachable only if `deepseek4_step` is still called with a non-hybrid model somewhere (tests/bench/tools). If it isn't, this is a fair amount of new, unexercised complexity on a path the PR description itself calls an HC-less stub. Please confirm the caller or drop it.

server/src/deepseek4/deepseek4_graph.cpp
```diff
@@ -950,45 +1147,80 @@ static ggml_tensor * build_mla_attention(
...
-            ggml_tensor * prev_f32 = ggml_cast(ctx,
-                ggml_view_2d(ctx, lc.raw_kv, head_dim, n_raw - 1,
-                             lc.raw_kv->nb[1], 0),
-                GGML_TYPE_F32);
-            kv_f32 = ggml_concat(ctx, prev_f32, cur_kv_f32, 1);
+            ggml_tensor * prev = nullptr;
+            if (kv_start < w.n_swa) {
+                prev = raw_kv_view(0, n_raw - 1);
+            } else {
+                const int raw_row = kv_start % w.n_swa;
```
DeanoC: Nice — this is a genuine latent-bug fix, not just a refactor. The old single-token decode read the first `n_raw-1` *physical* ring rows from offset 0, which was wrong once `kv_start >= n_swa` (included a stale row, dropped the newest). The tail+head split reads every row except the current slot; since key attention is permutation-invariant, membership is what matters, so this is correct. Worth a regression test that decodes past `kv_start > n_swa`.

DeanoC (commented):
Semi-Automated review pass. 
Codex: Overall this is careful, well-documented work and the risky paths (fused decode, FFN combine variants, GPU HC) are correctly gated behind default-OFF env flags. My comments concentrate on a few changes that land on the *default* code path plus one likely-dead block. Inline notes below;

Added a few from Codex, all just nits really. Works well, i've been tested it over the last few days and is works well!
Nice work davide!

server/src/deepseek4/deepseek4_backend.cpp
```diff
@@ -535,9 +535,26 @@ int DeepSeek4Backend::do_prefill(const std::vector<int32_t> & tokens,
...
+    const int chunk = (moe_hybrid_ || !unsafe_chunked) ? 1 : (cfg_.chunk > 0 ? cfg_.chunk : 512);
     const int n_total = (int)tokens.size();
     int pos = kv_offset;
+    // New sequence: clear the cache buffer so compressor state double-buffers
+    // and compressed-KV rows start from zeros, exactly like a fresh server.
+    // Without this, the first flush windows of a request pool over the
+    // previous request's leftover state rows and outputs from the 2nd/3rd
+    // request on can drift by a token or two.
+    if (kv_offset == 0 && cache_.buf) {
+        ggml_backend_buffer_clear(cache_.buf, 0);
```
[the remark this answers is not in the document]
davide221: Valid. The new-sequence path now calls reset_deepseek4_cache(), which clears the buffer, cur_pos, n_comp, and n_index_comp together. test_reset_deepseek4_cache covers the storage and topology reset.

server/src/deepseek4/deepseek4_internal.h
```diff
@@ -64,6 +64,11 @@ struct DeepSeek4StepTelemetry {
     uint64_t output_us = 0;
     uint64_t sample_us = 0;
     uint64_t emit_us = 0;
+    uint64_t full_graph_build_us = 0;
```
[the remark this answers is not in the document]
davide221: Valid. add_step_tel() now aggregates the fused build/set/compute/read counters and log_step_tel() emits them. The box4 validation log confirms non-zero full_compute timing for the fused path.

server/src/deepseek4/deepseek4_graph.cpp
davide221: Agreed. The cleaned branch restores the proven permute + cont + sum_rows reduction in both the generic and cached decode paths. The strict gate compared 64 full-vocabulary F32 logits rows with zero byte mismatches.

server/src/server/chat_template.cpp
davide221: Removed from the cleaned branch. Requests without a system message keep the existing API behavior; this PR no longer injects an implicit DeepSeek4 system prompt.

server/src/deepseek4/deepseek4_graph.cpp
davide221: Confirmed and removed. The dead full-step cache is gone; monolithic fused decode is implemented in the HC-complete layer-range path that the production backend actually calls.

server/src/deepseek4/deepseek4_graph.cpp
davide221: Added. deepseek4_previous_raw_ring_spans() is covered before wrap, exactly at the boundary, and after wrap (kv_start 3, 8, and 10 with n_swa 8). The fresh ROCm CTest run passes all 20 targets.

PR #525, opened 2026-07-14
fix(server): Parse Bare Tool-Name XML Calls

jkyamog:
Adds a narrow compatibility case for model output that uses the tool name as
the XML tag itself:

```xml
<bash>
<parameter=command>
grep -n "f5.test" /workspace/project/tests/bootstrap.cjs
</parameter>
</function>
```

Before this change, the C++ server did not parse this as a tool call. The raw
XML stayed in visible assistant text and the turn finished as normal text
instead of `tool_calls` / `tool_use`.

## Bug

Observed through a Pi agent session using Lucebox as the model runtime.

The session record showed reasoning was already correctly separated:

```json
{
  "type": "thinking",
  "thinkingSignature": "reasoning_content"
}
```

But the malformed tool call was left in the visible text content:

```xml
<bash>
<parameter=command>
grep -n "f5.test" /workspace/project/tests/bootstrap.cjs
</parameter>
</function>
```

The agent therefore displayed the tool XML as assistant text and did not run
the tool.

## Change

- Adds parser support for bare tool-name XML:

  ```xml
  <TOOL_NAME>
  <parameter=KEY>VALUE</parameter>
  </function>
  ```

- Also accepts a matching generic XML close:

  ```xml
  </TOOL_NAME>
  ```

- Keeps the behavior constrained:
  - only runs when request `tools` are present
  - only accepts names in the declared tool list
  - requires at least one `<parameter=...>` tag
  - reuses the existing XML parameter parser and argument JSON serialization

This avoids turning arbitrary XML-ish prose into tool calls.

Review follow-up:

- The matching generic close is implemented as a backreference (`</TOOL_NAME>`)
  rather than "any XML close" so the regex cannot stop early on inner tags like
  `</parameter>`.
- The fixture asserts that `</function>` is fully stripped from visible content.

<a href="https://cubic.dev/pr/Luce-Org/lucebox/pull/525?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/src/server/tool_parser.cpp
+++ b/server/src/server/tool_parser.cpp
@@ -8,6 +8,7 @@
 // 5. call:<ns>?<verb>{relaxed-JSON args}    (gemma plain-text emissions)
 // 6. Bare JSON objects with name+arguments fields
 // 7. Whole-response JSON args for exactly one declared tool
+// 8. <TOOL_NAME>...<parameter=K>V</parameter>...</function>  (bare tool tag)
 //
 // Pattern 5 runs *before* pattern 6 so that args like
 //   call:outer{"name": "inner", "arguments": {}}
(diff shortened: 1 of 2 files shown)


jkyamog:
I encountered this issue twice on long running Pi agent session, so I decided to fix this as it occurs frequent enough.  It was using Qwen 3.6 27B Q4 model.  I have then run the fix with 10M-20M more tokens and so far no tool call issues.

davide221:
Thanks for the fix @jkyamog 🚀

PR #526, opened 2026-07-15
fix(ds4): use monolithic backend for single-device launches, not only for HIP

Graffioh:
## Summary

- route DeepSeek 4 single-device launches through `DeepSeek4Backend` on **both** CUDA and HIP
- reserve `LayerSplitBackend` for explicit multi-device placement or a configured remote target shard
- preserve automatic CUDA/Halo layer splitting whenever remote target-shard IPC is enabled

## Why

CUDA launches previously always selected `DeepSeek4LayerSplitAdapter`. Without an explicit or remote split, its automatic sizing capped the local shard at 42 layers, so a single H200 executed a virtual `[0, 42) + [42, 43)` split on the same GPU.

That virtual boundary adds split orchestration overhead and prevents the full-model fused decode path, whose eligibility requires the same range to start at layer 0 and own the output layer. A single local device should instead use the monolithic backend; devices that cannot fit the full model can use its existing hybrid-expert fallback.

## Expected behavior

| Configuration | Backend | Behavior |
|---|---|---|
| H200/Halo single | `DeepSeek4Backend` | Full monolithic |
| 3090 single | `DeepSeek4Backend` | Full attempt → hybrid fallback |
| 3090 + Halo IPC | `LayerSplitBackend` | Auto-split CUDA/Halo |
| Multi-GPU explicit | `LayerSplitBackend` | Layer split |

## Validation

- `git diff --check`
- Run on H200 both single and virtual split (based on https://github.com/Luce-Org/lucebox/pull/507)

### DIFF
--- a/server/src/common/backend_factory.cpp
+++ b/server/src/common/backend_factory.cpp
@@ -238,11 +238,10 @@ std::unique_ptr<ModelBackend> create_backend(const BackendArgs & args) {
             return nullptr;
         }
 
-        // HIP single-device launches cannot rely on the CUDA/Halo auto-split
-        // path; use the single-backend loader, which can fall back to hybrid
-        // expert placement when a full monolithic load does not fit.
-        if (target_backend == PlacementBackend::Hip &&
-            !args.device.is_layer_split() &&
+        // A single local device uses the monolithic backend. Reserve the
+        // layer-split adapter for explicit multi-device placement or remote
+        // target shards.
+        if (!args.device.is_layer_split() &&
             !args.remote_target_shard.enabled()) {
             DeepSeek4BackendConfig cfg;
             cfg.model_path = args.model_path;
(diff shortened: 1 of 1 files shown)


davide221:
Rebased onto current main ae98774 and resolved the backend_factory.cpp conflict at head 6e5b149. The resolution preserves current DS4 prefill validation/config propagation and applies the intended routing rule: single local targets use DeepSeek4Backend; explicit or remote splits use LayerSplitBackend.

Lucebox4 validation (exact pushed head 6e5b149):

- git diff --check: PASS
- CUDA 13.3 / sm_86 Release: dflash_server and test_deepseek4_unit built successfully
- CUDA deepseek4_unit: PASS
- ROCm 7.2.2 / gfx1151 Release: dflash_server and test_deepseek4_unit built successfully
- HIP deepseek4_unit: PASS
- Model-backed RTX 3090 generation using DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix.gguf: PASS
  - DeepSeek4Backend selected (no virtual layer split)
  - expected 81.7 GiB full CUDA allocation failure cleanly fell back to hybrid placement
  - hybrid initialized with 2322 hot / 8686 cold experts
  - real chat request returned HTTP 200 and exactly OK
  - 5.75 s prefill; 4.4 tok/s decode; 6.20 s total

The isolated test server was stopped after validation. GitHub CI was approved and is running.

PR #527, opened 2026-07-15
fix(ds4): honor sampler settings in decode

dusterbloom:
## Summary

- honor `GenerateRequest.sampler` in the single-device DeepSeek4 decode path
- seed penalties with the prompt and append committed generated tokens to sampling history
- retain the existing direct argmax path, without allocating or copying history when logit processing is inactive
- add a model-free regression test against the actual DeepSeek4 decode loop

## Why

`DeepSeek4Backend` always selected the maximum logit and never read the request sampler. Temperature sampling and repetition, frequency, and presence penalties were therefore ignored. Penalties also require the prompt plus generated tokens as history; generated-only history gives incorrect first-token behavior.

This PR is intentionally limited to the single-device backend. Layer-split sampling remains a separate change so each runtime can be reviewed and tested independently.

## Validation

- configured with GCC 11 / CUDA 12.0, matching the working local toolchain
- `cmake --build server/build-gcc11 --target test_deepseek4_unit -j 4`
- `./server/build-gcc11/test_deepseek4_unit` — passed; GPU-only HC test skipped because the leased GPU was unavailable
- `git diff --check`

The regression test uses identical logits and prompt history to verify both branches: inert greedy sampling selects the original argmax, while a repetition penalty applied to a prompt token changes the selected token.

### DIFF
--- a/server/src/deepseek4/deepseek4_backend.cpp
+++ b/server/src/deepseek4/deepseek4_backend.cpp
@@ -881,7 +881,14 @@ bool DeepSeek4Backend::do_decode(int committed, int n_gen,
     const auto phase_t0 = Clock::now();
     DeepSeek4StepTelemetry tel_acc;
     int steps = 0;
-    std::vector<int32_t> history = history_prefix;
+    const bool process_logits = sampler_.needs_logit_processing();
+    std::vector<int32_t> history;
+    if (process_logits) {
+        history = history_prefix;
+        if (n_gen > 0) {
+            history.reserve(history.size() + (size_t)n_gen);
+        }
+    }
 
     for (int generated = 0; generated < n_gen; generated++) {
         if (io.cancelled) break;
@@ -929,7 +936,7 @@ bool DeepSeek4Backend::do_decode(int committed, int n_gen,
 
         int32_t next_token = 0;
         const auto sample_t0 = Clock::now();
-        if (sampler_.needs_logit_processing()) {
+        if (process_logits) {
             next_token = sample_logits(logits.data(), w_.n_vocab, sampler_,
                                        history, sampler_rng_);
         } else {
@@ -942,7 +949,9 @@ bool DeepSeek4Backend::do_decode(int committed, int n_gen,
             }
         }
         if (timing) tel_acc.sample_us += elapsed_us(sample_t0, Clock::now());
-        history.push_back(next_token);
+        if (process_logits) {
+            history.push_back(next_token);
+        }
         out_tokens.push_back(next_token);
         const auto emit_t0 = Clock::now();
         io.emit(next_token);
--- a/server/src/deepseek4/deepseek4_backend.cpp
+++ b/server/src/deepseek4/deepseek4_backend.cpp
@@ -871,6 +871,7 @@ int DeepSeek4Backend::do_prefill(const std::vector<int32_t> & tokens,
 }
 
 bool DeepSeek4Backend::do_decode(int committed, int n_gen,
+                                  const std::vector<int32_t> & history_prefix,
                                   std::vector<int32_t> & out_tokens,
                                   const DaemonIO & io,
                                   const BudgetHook & budget_hook,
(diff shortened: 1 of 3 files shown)


Graffioh:
hey, here -> #507 in the [commits](https://github.com/Luce-Org/lucebox/pull/507/changes/b6a14e629ccd60ae7a15812e226ab0b56d1d2b19) i also tried to fix the sampler for layer split + broken greedy decode - can you take a look? let me know if i missed something, trying to learn while contributing

i know is a part of a bigger PR, so since i need a bit of time to test it, i/you can cherry pick that changes and put it in another independent PR

dusterbloom:
I am not 100% certain I understand your comment so here my tentative answer. 

I found out both our implementations share a issue that is:
0. `run_layer_split_ar_decode()` gives `sample_logits()` generated tokens but not prompt/restored-prefix history, so repetition/frequency/presence penalties are incomplete. 

My single-device patch handles the request prompt but still needs a proper restored-history contract.
 
Here what I think we should do: 
- keep #527 focused on single-device correctness;
- keep #507  layer-split propagation and GPU argmax credited to you; 

After we need to coordinate a small shared-history change and common behavioral test matrix. 

I also read Davide’s two-AMD-in-one-HIP-process requirement. Passing the device through the layer-range API, keying runtime state and F16 mirrors by device, selecting the device at shard entry, and using independently locked scratch slots looks like the right design.


On #505 at the commits on sampling a few things worth fixing:
  
1. The argmax test currently runs on the CPU backend. We still need CUDA/HIP parity tests for cached, dynamic and fused output paths,  including tie-breaking.
2. The two-device scratch test switches devices sequentially. A barrier-based concurrent test would exercise the actual same-process race.
3. `HcCudaScratch::ensure()` tracks only `d_state` capacity. The other shape-dependent buffers should also resize, and `d_mix` should  support the declared maximum `n_hc`, not only the default `n_hc=4`.
4. When sampling needs full logits, the path also computes/copies argmax unnecessarily.

Graffioh:
> I am not 100% certain I understand your comment so here my tentative answer.
> 
> I found out both our implementations share a issue that is: 0. `run_layer_split_ar_decode()` gives `sample_logits()` generated tokens but not prompt/restored-prefix history, so repetition/frequency/presence penalties are incomplete.
> 
> My single-device patch handles the request prompt but still needs a proper restored-history contract.
> 
> Here what I think we should do:
> 
> * keep [fix(ds4): honor sampler settings in decode #527](https://github.com/Luce-Org/lucebox/pull/527) focused on single-device correctness;
> * keep [fix(ds4): correct layer-split HC handoff, per-device state, and greedy decode in local path #507](https://github.com/Luce-Org/lucebox/pull/507)  layer-split propagation and GPU argmax credited to you;
> 
> After we need to coordinate a small shared-history change and common behavioral test matrix.
> 
> I also read Davide’s two-AMD-in-one-HIP-process requirement. Passing the device through the layer-range API, keying runtime state and F16 mirrors by device, selecting the device at shard entry, and using independently locked scratch slots looks like the right design.
> 
> On #505 at the commits on sampling a few things worth fixing:
> 
> 1. The argmax test currently runs on the CPU backend. We still need CUDA/HIP parity tests for cached, dynamic and fused output paths,  including tie-breaking.
> 2. The two-device scratch test switches devices sequentially. A barrier-based concurrent test would exercise the actual same-process race.
> 3. `HcCudaScratch::ensure()` tracks only `d_state` capacity. The other shape-dependent buffers should also resize, and `d_mix` should  support the declared maximum `n_hc`, not only the default `n_hc=4`.
> 4. When sampling needs full logits, the path also computes/copies argmax unnecessarily.

ok i got it - what i can do right now is decouple in two different PRs:
- sampler propagation in layer split [[PR](https://github.com/Luce-Org/lucebox/pull/530)]
- argmax on GPU [[PR](https://github.com/Luce-Org/lucebox/pull/531)]

then i can look at the stuff you mentioned worth fixing (together with more HC testing for 507 PR) - thank you for the feedback - if you prefer coordinating via discord for quicker feedback my handle is `graffioh`

PR #528, opened 2026-07-15
refactor(draft): codify proposal row layout

dusterbloom:
## Summary

Make the draft proposal-row layout explicit instead of deriving every artifact as seed plus proposals.

This adds a small DraftProposalShape contract with two layouts:

- SeedThenProposals for existing DFlash and DS4 artifacts
- ProposalsOnly for draft artifacts that emit candidate rows without the accepted seed

DraftWeights defaults to SeedThenProposals, so existing artifacts keep their current behavior.

## Why

The current block_size field is used both as an artifact row count and as a target verification width. Those values are equal for seed-plus-proposal artifacts, but not for proposal-only artifacts: four proposal rows require a five-row target verification batch after the accepted anchor is prepended.

Encoding that distinction as row-layout metadata keeps proposal count and verification width unambiguous without model-name conditionals. No runtime path consumes the new field in this PR; it establishes the shared representation used by follow-up integrations.

## Validation

The CPU-only contract test covers:

- existing 16-row layout: 15 proposals, verify width 16
- proposal-only 4-row layout: 4 proposals, verify width 5
- zero and one-row boundaries: zero proposals, verify width 1
- default layout compatibility

```text
c++ -std=c++17 -Wall -Wextra -Werror -Iserver/src server/test/test_draft_contract.cpp -o /tmp/test_draft_contract
/tmp/test_draft_contract
git diff --check origin/main...HEAD
```

All passed locally. The test links no GGML or GPU library and is included in both forms of the CMake check target.

No throughput benchmark is included because this PR does not change a runtime graph, kernel, scheduler, or model-loading path.

### DIFF
--- a/server/src/internal.h
+++ b/server/src/internal.h
@@ -25,6 +25,7 @@
 #include "gguf.h"
 
 #include "dflash27b.h"
+#include "draft/draft_contract.h"
 
 namespace dflash::common {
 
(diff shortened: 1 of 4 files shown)


server/src/internal.h
```diff
@@ -314,6 +315,11 @@ struct DraftWeights {
     int n_target_layers = DFLASH27B_DRAFT_N_TARGET_LAYERS;  // captured target layers (5)
     std::vector<int> capture_layer_ids;                     // explicit captured target-layer ids (GGUF dflash.target_layer_ids); empty = derive from count
     int mask_token_id   = DFLASH27B_DRAFT_MASK_TOKEN_ID;    // noise mask token
+    DraftProposalLayout proposal_layout = DraftProposalLayout::SeedThenProposals;
```
howard0su, line 318: intentional no consumer for this field in this PR?

PR #530, opened 2026-07-16
fix(ds4): propagate sampler config in layer split

Graffioh:
## Summary

- propagate request sampler settings into the DeepSeek4 layer-split adapter
- reseed the layer-split sampler RNG when sampling with an explicit seed
- add unit coverage for sampler propagation and deterministic seeding

## Scope

This PR covers the layer-split path. #527 remains focused on single-device decode correctness.

## Known limitation

`run_layer_split_ar_decode()` currently supplies generated tokens to `sample_logits()`, but not the prompt or restored-prefix history. Temperature, top-k, and top-p propagation are covered here; repetition, frequency, and presence penalties remain incomplete until both decode paths share a proper request-history contract.

## Validation

- `git diff --check` passed
- unit coverage verifies sampler propagation and deterministic seeding
- runtime GPU validation pending

### DIFF
--- a/server/src/deepseek4/deepseek4_layer_split_adapter.cpp
+++ b/server/src/deepseek4/deepseek4_layer_split_adapter.cpp
@@ -405,7 +405,10 @@ bool DeepSeek4LayerSplitAdapter::init_mixed_target_split() {
 }
 
 void DeepSeek4LayerSplitAdapter::begin_request(const GenerateRequest & req) {
-    (void)req;
+    sampler_ = req.sampler;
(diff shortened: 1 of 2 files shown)


Graffioh:
Superseded by #531. The sampler propagation change from this PR has been integrated into #531, where it is combined with the DeepSeek 4 GPU argmax work, mutually exclusive logits/argmax output paths, and CPU/CUDA/HIP parity coverage. Closing this PR so the combined change can be reviewed and merged in one place.

PR #533, opened 2026-07-16
feat: DFlash support for tensor-parallel Qwen3.6 targets

jkyamog:
Extend Qwen3.6 tensor parallelism from target-only autoregressive generation
to DFlash speculative decoding, including the device-local state and transfer
optimizations previously planned as a separate follow-up.

The draft remains on its explicit local device while target verification runs
through the TP meta backend. The change gathers complete target features for
the draft and extends rollback, recurrent state, snapshots, and prefix-cache
restore to split tensors.

## Changes

- Snapshot and restore recurrent TP state directly between each rank's GPU
  tensors instead of gathering it through host memory.
- Restore split SSM and convolution state on-device and synchronize once per
  rank after rollback.
- Decide DDTree eligibility before projection, avoiding duplicate LM-head
  execution.
- Run draft top-K on rank 0's mirrored GPU logits.
- Use the optimized rollback threshold for TP while preserving upstream's
  configurable rollback policy for single-device and layer-split execution.
- Batch feature-ring synchronization with one `get_2d` and one `set_2d` per
  non-wrapping run. Long prefills and prefix restores no longer perform two
  host/device transfers and stream synchronizations per token.
- Add opt-in aggregate phase timing with `DFLASH_TP_PROFILE=1`.

The stacked target branch also contains the review fixes for invalid device
counts, repeated split-state propagation and copies, multi-segment debug
output, legacy callback compatibility, and NaN-safe fallback all-reduce.

## Verification

- Full native CUDA build and 19/19 CTest targets pass on `main@f258c4a`.
- Chain speculative generation, DDTree, and fast rollback pass.
- Streaming forced-function output produces structured `tool_calls`.
- A 26,704-token prefix snapshot is committed, found, and restored for a
  26,758-token OpenClaw request. The repeated request's prefill drops from
  21.4 s to 0.1 s and produces byte-identical output.
- The deterministic 1,024-token OpenClaw control at 98,304 context produces
  SHA-256 `3614d727a950d5416a42b5e0e3713054e59ddc9a0365fb4ac1a2cbd198f5964f`,
  matching the saved pre-optimization and target-only continuation.
- The final full matrix completed without warnings for every viable row. TP
  plus draft peaked at 15,208/12,640 MiB for Q5 and 20,742/18,176 MiB for Q6
  on the two 3090s.

## Sustained performance

Hardware: RTX 5090 plus an NVLink-connected RTX 3090 pair at stock power.
Workload: 26,758 prompt tokens, 1,024 generated tokens, 98,304 context, Q8 KV,
Q8 draft, DDTree budget 16, no prefix cache, and a 60-second cooldown between
rows. These are single runs on the final rebased head. Every row uses 98,304
context except the explicitly marked Q6/5090 capacity row.

| Model / configuration | DFlash TG | AR TG | Relative | Prefill | Decode | Total |
|---|---:|---:|---:|---:|---:|---:|
| Q5, RTX 5090 | 126.2 tok/s | 59.5 tok/s | 2.12x | 10.3 s | 8.1 s | 18.4 s |
| Q5, 2x3090 layer split | 28.7 tok/s | 20.6 tok/s | 1.39x | 44.7 s | 35.7 s | 81.1 s |
| Q5, 2x3090 TP + NVLink | 86.5 tok/s | 40.4 tok/s | 2.14x | 20.7 s | 11.8 s | 32.5 s |
| Q5, 2x3090 TP, P2P disabled | 75.5 tok/s | 38.9 tok/s | 1.94x | 29.2 s | 13.6 s | 42.8 s |
| Q6, RTX 5090, 32,768 context | 80.1 tok/s | 48.7 tok/s | 1.64x | 10.3 s | 12.8 s | 23.1 s |
| Q6, 2x3090 layer split | 21.8 tok/s | 17.7 tok/s | 1.23x | 46.0 s | 47.0 s | 93.8 s |
| Q6, 2x3090 TP + NVLink | 60.5 tok/s | 34.8 tok/s | 1.74x | 21.4 s | 16.9 s | 38.4 s |
| Q6, 2x3090 TP, P2P disabled | 55.2 tok/s | 34.3 tok/s | 1.61x | 29.9 s | 18.6 s | 48.5 s |

Q5 performs 192 target forwards for 192 DDTree steps; Q6 performs 190 for
190. The optimized Q5 NVLink phase profile is:

```text
draft=2.379s project=0.407s snapshot=0.354s verify=8.430s
rollback=0.192s replay=0.000s feature=0.061s accounted=11.823s
```

Q5 TP is 3.01x faster than Q5 layer split and Q6 TP is 2.78x faster than Q6
layer split. Direct-P2P and P2P-disabled TP runs for both quants produced the
same byte-identical output hash shown above.

Q5 on one 24 GB 3090 failed while allocating a 1,411 MiB rollback-cache buffer
after reaching 23,286 MiB. The server logged `prefill_failed`, but the HTTP
layer returned 200 with empty assistant content and `finish_reason=stop`; that
error-mapping behavior is separate from this TP PR. Q6 on the 5090 aborts with
CUDA OOM at both 98,304 and 65,536 context, but succeeds at 32,768 context with
a 31,569 MiB peak. That constrained run produced the same assistant-content
hash as Q5 on the 5090:
`ac5279693a48ecc1b5e19daeadda5872f03bb47e40d589c6a40524db0596ef34`.

HIP was not compiled on this NVIDIA host because ROCm is not installed. The
new pointer-query calls use the existing CUDA/HIP compatibility header.

<a href="https://cubic.dev/pr/Luce-Org/lucebox/pull/533?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/src/qwen35/qwen35_backend.cpp
+++ b/server/src/qwen35/qwen35_backend.cpp
@@ -1939,14 +1939,8 @@ bool Qwen35Backend::do_spec_decode(int committed, int n_gen,
     int n_hint_proposed = 0;
     int n_hint_accepted = 0;
     int target_forwards = 0;
-    // TP rollback operates directly on each rank's device-local recurrent
-    // state, so it is cheaper than replay even for a one-token acceptance.
-    // Keep the configurable upstream policy for single-device and layer-split
-    // paths, and lower only the effective TP threshold.
-    ChainRollbackPolicy rollback_policy = resolve_chain_rollback_policy();
-    if (cfg_.device.is_tensor_parallel()) {
-        rollback_policy.fast_rollback_threshold = 1;
-    }
+    const ChainRollbackPolicy rollback_policy =
+        resolve_chain_rollback_policy(cfg_.device.is_tensor_parallel());
     const int fast_rollback_threshold =
         rollback_policy.fast_rollback_threshold;
     RollbackDiag rollback_diag;
--- a/server/src/qwen35/qwen35_backend.cpp
+++ b/server/src/qwen35/qwen35_backend.cpp
@@ -1939,8 +1939,39 @@ bool Qwen35Backend::do_spec_decode(int committed, int n_gen,
     int n_hint_proposed = 0;
     int n_hint_accepted = 0;
     int target_forwards = 0;
-    const ChainRollbackPolicy rollback_policy = resolve_chain_rollback_policy();
+    // TP rollback operates directly on each rank's device-local recurrent
+    // state, so it is cheaper than replay even for a one-token acceptance.
+    // Keep the configurable upstream policy for single-device and layer-split
+    // paths, and lower only the effective TP threshold.
+    ChainRollbackPolicy rollback_policy = resolve_chain_rollback_policy();
+    if (cfg_.device.is_tensor_parallel()) {
+        rollback_policy.fast_rollback_threshold = 1;
+    }
+    const int fast_rollback_threshold =
+        rollback_policy.fast_rollback_threshold;
     RollbackDiag rollback_diag;
+
+    const char * tp_profile_env = std::getenv("DFLASH_TP_PROFILE");
+    const bool tp_profile = tp_profile_env && std::strcmp(tp_profile_env, "0") != 0;
+    double profile_draft_s = 0.0;
+    double profile_project_s = 0.0;
+    double profile_snapshot_s = 0.0;
+    double profile_verify_s = 0.0;
+    double profile_rollback_s = 0.0;
+    double profile_replay_s = 0.0;
+    double profile_feature_s = 0.0;
+    auto profile_start = [&]() {
+        return tp_profile
+            ? std::chrono::steady_clock::now()
+            : std::chrono::steady_clock::time_point{};
+    };
+    auto profile_add = [&](double & total,
+                           std::chrono::steady_clock::time_point start) {
+        if (tp_profile) {
+            total += std::chrono::duration<double>(
+                std::chrono::steady_clock::now() - start).count();
+        }
+    };
 
     auto log_target_forward_stats = [&]() {
         std::fprintf(stderr, "[spec-decode] target_forwards=%d forwards_per_token=%.6f forwards_per_step=%.3f\n",
@@ -2012,6 +2043,7 @@ bool Qwen35Backend::do_spec_decode(int committed, int n_gen,
             !use_remote_draft &&
             draft_feature_mirror_can_view(feature_mirror_, committed, draft_ctx, mirror_slot0);
 
+        const auto profile_draft_start = profile_start();
         if (use_remote_draft) {
             local_hidden.clear();
             if (!remote_draft_.propose(committed, draft_ctx, noise_embed, local_hidden)) {
@@ -2098,14 +2130,7 @@ bool Qwen35Backend::do_spec_decode(int committed, int n_gen,
                                         sizeof(float) * local_hidden.size());
             }
         }
-
-        // 3. Project draft hidden → token IDs via target LM head
-        if (!target->project_hidden_to_tokens(local_hidden.data(), q_len, draft_tok)) {
-            std::fprintf(stderr, "spec-decode: projection failed\n");
-            step_graph_destroy(draft_sg);
-            return false;
-        }
-        draft_tok[0] = last_tok;
+        profile_add(profile_draft_s, profile_draft_start);
 
         // ── DDTree tree-structured verify ────────────────────────────────
         // When --ddtree is on and the target supports tree verify, build a
@@ -2135,18 +2160,34 @@ bool Qwen35Backend::do_spec_decode(int committed, int n_gen,
             (cfg_.fa_window == 0 &&
              committed + cfg_.ddtree_budget + 1 + cfg_.kq_stride_pad <= kvflash_tokens_ &&
              kvflash_pager_.identity_prefix_covers(committed));
-        if (cfg_.ddtree_mode && target->supports_tree_verify() && kvflash_tree_ok &&
-            !use_remote_draft && q_len > 1 && tree_special_inactive) {
+        const bool use_tree_verify =
+            cfg_.ddtree_mode && target->supports_tree_verify() && kvflash_tree_ok &&
+            !use_remote_draft && q_len > 1 && tree_special_inactive;
+
+        // DDTree consumes top-K rows directly. Avoid projecting the same
+        // hidden block once for argmax and again for top-K on every step.
+        if (!use_tree_verify) {
+            if (!target->project_hidden_to_tokens(local_hidden.data(), q_len, draft_tok)) {
+                std::fprintf(stderr, "spec-decode: projection failed\n");
+                step_graph_destroy(draft_sg);
+                return false;
+            }
+            draft_tok[0] = last_tok;
+        }
+
+        if (use_tree_verify) {
             const int L = q_len - 1;
             const int K = (cfg_.ddtree_budget > L) ? 8 : 1;
             std::vector<float>   top_lp;
             std::vector<int32_t> top_ids;
+            const auto profile_project_start = profile_start();
             if (!target->project_hidden_to_topk(local_hidden.data(), q_len, K,
                                                 cfg_.ddtree_temp, top_lp, top_ids)) {
                 std::fprintf(stderr, "spec-decode: ddtree topk projection failed\n");
                 step_graph_destroy(draft_sg);
                 return false;
             }
+            profile_add(profile_project_s, profile_project_start);
             // Tree depth L draws from draft rows 1..q_len-1 (row 0 = the seed).
             // Known limitation: branch descendants beyond depth 1 still come
             // from one spine-conditioned block-draft forward, so a confident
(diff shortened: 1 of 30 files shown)


jkyamog:
fyi:  I have been using this for more than a day, likely used around 20-50M tokens already.  So far its been good, saw a parser bug but not related to this PR.

  - Hardware: 2x RTX 3090 24 GB
  - Power limit: 300 W per GPU
  - Target: Qwen3.6-27B-Q6_K.gguf
  - Draft: dflash-draft-3.6-q8_0.gguf
  - Target execution: tensor parallel across both GPUs
  - CUDA peer access/NVLink: enabled
  - Context: 262,144 tokens
  - Default max output: 16,384 tokens
  - Thinking limit: 15,488 tokens
  - KV cache: Q8_0 K and V
  - DFlash/DDTree budget: 22
  - Draft SWA: 2,048
  - Flash-attention window: disabled/full-context attention
  - Prefix-cache slots: 4
  - Prefill-cache slots: 0
  - Fast rollback: enabled

This not too far from my 5090, but the 2x 3090 being 48gb has 2x the context size.

server/src/qwen35/qwen35_tensor_parallel.h
```diff
@@ -0,0 +1,41 @@
...
+
+ggml_backend_meta_split_state qwen35_tensor_parallel_split_state(
+    const ggml_tensor * tensor,
+    const TargetWeights & weights,
+    std::size_t device_count);
+
+class Qwen35TensorParallelContext {
+public:
+    static std::unique_ptr<Qwen35TensorParallelContext> create(
+        const DevicePlacement & placement,
```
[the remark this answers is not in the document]
jkyamog: Already addressed in 3eb9d70. Qwen35TensorParallelContext::create() now rejects device counts above GGML_BACKEND_META_MAX_DEVICES before constructing the meta backend, with regression coverage for 17 devices.

server/src/common/backend_factory.cpp
```diff
@@ -48,10 +48,9 @@ std::unique_ptr<ModelBackend> create_backend(const BackendArgs & args) {
 
     std::fprintf(stderr, "[backend_factory] detected arch=%s\n", arch.c_str());
 
-    if (args.ds4_prefill_mode_set && arch != "deepseek4") {
+    if (args.device.is_tensor_parallel() && arch != "qwen35") {
```
howard0su: shall we check device number here as well?

server/src/qwen35/qwen35_backend.cpp
```diff
@@ -1928,9 +1939,40 @@ bool Qwen35Backend::do_spec_decode(int committed, int n_gen,
     int n_hint_proposed = 0;
     int n_hint_accepted = 0;
     int target_forwards = 0;
-    const ChainRollbackPolicy rollback_policy = resolve_chain_rollback_policy();
+    // TP rollback operates directly on each rank's device-local recurrent
+    // state, so it is cheaper than replay even for a one-token acceptance.
+    // Keep the configurable upstream policy for single-device and layer-split
+    // paths, and lower only the effective TP threshold.
+    ChainRollbackPolicy rollback_policy = resolve_chain_rollback_policy();
+    if (cfg_.device.is_tensor_parallel()) {
```
howard0su: why this logic is not inside resolve_chain_rollback_policy?

server/src/qwen35/qwen35_dflash_target.cpp
```diff
@@ -21,6 +22,179 @@ using to_fp32_cuda_t = void (*)(const void *, float *, int64_t, cudaStream_t);
...
+namespace {
+
+bool is_meta_tensor(const ggml_tensor * tensor) {
+    return tensor && tensor->buffer &&
+        ggml_backend_buft_is_meta(
+            ggml_backend_buffer_get_type(tensor->buffer));
+}
+
+bool set_device_for_tensor(const ggml_tensor * tensor) {
+    if (!tensor || !tensor->data) return false;
```
howard0su: what is the case tensor is NULL here? I worry this check hides the bugs.

server/src/qwen35/qwen35_dflash_target.cpp
```diff
@@ -21,6 +22,179 @@ using to_fp32_cuda_t = void (*)(const void *, float *, int64_t, cudaStream_t);
 extern "C++" to_fp32_cuda_t ggml_get_to_fp32_cuda(ggml_type type);
 
 namespace dflash::common {
+namespace {
+
+bool is_meta_tensor(const ggml_tensor * tensor) {
+    return tensor && tensor->buffer &&
```
howard0su: what is the case tensor is NULL here? I worry this check hides the bugs.

server/src/qwen35/qwen35_dflash_target.cpp
```diff
@@ -21,6 +22,179 @@ using to_fp32_cuda_t = void (*)(const void *, float *, int64_t, cudaStream_t);
...
+                                   ggml_backend_t meta_backend) {
+    const size_t n_backends = ggml_backend_meta_n_backends(meta_backend);
+    for (size_t rank = 0; rank < n_backends; ++rank) {
+        const ggml_tensor * src = ggml_backend_meta_simple_tensor(captured, rank);
+        ggml_tensor * dst = ggml_backend_meta_simple_tensor(destination, rank);
+        const int64_t source_elements = src
+            ? src->ne[0] * src->ne[1] * src->ne[2]
+            : 0;
+        if (!src || !dst || dst->type != GGML_TYPE_F32 ||
+            capture_index < 0 || capture_index >= src->ne[3] ||
```
howard0su: capture_index can be verified out of the loop. please double check so much protective checks. some checks looks like ASSERT to me.

server/src/qwen35/qwen35_dflash_target.cpp
```diff
@@ -21,6 +22,179 @@ using to_fp32_cuda_t = void (*)(const void *, float *, int64_t, cudaStream_t);
...
+                                    const std::vector<int> & source_slots,
+                                    ggml_tensor * destination,
+                                    ggml_backend_t meta_backend) {
+    const size_t n_backends = ggml_backend_meta_n_backends(meta_backend);
+    for (size_t rank = 0; rank < n_backends; ++rank) {
+        const ggml_tensor * src = ggml_backend_meta_simple_tensor(captured, rank);
+        ggml_tensor * dst = ggml_backend_meta_simple_tensor(destination, rank);
+        if (!src || !dst || source_slots.empty() ||
+            src->type != GGML_TYPE_F32 ||
+            dst->type != GGML_TYPE_F32 ||
```
howard0su: why this have to be F32?

server/test/test_cuda_comm_api.cpp
```diff
@@ -0,0 +1,31 @@
+#include "ggml-backend.h"
```
howard0su, line 1: what this test really tested? seems not related to the changed code of ggml in this PR?

howard0su (changes requested):

server/src/common/backend_factory.cpp
jkyamog: **Backend factory device validation:** tensor mode now validates its
  placement at the factory boundary, including the minimum device count,
  duplicate devices, and incompatible split weights. Runtime-visible CUDA
  device IDs and the 16-device meta-backend limit remain validated in
  `Qwen35TensorParallelContext::create`, where the CUDA registry is available.

server/src/qwen35/qwen35_backend.cpp
jkyamog: **Rollback policy:** moved the TP threshold selection into
  `resolve_chain_rollback_policy(bool tensor_parallel)`. The backend now only
  passes its placement mode. Tests cover TP with both default and F32
  checkpoint policies.

server/src/qwen35/qwen35_dflash_target.cpp
jkyamog: **Null tensor checks:** tensor, buffer, data, and rank-local tensor
  preconditions are now `GGML_ASSERT`s. These are internal invariants rather
  than recoverable runtime failures. The helpers still return `false` for
  actual CUDA API failures.

server/src/qwen35/qwen35_dflash_target.cpp
jkyamog: **Null tensor checks:** tensor, buffer, data, and rank-local tensor
  preconditions are now `GGML_ASSERT`s. These are internal invariants rather
  than recoverable runtime failures. The helpers still return `false` for
  actual CUDA API failures.

server/src/qwen35/qwen35_dflash_target.cpp
jkyamog: **Repeated validation:** `capture_index` and convolution source-slot
  validation now happen once before the rank loop. Shape, type, and bounds
  relationships are asserted.

server/src/qwen35/qwen35_dflash_target.cpp
jkyamog: **Convolution F32 restriction:** removed. Convolution rollback now supports
  any matching non-quantized scalar type and derives offsets and widths from
  the tensor element size. SSM destination remains F32 because that path
  explicitly converts captured F16/F32 state into the live F32 recurrent
  state.

server/test/test_cuda_comm_api.cpp
jkyamog: **CUDA communicator test:** this test is inherited from stacked PR #532. It
  calls `ggml_backend_cuda_allreduce_tensor` to verify the public compatibility
  symbol still links, then passes the same CUDA backend twice to verify
  duplicate ranks are rejected before NCCL communicator creation. The test now
  documents both purposes inline.

jkyamog:
@howard0su Thank you very much for reviewing this PR.  This code base is definitely not something I can fully understand, I really appreciate your time in looking over it.  All of the code is written by codex, I direct it and ask questions.  I then test is on my own workload which hopefully can surface some major issues.  Over the weekend 3 lucebox (Qwen 27B) was working and did all of the coding workload.
image

davide221:
Restacked on the updated #532 branch and updated this branch at 06b9600dd. The follow-up fix caps prefix feature-sync host scratch at about 24 MiB instead of growing to roughly 600 MiB for a 4096-token Qwen3.6-27B restore, propagates every local and remote feature-sync failure instead of continuing with stale draft state, and records the complete vendored lucebox-ggml#39 delta. Validation on lucebox3: CUDA build plus 336/336 relevant tests; HIP build plus 324/324 relevant tests, including rollback-policy coverage. The real two-NVIDIA NCCL path remains the only hardware test not runnable on lucebox3. @howard0su, please re-review the updated stack.

jkyamog:
# Performance Update

Rebuilt and tested the rebased head `06b9600` with the same 26,758-token
OpenClaw payload and deterministic 1,024-token continuation used for the
earlier PR measurements.

| Configuration | PP | TG |
|---|---:|---:|
| RTX 5090 | **2,170.9 tok/s** | **110.6 tok/s** |
| 2x RTX 3090 TP + NVLink | **1,048.5 tok/s** | **76.2 tok/s** |

PR #534, opened 2026-07-16
perf(ggml): reuse ROCmFP4 weights across q=4 columns

davide221:
## Summary

- decode each packed `ROCmFP4_FAST` weight fragment once for dense four-column MMVQ
- reuse the decoded weights across the four q8 activation columns while preserving each column's original DP4A and floating-point accumulation order
- select the specialization automatically only on `gfx1151`, with no new runtime setting

The dispatch is shape- and backend-based rather than DS4-specific, so other
models using the same quantized dense four-column path can reuse it. Routed
`MUL_MAT_ID`, AR (`ncols_dst == 1`), other widths, and other architectures keep
the existing kernels.

## Performance

Measured on lucebox2 (`gfx1151`, ROCm 7.2.2), pinned to the same 2.9 GHz
`performance/high` state for both sides. The workload used fixed DSpark `q=4`,
`LUCE_MMVQ_MAX_NCOLS=4`, top-k 4, and the accepted ROCmFPX AR specializations.

Stable matched window: the first seven identical GSM+Math requests, totaling
896 generated tokens and 263 verification steps:

| | Control | This PR | Delta |
|---|---:|---:|---:|
| Speculative decode | 30.15 tok/s | 31.51 tok/s | +4.50% |
| Verification step | 112.99 ms | 108.13 ms | -4.30% |

The candidate's ten requests ranged from 28.7 to 35.0 tok/s. The control
encountered a host-wide latency stall on requests 9-10, so those requests are
not included in the comparison above.

AR remains unchanged at 26.88 tok/s, 5/5 GSM, and 583 output tokens.

## Correctness

- 76/76 traced q=4 verifier inputs and host logits are byte-identical to control
- zero missing/extra trace entries and zero byte mismatches
- AR produced the same aggregate throughput, score, and token count as control

## Kernel checks

A full Release HIP/gfx1151 `dflash_server` build succeeds. The dedicated
specializations have no register spills, private segment, or dynamic stack:

| Kernel | SGPR | VGPR |
|---|---:|---:|
| q=4 fused specialization | 47 | 59 |
| q=4 unfused specialization | 28 | 46 |
| existing q=4 fused kernel | 50 | 67 |
| existing q=4 unfused kernel | 32 | 83 |

The existing generic q=4 kernel metadata is unchanged from control.

## Stack

This is intentionally based on `codex/deepseek4-dspark-draft` / #496 because
that branch owns the fixed-q=4 DSpark verification path used for validation.

<a href="https://cubic.dev/pr/Luce-Org/lucebox/pull/534?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/deps/llama.cpp/ggml/src/ggml-cuda/mmvq.cu
+++ b/server/deps/llama.cpp/ggml/src/ggml-cuda/mmvq.cu
@@ -501,8 +501,40 @@ static bool is_gfx1151(const int cc) {
     return cc == GGML_CUDA_CC_OFFSET_AMD + 0x1151;
 }
 
+// Dense four-column projections use the same weight row for every column.
+// Decode each packed ROCmFP4 fragment once while preserving the original
+// per-column DP4A and floating-point accumulation order.
+static __device__ __forceinline__ void vec_dot_rocmfp4_fast_q8_1_4cols(
+        const void * __restrict__ vbq, const block_q8_1 * __restrict__ y,
+        const uint32_t stride_col_y, const int kby, const int kbx, const int iqs,
+        float (&result)[4]) {
+    const block_rocmfp4_fast * bq4 = (const block_rocmfp4_fast *) vbq + kbx;
+
+    int2 weights[VDR_ROCMFP4_FAST_Q8_1_MMVQ];
+#pragma unroll
+    for (int l = 0; l < VDR_ROCMFP4_FAST_Q8_1_MMVQ; ++l) {
+        const int packed = rocmfp4_get_qs_i32(bq4->qs, iqs + l);
+        weights[l] = rocmfp4_get_int_from_codebook_16(packed, kvalues_rocmfp4);
+    }
+
+    const float weight_scale = rocmfp4_ue4m3_to_fp32_half_finite(bq4->e);
+    int sums[4] = {};
+#pragma unroll
+    for (int j = 0; j < 4; ++j) {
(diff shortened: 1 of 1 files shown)


davide221:
Tested on Lucebox6 (Strix Halo gfx1151) with the ROCmFP2 target and DSpark Q4RMFP4 draft, q=4.

Results:
- Fixed q=4: **+2.31% weighted decode** across 5,062 clean paired tokens; output was byte-identical for every prompt.
- Default adaptive mode: **+2.12% conservative weighted decode** across 2,271 paired tokens; output was byte-identical.
- One control-only `math_05` timing outlier was excluded from the conservative totals. Including it would report +4.00% adaptive, so +2.12% is the defensible figure.
- The benchmark isolated the patch against its exact pre-rebase parent (`a8465cca`). Final rebased SHA `022c7b56` retains the same one-file kernel change and passes a fresh ROCm Release build, `test_draft_topk_cuda` (10/10), and `test_deepseek4_unit`.

Scope: this improves DSpark q=4 speculative verification/decode on gfx1151. It does not improve prompt prefill or normal one-token AR.

PR #552, opened 2026-07-21
fix(server): accept malformed Qwen tool-call variants

jkyamog:
Some Qwen agent responses contain recognizable tool calls but vary from the
documented `<function=NAME>` XML form. The server previously streamed these
variants as visible assistant text instead of returning structured
`tool_calls`.

This change adds strict compatibility for three observed forms:

```xml
<parameter name="bash"><parameter name="command">...</parameter></function>

<funcname>read
<parameter=path>...</parameter>
</function>

<function read>
<parameter=path>...</parameter>
</function>
```

The streaming detector now holds each opener until parsing completes, so tool
syntax does not leak into assistant content.

Streaming and final parsing now share one opener scanner. It recognizes the
fixed compatibility forms plus `<DECLARED_TOOL>` for valid request-declared
tool names, avoiding a separate hard-coded streaming list that can drift from
the final parser.

# Safety

- A parsed function name must match a tool declared in the request.
- Compatibility forms require complete parameter elements.
- Unexpected body text, duplicate parameters, unknown tools, and incomplete
  syntax are rejected rather than partially parsed or guessed.
- Function arguments remain JSON strings in OpenAI-compatible responses.

# Reproduction

The newest failure was a two-call response emitted as visible text:

```xml
Let me read the file and compute its SHA-256 hash.

<function read>
<parameter=path>
/tmp/tool-input.md
</parameter>
</function>

<function bash>
<parameter=command>
sha256sum /tmp/tool-input.md
</parameter>
</function>
```

After this change, the preamble remains assistant content and the two blocks
become ordered `read` and `bash` tool calls.

### DIFF
--- a/server/src/server/sse_emitter.cpp
+++ b/server/src/server/sse_emitter.cpp
@@ -13,13 +13,6 @@ namespace dflash::common {
 
 static const char THINK_OPEN[]  = "<think>";
 static const char THINK_CLOSE[] = "</think>";
-static const char TOOL_OPEN[]   = "<tool_call>";
-static const char FUNCTION_OPEN[] = "<function=";
-static const char TOOL_CODE_OPEN[] = "<tool_code>";
-// Laguna emits `NAME<arg_key>K</arg_key><arg_value>V</arg_value>` with the
-// <tool_call> wrapper as special tokens that detokenization strips, so the
-// visible trigger is <arg_key> and the tool name sits immediately before it.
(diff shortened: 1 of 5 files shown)


jkyamog:
Discovered a few more issues with the parser this is over 100-200M tokens generated in the last few days.  2 of the issues has happened more than 2x.

jkyamog:
encountered more parser issues, so I have directed codex add more test cases, after which refactor the parsing instead small patches.

PR #567, opened 2026-07-31
feat(prefix-cache): sticky tools pins + DiffPin for warm agent TTFT

davidmroth:
## Summary

Follow-up to [#492](https://github.com/Luce-Org/lucebox/pull/492) (native tool-prefix caching).
That work made system/tools KV reusable across growing agent turns. Two gaps
remained in multi-chat / long-lived tool-heavy agent traffic:

1. **Sticky protected tools-boundary pins** — cold tool-heavy turns pin the
   system+tools head and mark those entries protected so deepen snaps from
   other sessions cannot thrash large tool prefixes out of the slot pool.
2. **DiffPin (PPP)** — when a small volatile island (session clock, banner) sits
   *inside* an otherwise-stable head, exact prefix match fails and the engine
   re-prefills tens of thousands of tokens. DiffPin diffs the head against
   recent traffic, annotates a sticky `pin_end`, and (optionally) relocates the
   ephemeral middle after the shared prefix+suffix so the contiguous stable
   blob can be pinned.

No conversation-id header or client protocol change is required. Defaults keep
DiffPin annotation on and **token rearrange off** (safe for tool JSON):

| Env | Default | Role |
|---|---|---|
| `DFLASH_PPP` | `1` | LCP pin-end + sticky protect |
| `DFLASH_PPP_REARRANGE` | `0` | Opt-in token-level float rewrite |
| `DFLASH_PPP_LCP_WINDOW` / `MIN_PIN` / `MAX_EPHEMERAL` | (see docs) | DiffPin knobs |

Docs: `server/docs/PIN_FRIENDLY_PROMPT.md` (linked from `TOOL_PREFIX_CACHE.md`).
`usage.timings` continues to report `cache_hit`, `cached_prefix_tokens`,
`prefilled_tokens`, and `effective_prompt_tokens`.

## Why rearrange defaults off

An early DiffPin wiring applied token-level rewrite even when `rearrange=0`.
That could peel a mid-head island out of a large tools/system prefix
(`diff-rewrite … middle≈170` on ~17–18K-token tool schemas) and scramble tool
JSON enough to yield empty post-tool completions. Gating rewrite behind
`DFLASH_PPP_REARRANGE` (default off) keeps the default path to **pin-end
annotate + sticky protect only** (`pin-only` / `no rewrite` in logs). Opt in
to rearrange when you deliberately want clock/banner float.

## Measured warm reuse

Self-hosted production `dflash_server` (native HTTP path), ~28-tool agent
prompts in the ~17–24K token range:

| Turn | Prompt | Cached prefix | Prefill | Restore | Notes |
|---|---:|---:|---:|---|---|
| Cold tools head | 19,502 | 0 | **86.3 s** | false | pin tools boundary |
| Warm deepen | 22,822 | 17,848 (78%) | **25.5 s** | true | sticky pin hit |
| Warm deepen (deeper) | 23,243 | 22,249 (96%) | **6.0 s** | true | long decode turn |
| Warm deepen | 23,903 | 23,173 (97%) | **4.9 s** | true | short stop turn |
| Warm deepen | 23,993 | 23,889 (99.6%) | **1.8 s** | true | ~13× fewer prefill seconds vs cold head |

On the last warm turn, ~24K prompt / 1.8 s ≈ **13K effective t/s**, while only
~100 uncached tokens were actually prefilling — the KV pin is doing the work.

Example log shape with safe defaults:

```
[ppp] enabled=1 rearrange=0 …
[ppp] pin_end=… (pin-only; prompt=…)
[pc] lookup hit … prefix_len=…
[server] chat CACHE … restore=true … prefill=1.8s
```

This is a **prefill / TTFT** result, not a claim about decode throughput.

## Reproduce

Against a `dflash_server` with prefix cache enabled:

```bash
python3 server/scripts/benchmark_tool_prefix_cache.py \
  --url http://127.0.0.1:8080 \
  --json-out /tmp/diffpin-prefix.json
```

Confirm warm turns report `cached_prefix_tokens > 0` and `restore=true` in
server logs (`[pc] lookup hit`, `[server] chat CACHE … restore=true`).

For DiffPin specifically, send two tool-heavy turns that differ only by a small
mid-head volatile line (e.g. session clock) and confirm either a sticky pin hit
without rewrite (`pin-only` / `no rewrite`), or — with `DFLASH_PPP_REARRANGE=1` —
a PPP rearrange + protected pin of the stable blob.

## Validation

- [x] Unit coverage for DiffPin / sticky-pin helpers in `test_server_unit`
- [x] `dflash_server` build with rearrange gated off by default
- [x] Startup logs `[ppp] enabled=1 rearrange=0`
- [x] Warm tool-heavy turns restore with large `prefix_len` near the tools pin
- [x] Post-tool completions stable with default `pin-only` path
- [ ] Reviewer: tool-prefix benchmark still shows warm restore (#492 regression check)
- [ ] Reviewer: multi-chat deepen does not immediately evict protected tools pins

@davide221, Hey, thanks so much for helping out with that last PR! Just give me a shout if there's anything else you need me to do on this one to get it all wrapped up and merged.

### DIFF
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -342,6 +342,7 @@ add_library(dflash_common STATIC
     src/server/tool_memory.cpp
     src/server/sse_emitter.cpp
     src/server/prefix_cache.cpp
+    src/server/pin_friendly_prompt.cpp
     src/server/disk_prefix_cache.cpp
     src/server/freeze_history.cpp
     # ── Jinja chat-template engine (vendored under deps/llama.cpp/common/) ──
--- a/server/docs/PIN_FRIENDLY_PROMPT.md
+++ b/server/docs/PIN_FRIENDLY_PROMPT.md
@@ -155,6 +155,8 @@ flowchart TD
 
 End-of-message markers (the “this turn is over” tokens) stay at the end of the head so the chat shape still reads as a coherent system turn.
 
+The rewrite head stops at the **first end-of-message marker**, not at the PrefixCache chat boundary. Those boundaries sit *after* the next role-start; cutting there would let a floated middle cross into the user turn. With no chat boundaries (custom templates), DiffPin does not rewrite.
+
 ---
 
 ## What you should feel in production
--- a/server/docs/PIN_FRIENDLY_PROMPT.md
+++ b/server/docs/PIN_FRIENDLY_PROMPT.md
@@ -1,0 +1,216 @@
+# DiffPin
+
+**Making agent prompts prefix-cacheable by floating volatility out of the stable head**
+
+---
+
+## In one sentence
+
+**DiffPin** finds the small changing part of a long prompt (a session clock, a banner, a one-line header), moves it *after* everything that stayed the same, and tells the inference engine: *pin the KV cache here — this contiguous block will come back*.
+
+---
+
+## Why this exists
+
+Modern agents send a huge head on every turn: identity, rules, dozens of tool schemas. That head is expensive to **prefill** — the model must write a **KV cache** (the memory of “what I already saw”) before it can answer.
+
+Most of that head does not change. A few tokens do — often a line like *Conversation started: Friday, July 31…*.
+
+Prefix caching only works when the **beginning** of the prompt matches exactly. One changed date in the middle of the head, and the engine treats the whole head as new. You pay the long prefill again.
+
+```mermaid
+flowchart LR
+  subgraph cold ["Without DiffPin"]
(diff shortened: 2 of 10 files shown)


davide221:
@davidmroth thanks for the PR! One problem seems that > DFLASH_PPP=0 disables PPP annotation/rewrite, but not sticky/protected tool-boundary pinning, despite the docs saying it controlsboth. Could you gate prefer_tools_boundary on config_.ppp_enabled?

Issue #572, opened 2026-08-01
DeepSeek V4 Flash on Strix Halo (gfx1151): 4 defects in the monolithic DS4 path, 2 validated patches attached


GRAVESIS:
Reported against `Luce-Org/lucebox` at commit `c3b71e4b2ed2d08db96d55e7985f20035aa08d6f`, using the ROCMFP2 Strix quant from Hugging Face `Lucebox/DeepSeek-V4-Flash-ROCMFPX`.

Found while qualifying local models on a two-node Strix Halo cluster. Everything below comes from the single-node monolithic DS4 configuration on one of those boxes. Four defects, two of them with patches that are validated on this hardware and inlined at the end, one with a working sidecar card, and one diagnosed but not fixed. Happy to raise the patched items as PRs instead if that is easier to review.

## Summary

| # | Defect | Severity | Status |
|---|---|---|---|
| 1 | Non-fused DSpark verify reads past the logits tensor and aborts | Fatal, deterministic | Patch below, validated |
| 2 | Prefix cache never stores or restores for DeepSeek models | Severe performance | Patch below, validated |
| 3 | No `deepseek4` model-card family, so reasoning silently disables | Silent quality loss | Root cause named, working card below, code fix suggested |
| 4 | DSpark output is not greedy-identical to target-only decoding | Correctness question | Diagnosed, not fixed |

## Environment

- AMD Ryzen AI Max+ 395 (Strix Halo), `gfx1151`, wave size 32, 128 GB unified memory, 122880 MiB reported to the HIP device
- ROCm 7.2.4, HIP backend, `dflash_server` built as `build-hip-strix` from the commit above
- Target: `DeepSeek-V4-Flash-ROCMFP2-STRIX.gguf`, 102,320,631,200 bytes, sha256 `8fa6c30d9badd8e72f83c62952fe24b2e8cd647e103aff086bece328cb9c9208`
- Drafter: `DeepSeek-V4-Flash-DSpark-draft-Q4RMFP4-denseF16.gguf`, 11,304,737,056 bytes, sha256 `48883d35b8a67ecfd2858a90e12a47d04cb5ac581acef868ca0f58544816f746`
- Common serving flags: `--target-device hip:0 --ds4-fused-decode --ds4-expert-top-k 6 --cache-type-k f16 --cache-type-v f16 --max-ctx 16384 --chunk 2048`
[...]

PR #579, opened 2026-08-03
fix(ds4): enable monolithic prefix cache

davide221:
## Summary

- recognize the exact DeepSeek V4 chat control tokens so reusable turn boundaries are found
- implement monolithic DeepSeek V4 snapshot save, restore, exact-hit decode, and suffix-only prefill
- preserve logits and the bounded DSpark feature window across cache hits
- right-size compressed KV snapshot rows so slot memory follows live state rather than maximum context
- track the cache position represented by decode logits and reject unsafe post-DSpark snapshots
- release all auxiliary snapshot RAM when the target is parked
- stop scheduling disk work after detecting a memory-only snapshot backend
- add marker, cache-state, zero-row, feature-window, lifecycle, disk-gating, and restore regression tests

This addresses prefix-cache defect 2 from #572. The separate all-logits safety fix described there is already present on main.

## Safety

DeepSeek detection requires all four strings to be exact special vocabulary tokens and leaves every other tokenizer family unchanged. Restore validates the complete snapshot layout before copying it into the live cache. Inactive compressed-cache capacity is intentionally omitted because its logical row counts hide it and future writes replace it.

A snapshot is now accepted only when its saved logits represent the same absolute position as the core cache. DSpark post-decode state is conservatively rejected because that API does not return final target logits.

## Validation

Validated on lucebox2 with ROCm gfx1151 and the 102 GB DeepSeek-V4-Flash ROCMFP2 Strix model:

- HIP build succeeded for the server and all test targets
- full CTest suite: 344/344 passed
- review-fix real-model smoke: cold and exact-hit responses were byte-identical; warm prefill was 0.0 ms
- memory-only disk capability was detected and disk work disabled after the first live inline snapshot
- AR exact hit: 2716.0 ms cold prefill to 0.0 ms warm; response content byte-identical
- AR suffix hit: 2986.9 ms cold to 576.7 ms warm; response content byte-identical
- DSpark exact hit: 2901.1 ms cold to 0.0 ms warm; response content byte-identical, acceptance 1.0
- DSpark suffix hit: 3050.7 ms cold to 554.2 ms warm; response content byte-identical, acceptance 1.0
- 968-token disjoint-window DSpark case: bounded 20.2 MiB snapshot, byte-identical response, acceptance 1.0

The exact and suffix probes exercised the real HTTP prefix-cache path, not only the snapshot helpers.

### DIFF
--- a/server/src/deepseek4/deepseek4_backend.cpp
+++ b/server/src/deepseek4/deepseek4_backend.cpp
@@ -22,6 +22,7 @@
 #include <cstring>
 #include <cinttypes>
 #include <limits>
+#include <new>
 
 namespace dflash::common {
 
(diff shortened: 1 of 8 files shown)


davide221:
Addressed all four automated-review findings in fd4ae9d:

- exact-hit unit coverage now decodes and asserts the restored-logit token
- logits carry an absolute cache position; stale post-DSpark snapshots are rejected and AR forwards keep logits synchronized
- target parking frees both core and auxiliary snapshot memory
- a live snapshot without SnapshotRef disables later disk-cache work instead of repeatedly attempting non-persistent saves

Revalidated on lucebox2: HIP/full build succeeded, CTest 344/344 passed, and the real 102 GB model produced byte-identical cold/exact-hit output with 0 ms warm prefill. The real inline-snapshot path also confirmed the memory-only disk gate.

PR #580, opened 2026-08-03
fix(server): preserve UTF-8 across stream boundaries

Amidwestnoob:
## Summary

- Preserve valid UTF-8 code points when token output splits a multibyte sequence across stream chunks.
- Reject malformed and impossible UTF-8 prefixes without retaining unbounded state.
- Discard a pending incomplete suffix when a stop sequence matches.
- Flush a genuinely truncated final code point as one replacement character.
- Use one finish-reason policy for streaming responses, non-streaming responses, and completion logs.
- Preserve semantic reasons such as `tool_calls`; report capped output as `length` / `max_tokens`.

## Problem

The emitter sanitized each token fragment independently. A valid multibyte code point split across fragments could therefore become replacement characters. Finish-reason derivation also differed between API responses and the final server log.

## Validation

- Release `dflash_server` build passed.
- Server unit suite: 344/344 passed.
- Feature gate: 126 assertions, 0 failures.
- CUDA communicator, Qwen3.5 tensor split, and two-GPU NCCL all-reduce tests passed.
- Live streaming and non-streaming output preserved `café — π ∈ ∅ ⊆` and `∀∃∈∅⊆⊇∩≠→` in layer and tensor modes.
- Live capped output returned and logged `length`.
- Live tool output preserved `tool_calls`.
- A one-hour mixed runtime test remained healthy with no transport, CUDA, NCCL, fatal, or replacement-character failures.
- Production was restored unchanged after validation.

Patch SHA-256:

`1c7b29b1fee3db63e52b696bc7e8812110f396feb1fd2b09ffa614a69a0470db`

## Scope

This patch changes stream assembly and finish-reason reporting only. It does not change tensor-parallel math, speculative decoding, model loading, draft compatibility, or production configuration.

<a href="https://cubic.dev/pr/Luce-Org/lucebox/pull/580?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/src/server/finish_reason.h
+++ b/server/src/server/finish_reason.h
@@ -1,0 +1,33 @@
+// Shared client-facing finish-reason policy for HTTP responses and logs.
+#pragma once
+
+#include <string>
+
+namespace dflash::common {
+
+// Resolve the public reason from the emitter's semantic reason and request
+// outcome. `generation_cap < 0` disables the cap comparison for callers that
+// do not have a generation budget. Disconnect and backend errors retain their
+// existing log-only states and take precedence over normal completion.
+inline std::string resolve_client_finish_reason(
+        const std::string & emitter_reason,
+        int completion_tokens,
+        int generation_cap,
+        bool degenerate_decode_close,
+        bool result_ok = true,
+        bool client_disconnected = false) {
+    if (client_disconnected) return "client_disconnect";
+    if (!result_ok) return "error";
+
+    std::string reason = emitter_reason;
+    if (reason == "stop" && generation_cap >= 0 &&
+        completion_tokens >= generation_cap) {
+        reason = "length";
+    }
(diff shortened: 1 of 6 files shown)


Amidwestnoob:
Closing this one. The core UTF-8 boundary handling has since landed upstream in #641, and current main now has newer EOS and tool-reasoning behavior in the same code. Rebasing this would mean untangling overlapping implementations for a fairly small remaining benefit. The stricter validation and failed-generation work were useful, but there isn't enough left here to justify carrying the PR forward. Thanks to everyone who reviewed it.

PR #590, opened 2026-08-06
perf(dflash): add opt-in exact prefill bands

cheese-cakee:
## Summary

Adds an opt-in exact-prefill band scheduler for DeepSeek-V4 while keeping tokenwise q1 exact prefill as the safe default.

- enables compressor-safe q2-q4 exact-prefill bands with `DFLASH_DS4_EXACT_PREFILL_BANDS=1`;
- integrates #666 by @dusterbloom so exact width-2-to-4 feed-forward graphs use the qualified narrow MMVQ path;
- preserves exact behavior across attention, MoE routing, cache updates, logits, snapshots, and repeated requests;
- respects an explicit `LUCE_MMVQ_MAX_NCOLS` threshold and retains fused MMQ whenever grouped MMVQ does not support the selected width;
- keeps q1 available immediately by leaving the exact-band flag unset.

This is an opt-in performance PR, not a general DeepSeek bug fix.

## Integrated history

Current head: `25625f95f9e364d3b953ad027fa563f57a2088aa`

PR #666 is merged as commit `27b615dc4b8246b2c1a7ed085fecd0bdfe783efb`, preserving dusterbloom's authorship and original commit. The integration also resolves its two pending dispatch-threshold review concerns in `2e568087`.

## Existing #590 qualification

The following qualification was completed before the #666 integration at #590 head `c8f9cc07e4ba708f0b2e9c473cef9f4196867f05`.

Matched Lucebox6 condition: R9700 gfx1201 + Strix Halo gfx1151, exact prefill, `--chunk 4`, identical model, topology, prompt, and runtime settings, caches disabled. Each arm used two warmups followed by five measured requests.

| Mode | Measured prefill samples (tok/s) | Median | MAD |
|---|---|---:|---:|
| q1 reference A | 16.82, 16.58, 16.68, 16.76, 16.77 | 16.76 | 0.06 |
| exact q2-q4 bands | 17.70, 17.62, 17.59, 17.58, 17.89 | 17.62 | 0.04 |
| q1 reference B | 16.62, 16.49, 16.58, 16.71, 16.71 | 16.62 | 0.09 |

The pooled 10-sample q1 median was **16.695 tok/s**. Exact bands reached **17.62 tok/s**, a **5.54% gain**. The two q1 brackets differed by **0.84%**.

All 21 requests were cache-cold and produced the same response SHA-256:

`cd5cb9fb5ac3c4f4007e8b41d117da21622439cd05c1728f3e82f90e4f869dad`

Evidence root: `/home/cheese/pr590-pr633-ready-20260825T1815Z`  
Evidence manifest SHA-256: `3517837f76bb5978154ca4de46e47625fbf2fbe91d1fbfd4aa6c0d0f13b52140`

## Standalone #666 qualification

PR #666 was independently qualified at `27b615dc` on gfx1151 with ROCm 7.2.2, exact all-hot prefill, no drafter, and caches disabled.

| Prompt | Main median | #666 median | Lower prefill time |
|---|---:|---:|---:|
| 64 tokens | 3341.6 ms | 2776.9 ms | 16.90% |
| 128 tokens | 6710.7 ms | 5602.6 ms | 16.51% |

Its frozen five-run qualification measured 17.72% and 16.62% lower total prefill time, with byte-identical q1/q4 route IDs, route weights, recurrent state, all-position F32 logits, and committed tokens. These are #666's standalone all-hot M64/M128 results, not a newly measured combined-head claim.

## Band selection

`--chunk` remains the operator's maximum exact-band cap. The scheduler downshifts at compressor boundaries and prompt tails. Leaving `DFLASH_DS4_EXACT_PREFILL_BANDS` unset preserves q1. The integrated narrow-MMVQ path applies only to active exact multi-token bands and does not change speculative verification or approximate prefill.


Graffioh:
i'm wondering: should `q{1,2,3,4}` be selected based on --chunk? i guess you measured for r9700 / strix halo and set based on that but if the best width varies by GPU or topology, and we want to support multiple GPUs without overfitting, then could startup select the largest qualified width automatically via some startup-level measurements, while retaining q1 as an explicit fallback? that would avoid requiring operators to choose a hardware-specific width through the general --chunk option

cc @davide221

cheese-cakee:
Good question. In this PR, --chunk acts as the maximum exact-band width, not a width forced onto every step. The scheduler downshifts at learned-compressor boundaries and prompt tails, while leaving DFLASH_DS4_EXACT_PREFILL_BANDS unset gives the explicit q1 fallback. I kept startup autotuning out of this diff because it would need a separate, fail-closed contract: only prequalified exact widths could be considered, calibration would need isolated device state, and the result should not silently vary numerical policy by startup noise. The updated head is now merged with current main and freshly qualified on the R9700 + Strix topology; I also added this distinction to the PR body.

cheese-cakee:
Fixed the fresh-review P1 in c8f9cc07. Exact q4 still splits owner evaluation as q3+q1 for reduction-order parity, but a device-resident full-batch activation is no longer sliced as an empty host buffer. Split batches perform one normalized-activation readback before owner evaluation; unsplit batches retain the device-resident path. The forced DFLASH_MOE_EXPERT_MAJOR_MIN_TOKENS=4 regression run exercised 30,100 traced split fallbacks across 7/7 full-model requests, preserved the established exact response SHA, and left KFD clean. Both gfx1201 and gfx1151 unit suites also exit OK.

Graffioh:
i discussed on discord with @dusterbloom and we think it is good to merge https://github.com/Luce-Org/lucebox/pull/666 inside this PR

666 is minimal and is a clean addition to this PR that is more complete overall

cheese-cakee:
Integrated #666 into this PR at `25625f95`, preserving `27b615dc` and @dusterbloom's authorship. I also merged current `main` and addressed the two pending MMVQ dispatch-threshold review concerns. The combined head is mergeable and the PR body now separates the existing #590 and standalone #666 measurements.

PR #618, opened 2026-08-17
feat(server): steer DeepSeek V4 to use <function_call> tags for tool calls

dpavlin:
Replaces the raw tools_json dump in the ChatFormat::DEEPSEEK4 native template renderer with an explicit system-prompt block that:

  1. Wraps each tool schema inside <available_tools>...</available_tools>, matching the upstream DeepSeek GGUF/Jinja chat template convention.
  2. Instructs the model to return every tool invocation as a single JSON object enclosed in <function_call>...</function_call> tags.

Without this steering the model defaults to emitting plain Markdown JSON code blocks (e.g. ```json {...} ```) which bypass the tag-based parser in tool_parser.cpp and leak raw text to the client.

This change is scoped exclusively to ChatFormat::DEEPSEEK4 (selected only when arch == "deepseek4") so no other model format is affected.

<a href="https://cubic.dev/pr/Luce-Org/lucebox/pull/618?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/src/server/chat_template.cpp
+++ b/server/src/server/chat_template.cpp
@@ -44,6 +44,25 @@ static const char QWEN3_TOOL_SUFFIX[] =
     "current knowledge and do not tell the user about function calls\n"
     "</IMPORTANT>";
 
+// Appends "<available_tools>\n{tool}\n...</available_tools>\n\n" to result.
+// Each tool is pretty-printed as compact JSON; falls back to the raw
+// tools_json string if parsing fails.
+static void append_available_tools(std::string & result,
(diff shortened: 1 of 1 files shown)


davide221:
lgtm

PR #622, opened 2026-08-18
fix(qwen35): size mixed concurrent target-step graphs

Graffioh:
## Problem

The Qwen target-step builder always allocates a 16,384-node graph. That is large enough for eight independent recurrent segments, but the default C16 continuous-serving path can legally schedule eight packed prefills plus compact decode in one traversal.

That ninth recurrent segment builds 16,741 nodes and aborts during graph construction. This reproduces on `main` with the normal continuous-decode scheduler; it is not caused by an experimental policy.

## Fix

- Keep the existing 16,384-node capacity through eight recurrent segments.
- Add 2,048 nodes of capacity for each recurrent segment beyond eight.
- Size the target-step graph from packed-prefill segment count plus compact decode.
- Add a host regression test for the C16 mixed shape and larger future cohorts.

The observed C16 schedule now receives an 18,432-node graph. Common shapes at or below eight recurrent segments retain the existing capacity.

## Scope

This PR is now only the graph-sizing correctness fix. The prefill-first scheduling experiment was removed after warmed three-repeat gfx1151 screening showed only a 1-3% C16 TTFT gain, a C8 TTFT regression, and materially worse overlap ITL. The existing invariant that every normal decode plan covers the complete live cohort remains unchanged.

## Validation

- Graph-sizing regression passed.
- Scheduler plan: 62 checks passed.
- Sequence-engine contract: 14 checks passed.
- Complete server unit suite: 382/382 passed.
- HIP server and focused targets rebuilt successfully.
- `git diff --check` passed.

### DIFF
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -1117,6 +1117,13 @@ if(DFLASH27B_TESTS)
         target_link_libraries(test_qwen35_roctx PRIVATE ${CMAKE_DL_LIBS})
         list(APPEND _raw_unit_test_targets test_qwen35_roctx)
     endif()
+    if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test/test_qwen35_graph_sizing.cpp")
+        add_executable(test_qwen35_graph_sizing
+            test/test_qwen35_graph_sizing.cpp)
+        target_include_directories(test_qwen35_graph_sizing PRIVATE
+            ${CMAKE_CURRENT_SOURCE_DIR}/src)
+        list(APPEND _raw_unit_test_targets test_qwen35_graph_sizing)
+    endif()
     if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test/test_ggml_rmsnorm_batch.cpp")
         add_executable(test_ggml_rmsnorm_batch test/test_ggml_rmsnorm_batch.cpp)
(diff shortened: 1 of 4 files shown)


Graffioh:
Emitter latency follow-up (commit abae692e)

Upstream behavior checked before implementation:
- vLLM buffers only request stop-string suffixes in its incremental detokenizer; reasoning/tool parsing is handled separately.
- SGLang emits clean incremental text immediately and mainly delays incomplete UTF-8.

Luce had imposed its 15-byte tool-syntax holdback even for ordinary requests with no declared tools. This follow-up makes ordinary content immediate while retaining split reasoning-marker detection; requests with tools or custom stops retain the existing conservative buffering.

Validation:
- server unit suite: 362/362 passed, including immediate short content and a split <think> opener regression
- production gfx1151 HIP server linked
- exact long-C16 Strix screen: same selected prompt hash, 54,148 prompt tokens, 16/16 valid, 64 output tokens/request
- median TTFT: 140.464 s -> 123.210 s (-17.254 s, -12.3%)
- max TTFT: 158.331 s -> 157.854 s (essentially flat)
- median visible-minus-native-prefill delay: 28.339 s -> 22.367 s

This is a one-repeat screen. Prefill order varied between runs, so the direct unit regression and visible-minus-native gap are the cleaner causal evidence. Six C16 requests became visible about 0.33 s after native prefill; ten still waited about one 22.37 s traversal because their first sampled token produced no visible text. That remaining delay is a separate scheduler/detokenizer follow-up.

PR #606, opened 2026-08-13
perf(qwen36): add SpecLA factor-capture speculative decoding

Graffioh:
## Current gain over `main`

Qwen3.6-27B Q4_K_M target, Q8_0 DFlash draft, DDTree budget 22, top-k 4, `tau=6`, ten HumanEval-style prompts, and 128 generated tokens:

| Configuration | Mean decode throughput |
|---|---:|
| Current `main` | 25.14 tok/s |
| SpecLA | 27.19 tok/s |

**SpecLA gain over current `main`: +8.2%.**

Paper: https://arxiv.org/abs/2607.16673

## CLI

The normal entry point is one flag:

```sh
dflash_server TARGET.gguf --draft DRAFT.gguf --specla
```

`--specla` enables speculative linear-attention verification when the loaded target and placement support it. SpecLA is the verification/state-management mode; it is not an alias for a particular proposal algorithm.

For the Qwen3.5/Qwen3.6 dense path implemented in this PR, the runtime currently:

- selects DDTree as the compatible proposal adapter;
- enables fast rollback;
- uses DDTree budget 22;
- uses confidence-pruning margin `tau=6`;
- uses draft-tree top-k 4.

This separation is intentional. A future Qwen3.8 DSpark implementation can provide candidates and parent topology behind the same `--specla` flag without changing the user-facing CLI. That DSpark adapter and its target-specific recurrence/factor capture are not part of this PR.

## Supported and permitted combinations

The path in this PR is supported for a monolithic Qwen3.5/Qwen3.6 dense target on one local device. It has been built and parity-tested on HIP.

The current Qwen adapter requires:

- `--draft DRAFT.gguf`;
- fast rollback, which `--specla` enables automatically;
- a non-split, single-device target placement.

Setup-dependent overrides remain available:

| Option | Constraint | Default with `--specla` | Purpose |
|---|---:|---:|---|
| `--ddtree-budget N` | positive integer | 22 | Maximum DDTree proposal nodes |
| `--ddtree-tau T` | finite, non-negative number | 6 | Cumulative log-probability pruning margin |
| `--specla-top-k K` | positive integer | 4 | Draft-tree expansion width |
| `--draft-swa N` | positive window when needed | checkpoint metadata | Override draft sliding-window metadata for older checkpoints |

[...]

PR #625, opened 2026-08-18
qwen35: DSpark speculative decoding (Qwen3.8-27B drafters)

davide221:
Adds DSpark and DFlash 2 drafter support to the qwen35 family spec-decode loop, targeting the published Qwen3.8-27B checkpoints (`RadixArk/Qwen3.8-27B-DSpark`: DFlash backbone + low-rank Markov bigram head + confidence head; `z-lab/Qwen3.8-27B-DFlash2`: DFlash backbone + grouped dynamic convs + candidate selector), plus the decode/prefill/tree optimizations listed below. Merged with current `main` (SpecLA #606, packed concurrent prefill #595); current validation is summarized below.

## What's in here

**Spec loop** (`qwen35_backend.cpp`)
- DSpark markov-corrected greedy chain (fused single-graph variant, non-fused fallback) when the drafter ships DSpark heads.
- DFlash 2 candidate selector for the greedy chain, and selector-scored DDTree candidates (branch score = logp + selector compatibility with the branch's actual parent, log-softmax-normalized per position). Env gates, default-on: `DFLASH_QWEN35_DSPARK`, `DFLASH_QWEN35_FUSED_DSPARK`, `DFLASH_QWEN35_DSPARK_TREE`, `DFLASH_QWEN35_DFLASH2_TREE`.
- Adaptive speculation policy: EMA of accepted drafts/step against the live measured spec/plain step-time ratio, plain-decode bursts below break-even (`DFLASH_QWEN35_SPEC_STEP_RATIO`, `DFLASH_QWEN35_AR_BURST`).
- Target capture layers follow the drafter GGUF's `dflash.target_layer_ids`; `dflash.mask_token_id` from the GGUF wins over the family default. Either one wrong silently destroys acceptance.
- `--draft-block-size` accepts 2..2x the checkpoint metadata (max 32). DFlash2's acceptance extrapolates past its trained block-8 horizon: block 16 is the recommended code/math config (HumanEval ten-prompt decode 143.6 -> 204.1 tok/s), byte-exact at every width because verification is exact. Widths past 16 hit a step-time cliff and stay rejected; long contexts (committed >= 8192) cap the verify width at 8, draft width untouched.

**Draft side**: DFlash 2 dynamic convs + selector in the draft graph and converter; optional YaRN rope keys parsed and honored; single-file DSpark releases convert directly.

**ggml (HIP) kernels**
- Fused DeltaNet decode: `ggml_ssm_conv_step`, raw-gate `gated_delta_net` (gate tensor `[dt_bias | A]` built by the loader, `src[9]`/`op_params[10]`, coexists with SpecLA's `src[8]`/`op_params[2]`), residual ADD+RMS_NORM+MUL fusion, in-place recurrent state. Launches per decode token 1929 -> ~1140.
- Tree-mode support for the grouped-cols GDN kernel: DDTree verify no longer falls back to the generic per-token kernel (61-196 us/layer -> ~37 us); DFS parent-state reload verified to 8.4e-7 against the generic kernel.
- 64x64 MMQ tiles for dense verify widths on RDNA (`GGML_CUDA_MMQ_SMALL_TILE`; Q4_K keeps main's 128x64), pathological mmq_x=32 tile skipped, FA vec short-KV split, legacy pool MAX_BUFFERS 256 -> 1024.
- Chunked delta-net prefill path fixed and gated off by default (`DFLASH27B_CHUNKED=1` opts in): chunk size 32 keeps the triangular solve on the fast warp kernel, `ggml_cont` on chunk slices avoids ROCm's pinned-host staging in `cublasGemmBatchedEx` (39 s -> 0.6 s per 512-token forward, verified ~1e-6). On gfx1201 the sequential fused GDN kernel still wins at ubatch 512, hence off.
- `DFLASH_PREFILL_TIMING=1` prints per-ubatch build/alloc/compute.

## Numbers (R9700 / gfx1201, ROCm 7.2, greedy, 300-token gens)

Historical rows below were measured on the pure-IQ4_XS requant. **The recommended serving target is now unsloth UD-IQ4_XS as downloaded** (13.27 GiB): vs a Q8_0 reference (300x512-chunk wikitext-2 paired `--kl-divergence`; HumanEval-164 + GSM8K-200 through the serving path) it is Q8-class (+0.7% PPL, KLD 0.018, 94.1% top-1, HE 151/164, GSM8K 177/200 = the reference), and despite a 14% slower plain forward (AR 32.3 vs pure 37.4; sub-4-bit tensors hit slower kernels) it is the FASTEST code config end to end because acceptance with DFlash2 is higher on the less-degraded target: HE-10 block-16 decode 208.1 avg / 227.8 peak, e2e 156.2 (pure: 204.1 / 146.5). Same-file llama.cpp PR#27342 + the same drafter: 54.6 decode / 47.2 e2e, so 3.8x / 3.3x.

Decode, tok/s (code / prose / mixed):

| config | code | prose | mixed |
|---|---|---|---|
| plain decode (AR) | 36.4 | 36.6 | 36.6 |
| **DFlash 2 chain (baseline)** | **111-121** | **62** | **123-124** |
| DFlash 2 + DDTree budget 12 (`--ddtree --ddtree-budget 12`) | 126 | 56 | 112 |
| DFlash 2 + selector-scored DDTree 12 | 121 | 63 | 116 |
| DSpark chain | 48 | 32 | 40 |
| upstream llama.cpp 9731ad3 (HIP, tg128) | 32.5 | | |
| hipfire (published, spec off) | 36.2 | | |

The older raw/selector DDTree results remain useful workload data. The out-of-the-box default is the plain chain at the checkpoint's block 8; `--draft-block-size 16` is the recommended code/math config, and block-8 SpecLA with budget 8 is a quality-validated alternative: repeated HumanEval aggregate decode was 163.9 versus 159.0 tok/s for block-8 chain (+3.1%). HumanEval+ was 145/164 versus 143/164, with two gains and no losses.

Prefill: 1036 / 1102 / 1038 tok/s at 512 / 2048 / 6000 prompt tokens (hipfire published 759 / 737 / 663; upstream llama.cpp 1318 / 1277 / 1123). Long context and thinking-mode numbers in the review thread; with max_tokens below `hard_limit_reply_budget` (4096) the reasoning budget hook force-closes thinking and the reply runs as plain decode.

## Production serving config

```
dflash_server Qwen3.8-27B-UD-IQ4_XS.gguf --draft qwen38-dflash2-q8_0.gguf \
  --target-device hip:0 --draft-device hip:0 \
  --draft-block-size 16 \
  --cache-type-k q8_0 --cache-type-v q8_0
```

No env vars: the tuned decode path (F32 chain checkpoints, rollback threshold 1, q8 activation memo, KV rotation off for f16/q8_0 caches) is the default as of d461bad9; `DFLASH_SINGLE_CHAIN_CHECKPOINT_F32=0` restores the legacy F16 path (and ~1.1 GiB of VRAM at 128K ctx). Deploy notes: the disk prefix cache version bump to 2 invalidates every existing `.dkv` on first start, and finite `--fa-window` silently costs long-context recall on this model (leave it off; the server warns).

Greedy verification remains target-verified. In the block-8 chain/SpecLA A/B, all ten short replies and 133/164 full-suite replies were byte-identical; the remaining floating-point path differences improved two HumanEval+ tasks and regressed none. DSpark and the 3.6-DFlash drafter keep working (the z-lab Qwen3.6-27B drafter transfers to 3.8 unchanged).

## Known limitations / follow-ups
- The disk-cache identity salt resolves the K-rotation basis from the CLI `--cache-type-k` string; configs that reach f16/q8_0 via env shorthands (`DFLASH27B_KV_K`, `DFLASH27B_KV_F16`) or a family default evaluate the placeholder string instead. All documented configs are consistent; follow-up: feed the resolved type name into the salt.
- The pre-made `incoai/Qwen3.8-27B-DFlash2-GGUF` drafter file (llama.cpp PR#27342 layout, arch `dflash`, different key/tensor names) is not loadable by this server; run `server/scripts/convert_dflash_to_gguf.py` on the z-lab checkpoint as the quickstart shows. Follow-up: accept the incoai layout via key/tensor aliases in the draft loader.
- Verify MMQ runs at ~80% of practical bandwidth (457 GB/s at widths 4-16 vs 550 for the N=1 GEMV); it already uses int8 WMMA on RDNA4. Refuted: K-loop weight-tile double buffering (occupancy), 8-warp small tiles (write-back constraint), decode-once wide MMVQ (ties MMQ at N=8).
- Tree mode still pays F32 node snapshots and un-fused conv copies (~6% of a tree step); a parent-aware fused conv step is the next tree lever.
- Confidence-gate adaptive block length is wired but not validated; the chain runs with the gate off.
- Prose acceptance is below break-even with every drafter; the adaptive policy runs prose mostly as plain decode.


jkyamog:
Bots been following this and testing it.  They have noticed a regression TP/NVLink speed has dropped significantly vs old DFlash.  They have a fix, but still cleaning and testing.

davide221:
R9700 / `gfx1201` correction and supported follow-up (`539ca43`):

The Qwen3.8 DFlash2 checkpoint declares `block_size=8`, and its published configuration uses seven speculative tokens per verification step. The earlier width-12/16 numbers were useful experiments, but those widths exceed the checkpoint's trained/published horizon and should not be production recommendations. The server now rejects `--draft-block-size` values above the drafter metadata; the flag remains available for safe down-tuning from 8.

Supported measurements on Qwen3.8-27B IQ4_XS + Q8_0 DFlash2:

| Setting | HE 10-prompt aggregate decode | HumanEval+ pass@1 |
|---|---:|---:|
| plain AR | 36.9 tok/s | — |
| block-8 chain | 159.0 tok/s | 143/164 |
| block-8 SpecLA, budget 8 | 163.9 tok/s | 145/164 |

SpecLA's gain was reproducible across repeated A/B runs (+3.1% aggregate decode). It gained two HumanEval+ tasks and lost none; all 10 short A/B replies and 133/164 full-suite replies were byte-identical.

Production flags are now:

```text
--specla --ddtree-budget 8
```

with the checkpoint metadata block size (no `--draft-block-size` override). The rebuilt `gfx1201` server passed `test_feature_gate`, rejected a live width-12 launch with the checkpoint-specific `[2, 8]` bound, reproduced 164.45 aggregate tok/s, and is serving healthily on `lucebox8`.

davide221:
R9700 pure-AR follow-up (commit `739cf84`)

- Qwen3.8-27B IQ4_XS, pure AR, same HumanEval prompt: original 8-wave kernel **37.230 tok/s** vs tuned 6-wave kernel **37.413 tok/s** in a hot-state 4-run A/B (**+0.49%**).
- The generated response was byte-identical to the original-kernel/cache baseline; four tuned repeats also produced the same SHA-256.
- A full cache sweep found `K=q8_0, V=f16` fastest: **37.21 tok/s** before the wave change, vs **36.90** for q8/q8. Disabling `LUCE_Q8_MEMO` fell to **36.65**.
- Pure-AR profiling attributes about **89%** of GPU kernel time to IQ4_XS MMVQ. Wider 64-bit weight loads and a two-block occupancy hint were neutral and were removed.
- Existing classical quantizations did not approach IQ4_XS: Q4_0 32.76, Q2_K 32.04, IQ2_XXS 30.65, Q4_K_S 29.45 tok/s. Their extra decode/unpack work costs more than their bandwidth savings on gfx1201.

This does **not** claim 40 tok/s exact AR. From the measured bottleneck, 40 needs roughly another 7% end-to-end / about 8% in the already bandwidth-bound IQ4_XS kernel; that likely requires a new simple sub-4-bit format or hardware clock/power work, not another stock GGUF quant.

davide221:
@cubic-dev-ai review

PR #628, opened 2026-08-18
fix(hip): restrict wavefront guard to device pass

pepuscz:
## Summary

Restrict the rocWMMA wavefront-size compile-time check to the HIP device pass.

Some ROCm 7.1 host passes expose `__AMDGCN_WAVEFRONT_SIZE` without making it an integer constant expression. Evaluating it in a preprocessor arithmetic condition then rejects an otherwise valid gfx1151 build. The device pass still performs the existing wave32 safety check, so wave64 targets continue to fail loudly.

Published qualification patch: [`lucebox-rocm71-host-wavefront.patch`](https://github.com/pepuscz/strix-halo-deepseek-v4-flash/blob/main/patches/lucebox-rocm71-host-wavefront.patch)

## Validation

- `git diff --check`
- ROCm 7.1.1 / gfx1151 qualified build completed with this guard
- The patch changes only preprocessor scope; kernel code and device-pass validation are unchanged

<a href="https://cubic.dev/pr/Luce-Org/lucebox/pull/628?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/src/flashprefill_kernels.hip.cu
+++ b/server/src/flashprefill_kernels.hip.cu
@@ -29,12 +29,15 @@
 // which is a 32-lane instruction. A wave64 target would need a different WMMA
 // instruction and fragment layout, not a warp-width tweak, so we fail the build
 // loudly rather than emit silently-wrong results. All currently declared AMD
-// targets (gfx1100 / gfx1151, RDNA3) are wave32. __AMDGCN_WAVEFRONT_SIZE(__) is
-// defined only in the device compile pass.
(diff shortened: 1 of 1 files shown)


davide221:
@pepuscz thanks for your contribution! lgtm

PR #637, opened 2026-08-20
NVLink fix for PR625

jkyamog:
Branch `work/625-nvlink-fix-clean` (fork of `Luce-Org/lucebox`), tracking
#625. Fixed two first-generate crashes on the 2x3090 NVLink tensor-split
bench bed that blocked every spec path. Worked on by Qwen 3.8 (Lucebox) +
DeepSeek V4 Flash, reviewed by GPT-5.6 and 0x Alpha

| crash | symptom | commit |
|---|---|---|
| AR + all configs | `GGML_ASSERT(ret.axis != GGML_BACKEND_SPLIT_AXIS_UNKNOWN)` at `ggml-backend-meta.cpp:570/556`, op=SSM_CONV, during tensor-split graph alloc | `2aed638b` |
| DSPARK fused | illegal memory access (placeholder pointer `0x2000000000000000`) in fused spec decode under tensor split | `9b968fc5` |

### Fixes

- `2aed638b` ggml-meta: split SSM_CONV step/SpecLA by axis layout, not
  `op_params`. `handle_ssm_conv` now dispatches on the actual src axis layout;
  the fused step mode returns the same conv_out state layout as the proven
  non-fused path (previously fell into `handle_generic` -> `SPLIT_AXIS_UNKNOWN`
  -> assert).
- `9b968fc5` fix(qwen35): resolve fused DSpark lm_head to draft-GPU simple
  tensor. Under tensor split the target lm_head is a meta tensor (placeholder
  data pointer); `resolve_fused_lm_head()` returns the simple tensor of the rank
  whose `layer_split_gpus[j] == draft_gpu` (lm_head is a full-copy mirror on
  every rank -> local read is valid), else nullptr -> host-chain fallback.
  Applied at both draft-backend call sites (fused greedy chain + DDTree topk) —
  the same latent leak existed in both. Partially a port of the deployed
  perf-line fix `fd7b294`.
- `997a36c4` fix(server): deepen prefix cache past the tools-head pin on
  tool-heavy turns (a prefix-cache/perf fix, not an NVLink crash). The
  inline-snap deepen decision compared the chunk-aligned restored KV position to
  the forced cut; on tool-heavy turns the logical length sits just above the
  chunk gap (e.g. logical 3620 vs restored 3584) so `forced_cut > restored`
  stayed true and the pin kept re-targeting the already-cached head — the full
  conversation was re-prefilled every turn. Now captures the logical entry
  length (`cache.prefix_len` before the restore block overwrites it with the
  snapshot position) and feeds that to the deepen decision. 9-line dataflow
  change in `http_server.cpp`.

### Verification (2x RTX 3090 + NVLink, `--target-split-mode tensor --peer-access`, Q4_0 KV, greedy, verify-width 8; prompt 24538 tok, 1024 gen)

| cell | rep1 | rep2 | accept | crash |
|---|---|---|---|---|
| AR | 36.9 t/s | 36.6 | — | 0 |
| DSPARK | 31.4 | 31.3 | 0.164 | 0 |
| DFLASH2 | **79.6** | **79.7** | **0.569** | 0 |

- DFLASH2 = 2.16x AR on NVLink, above the llama.cpp 1.7-2.0x speculative
  reference.
- AR A/B value-correct: fused vs `DFLASH_QWEN35_NO_FUSED_KERNELS=1` produce
  byte-identical completions.
- NVLink genuinely engaged: `target_split=tensor peer_access=ON`, 4x links.
- Pre-fix AR (SSM_CONV assert) and DSPARK (illegal memory access) both crashed;
  post-fix all cells complete, zero crash lines.

### Status (2026-08-24)

Rebased onto latest #625 (`qwen38-dspark @ 46b1d6a0`, +5 commits past the
earlier sync) — clean rebase, no conflicts. The two crash fixes are now
`2aed638b` + `9b968fc5`; the prefix-cache deepen fix `997a36c4` is the 3rd commit
beyond 625. Force-pushed to the fork (`feab01b` -> `997a36c4`).

Rebuilt image `lucebox-hub:pr637-sync-cuda12` (arches 86;120); `test_server_unit`
393/393 pass. Deployed to the 2x3090 service (`lucebox-3090s`, host port 8001,
tensor-parallel, `--max-ctx 262144`), healthy; the 5090 service was left on the
previous image. Live prefix-cache check confirms the deepen fix: across three
synthetic tool-heavy turns `cached_prefix` went 0 -> 512 -> 1024 and prefill
dropped 1040 -> 786 -> 380 tok; the inline snap committed at logical 1280 (past
the chunk-aligned 1024) — exactly the case the fix targets.

Open items (not blocking):
1. DSPARK 31.4 vs old perf-line 39.0 — ~20% gap (FEAT_CAP default 4096 vs 8192,
   quant/prompt mix). Speed-parity work item, tracked separately.
2. Q8_0-vs-Q6_K target bench — deferred until a Q8_0 target file is available on
   the bench host.
3. Independent axis-correctness review of the two crash fixes — in flight.
4. Confirm the prefix-cache deepen under real ~150K-token pi traffic on the
   3090s (the synthetic check used ~1.4K-token prompts).

## PR #533 — Performance update

Rebuilt and tested the rebased head `06b9600` with the same 26,758-token
OpenClaw payload and deterministic 1,024-token continuation used for the
earlier PR measurements.

| Configuration | PP | TG |
|---|---:|---:|
| RTX 5090 | **2,170.9 tok/s** | **110.6 tok/s** |
| 2x RTX 3090 TP + NVLink | **1,048.5 tok/s** | **76.2 tok/s** |

Both runs were prefix-cache misses, generated all 1,024 requested tokens, and
completed without CUDA, OOM, or server errors. The generated output for each
configuration matched its corresponding reduced-power run byte for byte.

<a href="https://cubic.dev/pr/Luce-Org/lucebox/pull/637?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/deps/llama.cpp/ggml/src/ggml-backend-meta.cpp
+++ b/server/deps/llama.cpp/ggml/src/ggml-backend-meta.cpp
@@ -825,6 +825,18 @@ static struct ggml_backend_meta_split_state ggml_backend_meta_get_split_state(
     };
 
     auto handle_ssm_conv = [&](const std::vector<ggml_backend_meta_split_state> & src_ss) -> ggml_backend_meta_split_state {
+        // Step / SpecLA variants (ggml_ssm_conv_step, ggml_ssm_conv_specla):
+        // x [C,T,S] -> out [C,T,S]; the channel axis stays axis 0, while the
+        // weight [K,C] and conv_state [K-1,C,S] carry the same channel
+        // partition on axis 1. The axis layout (not the op_params flag, whose
+        // encoding differs between the step and SpecLA variants) determines
+        // the split.
+        if (tensor->src[2] != nullptr &&
+            src_ss[0].axis == GGML_BACKEND_SPLIT_AXIS_0 &&
+            src_ss[1].axis == GGML_BACKEND_SPLIT_AXIS_1 &&
+            src_ss[2].axis == GGML_BACKEND_SPLIT_AXIS_1) {
+            return {GGML_BACKEND_SPLIT_AXIS_0, {0}, 1, {1}};
+        }
         if (src_ss[0].axis == src_ss[1].axis) {
             if (src_ss[0].axis == GGML_BACKEND_SPLIT_AXIS_0) {
                 return {GGML_BACKEND_SPLIT_AXIS_1, {0}, 1, {1}};
--- a/server/deps/llama.cpp/ggml/src/ggml-backend-meta.cpp
+++ b/server/deps/llama.cpp/ggml/src/ggml-backend-meta.cpp
@@ -830,11 +830,16 @@ static struct ggml_backend_meta_split_state ggml_backend_meta_get_split_state(
         // weight [K,C] and conv_state [K-1,C,S] carry the same channel
         // partition on axis 1. The axis layout (not the op_params flag, whose
         // encoding differs between the step and SpecLA variants) determines
-        // the split.
+        // the split. The mode is asserted after the layout match so a future
+        // tree/dynamic-conv layout collision fails loudly instead of being
+        // silently treated as Step/SpecLA.
         if (tensor->src[2] != nullptr &&
             src_ss[0].axis == GGML_BACKEND_SPLIT_AXIS_0 &&
             src_ss[1].axis == GGML_BACKEND_SPLIT_AXIS_1 &&
             src_ss[2].axis == GGML_BACKEND_SPLIT_AXIS_1) {
+            const int mode = ggml_get_op_params_i32(tensor, 0);
+            GGML_ASSERT(mode == 1 || mode == 2);
+            GGML_ASSERT(tensor->src[2]->type == GGML_TYPE_F32);
             return {GGML_BACKEND_SPLIT_AXIS_0, {0}, 1, {1}};
         }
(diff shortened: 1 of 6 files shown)


jkyamog:
Wow Dflash2 makes the NVLinked 3090s faster than a 5090 in TG.

  | Metric | Lucebox RTX 5090 (131k ctx) | 2×RTX 3090 TP + NVLink (262k ctx) |
   |---|---:|---:|
   | Prefill (49.4k tokens) | ~1,871 tok/s (26.4 s) | ~951 tok/s (51.9 s) |
   | Decode (DFlash2) | 63.0 tok/s | 68.2 tok/s |
   | DFlash2 acceptance | 60.0% (72/120) | 62.5% (80/128) |
   | Target forwards / spec step | 1.53 | 1.00 |

PR #645, opened 2026-08-22
lg-

Graffioh:
## Summary

Add LuceGraph, an opt-in service-round profiler for Qwen3.8-27B concurrency.

This PR is stacked on #642. It adds no scheduling or kernel changes. Generated captures remain local and are not committed.

## Captured data

- Record scheduler plans, prefill work, KV pressure, graph shape, padding, effective speculation tree width, and high-level phase spans.
- Track each speculative lane from eligibility through reservation, drafting, verification, acceptance, durable state, and scheduler consumption.
- Correlate queueing, admission, prefill completion, first token, token bursts, and request completion.
- Record run identity: configured Git SHA, model paths and architecture, runtime backend, concurrency and speculation settings, selected environment knobs, and wall-clock/steady-clock anchors.
- Bound every in-memory record set and report dropped records in the JSONL footer.

The disabled path passes a null profile pointer. It does not allocate, read clocks, lock, format, or write from the inference path. The profiler adds no device synchronization.

## Local workflow

Set `DFLASH_PROF=concurrency` and `DFLASH_PROF_OUT=<path>`.

By default, a clean shutdown writes JSONL. Set `DFLASH_PROF_CHECKPOINT_EVERY=<rounds>` for atomic periodic checkpoints, so a killed or hung experiment retains its last completed checkpoint. Requests still running at checkpoint time use `"ok": null` and are not counted as failures. Signal-handler flushing is intentionally omitted because the profiler uses operations that are not async-signal-safe.

`profile_report.py` reads that JSONL and writes:

- Markdown summary
- Versioned machine-readable JSON summary
- Perfetto JSON
- Folded and per-token folded stacks
- LuceGraph, a self-contained offline HTML report

Markdown warnings report observed failures, incomplete or dropped capture data, and accepted-versus-durable mismatches. The report does not impose guessed performance thresholds.

### LuceGraph

`profile_report.py --html report.html` writes one self-contained LuceGraph file that opens from `file://`. Python embeds pre-aggregated JSON, and vanilla JavaScript renders the SVG without a CDN or build step.

The phase-budget wall compares C=1 through C=5 with per-token, per-round, and wall-share normalization. The request waterfall separates queue, prefill, first-token, and decode time. The speculation funnel reports conversions and acceptance by draft position.

Roofline color applies only to target and draft compute. Tooltips expose arithmetic intensity and machine balance. Unknown devices stay neutral and show a notice. `--baseline` adds side-by-side walls, numeric deltas, and mismatch warnings.

`GET /observability` serves the dependency-free Lucebox dashboard. `GET /observability/snapshot` provides the raw local snapshot.

There is no Prometheus endpoint. JSONL, Markdown, JSON summaries, Perfetto traces, rocprof traces, and benchmark result files stay outside git.

Retention is explicitly keep-first: after `DFLASH_PROF_MAX_ROUNDS`, later rounds are counted as dropped. The capture metadata exposes the retention policy, round cap, and `StepProfile` size so memory cost is visible. A ring buffer is left out until a real debugging workflow needs it.

## Device drill-down

`harness/benchmarks/concurrency/rocprof_server_wrapper.sh` creates a local diagnostic bundle for a selected C=4 run. It enables the service capture and Qwen ROCTX, then asks rocprofv3 for marker, kernel, HIP-runtime, and memory-copy tracks as CSV and native Perfetto, plus native summaries.

Packed and speculative target paths now emit equivalent ROCTX ranges. Their marker text carries the service `round_id`, path, tree width, live slots, bucket, row count, prefill shape, and KV length. The high-level report identifies the expensive round; searching the native trace for `qwen35.graph_compute round_id=<id>` reveals the kernels, launch delays, copies, graph APIs, and real device idle gaps for that round.

`qwen35.graph_compute` is a host submission range. Device work may continue into `qwen35.argmax_readback`; rocprof's queue tracks are the device timing authority. `GGML_CUDA_GRAPH_STATS` remains an optional aggregate cross-check and is not forced by the wrapper. Occupancy and bandwidth counters remain a separate, hardware-specific PMC run.

No rocprof CSV importer or trace merger is included. The native trace already owns one coherent device clock, and the stable round ID is enough to link it to the service artifact. The CSV bundle leaves a clean path for automated evidence summaries after a real Qwen trace proves the correlation fields.

## C=4 results

On #642 head `aec43dc9`, three measured repeats found no measurable overhead.

| Version | Aggregate tok/s | Output-window tok/s | Request decode tok/s |
| --- | ---: | ---: | ---: |
| #642 | 194.81 | 350.59 | 86.34 |
| Profile disabled | 194.42 | 350.59 | 86.32 |
| Profile enabled | 194.62 | 350.53 | 86.32 |

Profile disabled was -0.20% vs #642. Profile enabled was +0.11% vs disabled. Treat both deltas as noise.

PR #642 later advanced to `0e82772b`. That exact head still fails the C=4 warmup with `fixed chain commit preflight failed` and 0/4 requests, so a current-head overhead comparison is not valid.

The rebased capture narrows that parent failure to state promotion or commit. It records 28 proposed, verified, and accepted draft tokens, then 0 durable and 0 consumed tokens.

The intended operating range for this profiler work is C=4-5. C=16 is not an acceptance gate.

## Stacking and merge order

Merge #642 first, then rebase this PR on the merged parent. Repeat the C=4 overhead measurement after #642 is stable; the previous numbers were collected on an older parent head.

## Verification

- Built `dflash_server` and `test_qwen35_roctx` with `-j "$(nproc)"`.
- Passed `test_qwen35_roctx`, `test_inference_profile`, and `test_observability`.
- Passed all 15 `test_seq_engine_contract` checks.
- Passed all 382 `test_server_unit` checks.
- Passed all 77 concurrency-tool and profile-report Python tests.
- Validated the rocprof wrapper contract with a fake profiler and passed `bash -n`.
- Validated the dashboard JavaScript and every referenced DOM ID.
- Exercised the report CLI end to end with Markdown, JSON summary, and Perfetto outputs.
- Passed `git diff --check`.

The repository-wide all-target HIP build currently stops in the unrelated `test_rms_norm_hip.cpp` because the HIP compatibility header does not define `cudaErrorNoDevice`. The scoped production and observability targets above build and pass.

The benchmark runners remain unmodified: traced runs are diagnostics and should not be used for throughput deltas. The standalone wrapper creates the local device bundle for the selected concurrency experiment.

<a href="https://cubic.dev/pr/Luce-Org/lucebox/pull/645?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/harness/.gitignore
+++ b/harness/.gitignore
@@ -1,4 +1,5 @@
 .harness-work/
 results/
 *.json
+!benchmarks/concurrency/device_specs.json
 __pycache__/
--- a/harness/benchmarks/concurrency/README.md
+++ b/harness/benchmarks/concurrency/README.md
@@ -5,6 +5,10 @@ prefill and concurrent decode. It has two user-facing runners that share the
 streaming client, deterministic prompt generation, and summary tooling:
 the paired ragged workload runner and the canonical-suite runner.
 
+To drill from a slow high-level service round into kernels, launches, copies,
+graph replay/capture, or device idle gaps, use the local rocprof workflow in
+[Concurrency observability](../../../server/docs/CONCURRENCY_OBSERVABILITY.md#drill-into-one-target-round).
+
 ## Canonical and blog workloads
 
 The synthetic ragged profiles below isolate serving mechanics. To measure how
--- a/harness/benchmarks/concurrency/concurrent_benchmark.py
+++ b/harness/benchmarks/concurrency/concurrent_benchmark.py
@@ -17,7 +17,7 @@ from pathlib import Path
 from typing import Any
 
 PromptInput = str | list[dict[str, str]]
-DEFAULT_CLIENT_LEVELS = (2, 4, 8, 16)
+DEFAULT_CLIENT_LEVELS = (1, 2, 3, 4, 5)
 
 
 def sha256_text(text: str) -> str:
--- a/harness/benchmarks/concurrency/device_specs.json
+++ b/harness/benchmarks/concurrency/device_specs.json
@@ -1,0 +1,35 @@
+{
+  "devices": {
+    "gfx1151": {
+      "name": "Strix Halo",
+      "mem_bw_gbps": 256,
+      "fp16_tflops": 60,
+      "note": "Published peak FP16 vector throughput and shared-memory bandwidth. Bandwidth is shared with the CPU.",
+      "source_urls": [
+        "https://www.amd.com/en/products/processors/desktops/ryzen/ryzen-ai-halo/ryzen-ai-max-plus-395.html",
+        "https://ir.amd.com/news-events/press-releases/detail/1270/amd-expands-ai-leadership-across-client-graphics-and-software-with-new-ryzen-ryzen-ai-and-amd-rocm-announcements-at-ces-2026"
+      ]
+    },
+    "gfx1201": {
+      "name": "R9700",
+      "mem_bw_gbps": 640,
+      "fp16_tflops": 95.7,
+      "note": "Published peak FP16 vector throughput and memory bandwidth.",
+      "source_urls": [
+        "https://www.amd.com/content/dam/amd/en/documents/partner-hub/radeon-pro/radeon-ai-pro-r9700-datasheet.pdf"
+      ]
+    }
+  },
+  "models": {
+    "Qwen3.6-27B-Q4_K_M.gguf": {
+      "weight_bytes": 16800000000,
+      "active_params": 27000000000,
+      "kv_bytes_per_token_per_seq": 18432,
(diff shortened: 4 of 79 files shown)


Graffioh:
Superseded by Luce-Org/luce_box#63, which packages LuceGraph with an on-demand profiling skill.

PR #654, opened 2026-08-24
feat(qwen35): support sampled concurrent speculation

Graffioh:
## Stack

This draft PR is built on top of #651 at `f1c85e05`. It therefore contains the exact current #642 history, #651's projection batching, and the separate replay-log rename.

#654 adds two commits:

1. `0445b161` — `feat(sampler): add distribution-preserving spec verification`
2. `e0647848` — `feat(qwen35): speculate sampled concurrent lanes`

Both commits are patch-equivalent across the final restack.

## Summary

- Keep temperature, top-p, top-k, repetition penalty, frequency penalty, and presence penalty lanes in DFlash2 chain rounds.
- Share one target forward across greedy speculative lanes, sampled speculative lanes, and autoregressive fallback lanes.
- Apply distribution-preserving speculative accept/reject sampling.
- Keep one transactional RNG state per slot and publish it only after prefix promotion succeeds.
- Roll back at the first rejected draft token.
- Commit only the accepted KV, target-feature, GDN, and convolution prefix.
- Batch sampled target-logit transfers by depth, stop transfers after a lane rejects, and reuse sampler buffers.
- Preserve the sparse ordinary AR path for lanes that cannot use the sampled protocol.

For a proposed token `x`, verification accepts with `min(1, p(x) / q(x))`. At the first rejection, it samples from normalized `max(p - q, 0)`. If the chain fully accepts, it samples the bonus token from the target distribution on the final verifier row. Target and residual distributions use the same sampler policy as ordinary generation.

## Three-lane reference benchmark

These measurements were taken on the previous stacked head `13590205`, before the latest #625 refresh. They remain useful as a reference for the unchanged #654 patches, but they are not presented as final-head measurements. A clean rerun was deferred because another user continuously occupied the shared GPU host; contaminated numbers were excluded.

Setup:

- Radeon R9700 (`gfx1201`)
- Qwen3.8-27B Q4_K_M target
- Q8_0 DFlash2 draft with W8 and SWA 2048
- Concurrency 3, context 4,096, and a 12,288-token KV pool
- 96 output tokens per lane with `DFLASH_MIN_TOKENS=96`
- One warmup batch, followed by three measured batches per mode
- Aggregate throughput is 288 completion tokens divided by batch wall time

Sampled lanes use temperature 0.8, top-p 0.9, top-k 40, frequency penalty 0.4, and presence penalty 0.2. The mixed batch is greedy, sampled, greedy.

| Lanes | DFlash2 | One-token AR control | Speedup |
[...]

PR #655, opened 2026-08-24
feat(server): add prefix cache to paged concurrency (qwen AR)

Graffioh:
## Summary

- Enable copied in-memory prefix checkpoints for Qwen continuous batching with paged attention. Full-prefill and disk snapshots remain disabled in paged mode.
- Keep token matching, boundary selection, pinning, LRU, and hit accounting in PrefixCache while Qwen owns checkpoint payloads, page-table restore, and recurrent state.
- Introduce an opaque checkpoint/ticket protocol and a move-only capture transaction so the scheduler never manipulates backend snapshot storage.
- Gather and scatter logical K/V rows through fresh sequence-owned pages, copy the matching recurrent slab, coalesce consecutive physical blocks, and publish replacements atomically.

## Correctness review

The pre-draft review fixed these lifecycle hazards:

- stale restore cleanup now happens before busy or failed admission returns;
- unsupported engines use cold admission without deleting a valid checkpoint;
- malformed or mismatched events can abort only their own capture ticket;
- fatal step failures discard the pending destination instead of preserving potentially mutated state;
- failed replacement leaves the incumbent checkpoint intact;
- cache hits and MRU promotion are recorded only after a validated restore;
- exact-prompt checkpoints cannot mask a shorter restorable prefix when next-token logits are unavailable;
- tensor type, contiguity, extents, vector topology, recurrent slab shape, and per-layer tensor pairs are validated before destination writes.

## Verification

- HIP Release build of dflash_server
- test_server_unit: 418 passed
- test_seq_slot_manager: 186 checks
- test_parallel_prefix_txn: 21 checks
- test_seq_engine_contract: 19 checks
- test_recurrent_snapshot: 2 passed
- git diff --check

## Draft gates

- Measure capture and restore pause time plus snapshot memory at realistic long-prefix sizes while decode traffic is live.
- Run real Qwen GPU parity for a partial final page and a forced mid-message boundary, restoring into non-identity fresh pages.
- Decide the copied-checkpoint byte budget before broad production enablement.

This PR targets main as the AR ownership/mechanism base. Open speculative concurrency PRs #642 and #654 overlap in SeqEngine, Qwen, scheduler, and target-graph files; they should consume this opaque prefix-store boundary during their rebase rather than duplicate snapshot ownership.

<a href="https://cubic.dev/pr/Luce-Org/lucebox/pull/655?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/props-endpoint.md
+++ b/docs/specs/props-endpoint.md
@@ -363,15 +363,36 @@ enabled, fields carry the runtime configuration:
 
 ```json
 "prefix_cache": {
-  "capacity":      0,
-  "in_use":        0,
-  "lifetime_hits": 0
+  "capacity":                       0,
+  "in_use":                         0,
+  "lifetime_hits":                  0,
+  "agent_turn_enabled":             false,
+  "max_resident_bytes":             4294967296,
+  "resident_bytes":                 0,
+  "budget_skips":                   0,
+  "capture_attempts":               0,
+  "capture_failures":               0,
+  "capture_stall_ms_total":         0.0,
+  "capture_stall_ms_max":           0.0,
+  "restore_attempts":               0,
+  "restore_invalidations":          0,
+  "restore_stall_ms_total":         0.0,
+  "restore_stall_ms_max":           0.0
 }
 ```
 
 The inline prefix cache (system-prompt KV reuse). Same atomic /
 non-strictly-consistent semantics as `full_cache` (§4.7).
-`capacity = 0` means the cache is disabled.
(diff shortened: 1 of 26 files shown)


Graffioh:
@cubic-dev-ai review this PR

PR #651, opened 2026-08-23
perf(qwen38): batch draft projections across lanes for concurrency

Graffioh:
## Stack

This PR is restacked on the current #642 head at `f45f874e`, followed by two #651 commits:

## Summary

- Batch shared draft dense projections across active lanes.
- Keep convolution, cache updates, RoPE, masks, positions, attention, and draft KV state lane-local.
- Use one builder for C1 through C6, without a runtime switch or duplicate execution path.
- Include the backend in the cached graph replay key.
- Inherit #642's metadata-preserving DFlash2 sliding-window handling. The tested Qwen3.8 draft reports `SWA layers: 5/5 (window=2048)`.

The previous path ran each lane's dense draft backbone in a separate graph. The packed path combines projection columns and removes repeated MMQ launches without sharing lane state.

## Projection A/B

This matched A/B isolates the unchanged #651 projection patch against its #642 parent. Warm medians exclude the first speculative round; each run generated a fixed 64-token output.

| GPU | C | Draft compute, ms | Complete round, ms |
| --- | -: | ---: | ---: |
| R9700 | 1 | 5.864 -> 5.864 (-0.0%) | 62.161 -> 62.107 (-0.1%) |
| R9700 | 2 | 11.622 -> 8.498 (-26.9%) | 74.153 -> 70.951 (-4.3%) |
| R9700 | 3 | 17.424 -> 11.392 (-34.6%) | 96.811 -> 90.871 (-6.1%) |
| Strix Halo | 1 | 12.129 -> 12.120 (-0.1%) | 137.892 -> 138.107 (+0.2%) |
| Strix Halo | 2 | 24.039 -> 15.341 (-36.2%) | 166.877 -> 158.065 (-5.3%) |
| Strix Halo | 3 | 36.378 -> 19.665 (-45.9%) | 224.446 -> 209.150 (-6.8%) |

C1 is neutral. C2 and C3 reduce complete-round latency by 4.3% to 6.8%. The R9700 C3 rocprof trace reduced MMQ launches from 336 to 196 (-41.7%) and MMQ device time from 11.938 ms to 6.103 ms (-48.9%).

## Lucebox versus llama.cpp: generic-prompt concurrency baseline

[...]

PR #659, opened 2026-08-25
perf(qwen38): various DFlash2 optimizations concurrency C1-C5

Graffioh:
Depends on #651 (`codex/draft-projection-batching`). GitHub keeps this PR based on `main` because #651 comes from a fork; after #651 lands, the visible diff reduces to this follow-up stack.

## Summary

This draft contains six independently reviewable mechanisms:

1. batch fixed-width append projections across active lanes before shared `fc`, `wk`, and `wv` matrix multiplications
2. pack normalized columns before shared attention and MLP dynamic-convolution coefficient projections
3. tile the hot GDN replay transition loads in shared memory for state size 128, replay width 257, and at most eight replay tokens
4. skip redundant Q/K row repetition on the generic active GDN route, whose kernel already broadcasts those heads
5. use an exact five-lane graph for C5 draft computation while retaining the six-lane target graph for numerical stability
6. retain stable paged target graphs and their metadata arena across rounds, invalidating them when owner, backend, or topology identity changes

The first four mechanisms are the measured C1-C4 stack previously described here. The last two are profiling-led follow-ups: one targets the important C5 case, and one removes repeated host graph construction across C1-C5.

Lane-local positions, masks, cache writes, RoPE, attention, dynamic-convolution history, accepted prefixes, active slots, recurrence order, and transaction synchronization remain unchanged. Dummy lanes stay after real lanes and their proposals are discarded.

## C1-C4 mechanisms

Qwen3.8 27B IQ4_XS target, DFlash2 Q8 width 8, Q8 KV, HumanEval raw prompts, ten fixed-C waves, and 256 forced output tokens on one R9700:

| C | Append draft p50 | Dynamic-conv draft p50 | GDN state-promotion p50 | Active-repeat target p50 |
| ---: | ---: | ---: | ---: | ---: |
| 1 | -0.33% | -0.06% | -37.58% | -0.53% |
| 2 | -8.28% | -3.14% | -39.11% | -1.22% |
| 3 | -12.60% | -4.19% | -37.65% | -0.68% |
| 4 | -14.38% | -5.84% | -34.62% | -1.09% |

| C | Append goodput | Dynamic-conv goodput | GDN replay goodput | Active-repeat goodput |
| ---: | ---: | ---: | ---: | ---: |
| 1 | -0.36% | +0.79% | +0.54% | +0.29% |
| 2 | -0.03% | +0.23% | +0.82% | +0.99% |
| 3 | +0.94% | +0.43% | +0.99% | +0.41% |
| 4 | +0.61% | +0.47% | +1.40% | +0.88% |

Each column is the causal comparison for that mechanism against its immediate baseline; the percentages are not additive.

Launch accounting remains mechanism-specific:

- append packing removes 22 projection dispatches per added lane: 0, 22, 44, and 66 fewer per round at C1-C4
- dynamic-convolution packing removes a net ten dispatches per added lane: 0, 10, 20, and 30 fewer per round
- replay tiling changes no launch count; its gain comes from lower repeated transition-load work inside the same 48 launches
- active-repeat removal removes a fixed 96 graph launches per speculative round

On Strix Halo, append packing preserved complete outputs and changed goodput by -0.45%, +0.18%, +0.95%, and +1.08% at C1-C4. Dynamic-convolution packing passed correctness and capture integrity there, but its timing was inconclusive because the unchanged C1 control moved by 8.65%. No Strix performance claim is made for the GDN or new follow-up mechanisms.

## Exact C5 draft shape

The existing graph bucketing padded C5 draft work to six lanes. Specializing only the draft graph produced this A/B/A result:

| Metric | vs baseline before | vs baseline after | bracket mean |
| :--- | ---: | ---: | ---: |
| goodput | +1.226% | +1.435% | +1.331% |
| draft-compute p50 | -10.601% | -10.746% | -10.674% |

Baseline goodput drift was -0.206%, so the bracket-mean gain clears 2x measured drift. Complete output digests match. The matched kernel capture measured 125.002 fewer speculative dispatches per round. The target graph intentionally remains at width six; specializing target and draft together was not promoted.

## Stable target-graph retention

After the C5 launch reduction, profiling showed repeated host-side `target_graph_build` work rather than a remaining launch-count bottleneck. Retaining a graph while its owner/backend/topology identity remains stable produced:

| C | Goodput vs A/B/A bracket mean | A-vs-A2 drift | `target_graph_build` p50 |
| ---: | ---: | ---: | ---: |
| 1 | +0.993% | +0.209% | -99.983% |
| 2 | +0.511% | -0.233% | -99.985% |
| 3 | +0.424% | +0.091% | -99.979% |
| 4 | +0.392% | -0.058% | -99.977% |
| 5 | +0.267% | -0.064% | -99.983% |

Complete outputs match at every C. Total `target_graph_build` time fell 89.70-92.12%. A duration-based analyzer inferred reuse on 95.13-95.25% of rounds by classifying phases below 50 microseconds; the capture format has no direct cache-hit marker.

This is a small host-side gain, not a kernel or launch-count claim: the C5 speculative dispatch count was identical before and after, and kernel-family device-time changes stayed within 0.59%. The minimum all-C gain clears the maximum measured A/A2 drift at 1x but not 2x. C6+ and memory scaling across multiple backend/owner identities were not measured.

## Rejected follow-up

Asynchronous metadata upload reduced `metadata_upload` p50 by 70.23-85.35%, but the wait moved into target compute. `target_compute` p50 regressed by 0.143-0.588%, and C1-C5 goodput remained mixed from -0.069% to +0.460%. No code from that candidate is retained.

## Correctness and verification

- full `dflash_server` build passes on the combined head
- Qwen target graph-chain tests: 28 passed
- paged-attention tests: 15 passed
- recurrent tests: 4 passed
- `test_gdn_replay_log` covers the owned tile, scalar/grouped variants, generic fallback, padded and permuted slots, and transaction preflight
- complete output digests match for every retained causal comparison
- fixed-C captures report zero drops and preserve lower-C drain accounting
- combined `#651..HEAD` ancestry and `git diff --check` pass

The retrospective C1-C5 comparisons, metrics, analyzer, decision log, and hashes are in [LuceForge PR #73](https://github.com/Luce-Org/luce_box/pull/73), under [`qwen38-dflash2-pr659-c1-c5-observations`](https://github.com/Luce-Org/luce_box/tree/codex/lucebox-optimize-luceloop/luce_forge/luce_loop/campaigns/qwen38-dflash2-pr659-c1-c5-observations). They are recorded as observations rather than backfilled accepted trials because no Luceloop draft existed before these edits.

### DIFF
--- a/docs/handoffs/pr651-draft-batching-performance.md
+++ b/docs/handoffs/pr651-draft-batching-performance.md
@@ -1,0 +1,78 @@
+# PR 651 draft batching performance follow-ups
+
+> Historical planning note: PR 659 implements the batched append projections and
+> packed dynamic-convolution coefficient projections described below. Keeping draft
+> hidden states on the device was evaluated but is not part of PR 659; the existing
+> host handoff remains intentional until a separately measured change replaces it.
+
+## Scope
+
+PR 651 packs the main drafter projections across concurrent lanes. It keeps each lane's cache writes, RoPE, masks, and attention separate.
+
+This note covers three follow-up optimizations that do not belong in the ownership cleanup:
+
+- Batch the append projections.
+- Pack the dynamic-convolution coefficient projections.
+- Keep draft hidden states on the device through chain selection.
+
+Treat each item as a measured change. Do not combine all three into one patch.
+
+## Keep these invariants
+
+- Keep `build_draft_kv_steps()` as the shared C1-C6 implementation.
+- Preserve lane-local positions, masks, cache writes, RoPE, attention, and dynamic-convolution history.
+- Rebuild a cached graph when the backend or the ordered lane-state pointers change.
+- Keep C1 output equivalent to the existing single-lane graph.
+- Keep dummy lanes after real lanes, and discard dummy proposals.
+
+## Batch the append projections
+
+`draft_kv_batch_build()` currently calls `build_draft_kv_append()` once per lane. Each call runs `draft_fuse_features()`, then the per-layer `wk` and `wv` projections.
+
+Pack every lane's `ap_feat` columns before those shared-weight matrix multiplications. Split the projected columns before RoPE and `ggml_set_rows()`, because positions, destination rows, and caches remain lane-local.
+
+The packed append path must preserve the fixed `a_step` width. Padded append rows must still write only to each lane's trash slot.
+
+Suggested shape:
+
+```cpp
+bool build_draft_kv_appends(
+    ggml_context * ctx,
+    ggml_cgraph * gf,
+    const DraftWeights & weights,
+    const std::vector<DraftKvAppendLane> & lanes);
+```
+
+Pass one lane from the normal graph and C1-C6 lanes from the batched graph. Delete the old singular builder after migrating both callers.
+
+Prove the change with the existing multilane output comparison. Add cases with zero, partial, and full append counts so padding and trash-slot writes are covered. Measure draft compute separately at C1, C2, C3, C5, and C6.
+
+## Pack dynamic-convolution coefficient projections
+
+`build_draft_kv_steps()` calls `draft_dyn_conv_kernel()` once per lane for both attention and MLP. The function projects normalized hidden columns with shared weights. The temporal convolution in `draft_dyn_conv_apply()` is lane-local and must remain separate.
+
+Pack the normalized columns before the coefficient projection. Slice the projected coefficients back into lanes, then call `draft_dyn_conv_apply()` per lane.
+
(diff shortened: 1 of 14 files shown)


Graffioh:
Refreshed onto #651 head 6ab5d01b via merge commit 82a34735. The exact measured follow-up commits remain ancestors, and the #651-relative product diff is unchanged at six files (565 additions, 117 deletions). The stale NVIDIA CI target fix now applies to both PRs.

PR #691, opened 2026-09-02
refactor(server): seal backend construction behind a plan

Graffioh:
## Summary

First PR in the server-refactor stack. It makes backend construction consume one model-aware, normalized plan and gives the resulting values stable, read-only groupings.

- make `BackendArgs` own operator input, including model and draft paths
- reduce server admission input to a narrow `BackendAdmissionContext`
- resolve GGUF metadata, placement compatibility, warnings, and SpecLA fallback once in `prepare_backend()`
- make `create_backend(const BackendPlan&)` the only backend-construction API
- make retained architecture and layer-split configs own their path storage
- group effective values into model, placement, cache, speculation, and execution views
- route the effective SpecLA mode through the plan and `Qwen35Config` instead of process environment
- use the same normalized values for construction, startup logging, tokenizer setup, cache policy, and `/props`
- keep `ServerConfig` and mutable `BackendArgs` out of backend construction

```text
operator input + server admission facts
                 |
            prepare_backend
                 |
        read-only BackendPlan
      / model     placement \
 cache  speculation        execution
                 |
     construction + server reporting
```

## Ownership and boundary

Persistent backend configs use `std::string` for required paths and `std::optional<std::string>` for optional draft paths. Configs are moved into backend members, and `.c_str()` is used only at immediate C API boundaries. Constructed backends therefore do not borrow path storage from `BackendArgs` or `BackendPlan`.

`BackendPlan` is still the effective configuration source shared by construction and the server. This keeps normalization consistent without adding a second reporting snapshot. Such a snapshot should be introduced only if it can replace this dependency rather than duplicate the same state.

SpecLA has a single authoritative decision path: `--specla` (or `BackendArgs::specla_mode` for programmatic callers) is normalized once in `prepare_backend()`, stored on the plan, and delivered to the backend through `Qwen35Config::specla_mode`. `DFLASH_SPECLA` is gone as an activation channel — `create_backend` no longer mutates process environment, and graph code keys off the SpecLA factor buffers the plan decision allocates. `DFLASH_SPECLA_TOPK` remains a read-once input default for `specla_top_k`; `DFLASH_SPECLA_CONDITIONAL_DRAFT` and `DFLASH_SPECLA_FUSED_COMMIT` stay as algorithm/debug controls.

## Stack

This is PR 1 of 2:

1. **#691 — backend planning, owned configs, and grouped effective values** (this PR, head `915bddfe`)
2. **#692 — owned generation lifecycle and runtime owner** (head `2e6309477`)

[...]

PR #693, opened 2026-09-03
refactor(server): group backend plan by concern

Graffioh:
## Summary

Second slice of the server refactor stack, following #691.

- replace the flat `BackendPlan::args()` view with read-only snapshots by concern
- keep model path and inspected GGUF metadata together in `Model`
- separate `Placement`, `Cache`, `Speculation`, and generic `Execution` values
- put the four current DeepSeek-specific construction values in a small explicit `DeepSeek4` group
- consume mutable `BackendArgs` once at preparation, after normalization and admission
- make backend dispatch and server startup/reporting read the same grouped plan
- remove the unused server-admission parameter from inert-feature warning collection

`Execution` now contains only `stream_fd` and `chunk`; it is no longer a miscellaneous model-path and DeepSeek bag. No architecture hierarchy or variant is introduced.

## Stack

This is PR 2 of 3:

1. #691 — backend planning and owned architecture configs, head `d6ef2f7cf`
2. **#693 — grouped effective backend plan** (this draft, single review commit `90062dc21`)
3. #692 — owned generation lifecycle, head `66b9295bb`

This branch is directly based on the current #691 head. Because the branches live in a fork, the PR targets `main` and temporarily displays the preceding stack commits.

Review only this PR's delta: https://github.com/Graffioh/lucebox-hub/compare/codex/backend-plan-boundary...codex/backend-plan-structure

The following #692 delta is: https://github.com/Graffioh/lucebox-hub/compare/codex/backend-plan-structure...codex/luce-engine-implementation

## Resulting boundary

```text
BackendArgs (mutable input)
        |
  prepare_backend (consume once)
        |
BackendPlan {
  model,
  placement,
  cache,
  speculation,
  execution,
  deepseek4
}
   /          \
factory     read-only startup/reporting
```

The grouped field carriers are public read-only views; only the private builder can populate the plan. Persistent architecture configs introduced in #691 own their path storage, so the plan and returned backend have independent lifetimes.

## Why this shape

- `Model` couples the path with the metadata inspected from that path.
- `Execution` remains small and architecture-neutral.
- `DeepSeek4` is a simple field group, not a hierarchy or variant. A variant becomes useful only if more families accumulate distinct option sets.
- `BackendPlan` remains the one effective launch snapshot for construction and startup reporting.

A separate `BackendRuntimeInfo` is deferred until it can replace reporting state rather than duplicate it. Explicit SpecLA configuration is also deferred because removing the current environment bridge would be a broader behavior change across low-level graph helpers.

## Tests

- HIP/gfx1151 builds: `dflash_server`, `test_feature_gate`, `test_server_unit`, `test_deepseek4_unit`, `test_generate`, and `test_dflash`
- 449/449 model-free server and plan/feature tests pass
- compile-time checks verify the direct `ModelBackend` return type and absence of a flat `args()` view
- focused tests cover owned model data and the explicit DeepSeek option group
- `git diff --check`

### DIFF
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -541,6 +541,7 @@ add_library(dflash_common STATIC
     src/common/backend_precision.cpp
     src/common/daemon_loop.cpp
     src/common/gguf_inspect.cpp
+    src/common/backend_plan.cpp
     src/common/backend_factory.cpp
     src/common/feature_gate.cpp
     src/placement/placement_config.cpp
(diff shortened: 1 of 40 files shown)


Graffioh:
Incorporated into #691 as commit 6c9abb09d. #692 is now rebased directly onto the combined #691 endpoint, leaving a two-PR server-refactor stack.

PR #701, opened 2026-09-04
feat(moe): cold-owner None, and expert-major prefill for a hot-only stack

maikzz32:
Two commits, one enabling change and the performance path it unlocks.

## 1. `MoeHybridColdBackend::None`

`MoeHybridColdBackend` has `Cpu` and `Gpu`: the experts a rank does not hold hot are evaluated *somewhere in this process*. There is no way to say "the non-resident experts are not mine at all — their contribution arrives from outside". Expert-parallel inference across processes needs exactly that: each owner evaluates its own experts and a reduction outside ggml sums the partials.

With `None`:

* storage allocates no cold buffers and never materializes or swaps a cold expert;
* the evaluators build no cold graph and never fall back to CPU or streamed evaluation. A route whose id was masked to `-1` contributes zero instead of being routed to a cold owner.

Purely additive: the enum grows a third value that nothing selects by default, and every `Cpu`/`Gpu` path is untouched.

Also adds `eval_moe_shared_expert_batched`, the piece such an owner needs that no existing entry point provides: the shared expert alone, batched. An owner that reduces across processes has to keep the shared term **out** of the reduced partial — it is replicated, so summing it would count it once per owner — and add it afterwards, which means evaluating it without the routed experts.

## 2. Expert-major packing for a hot-only prefill batch

An owner with cold-owner `None` holds a **reduced** expert stack, and `mmq_full_batch_ok` is false for a reduced stack because MMQ's `mul_mat_id` illegal-accesses on it. `eval_moe_hybrid_ffn_batched` therefore falls into its sub-batch loop, whose size is

```cpp
min(mmq_safe_sub_batch(), moe_hybrid_prefill_hot_sub_batch_limit())
```

which is **1** on gfx1151, where `mmq_safe_sub_batch()` returns 1 for compute < sm_80. A 1517-token prefill then runs 1517 × 43 = 65k graph computes.

With cold-owner `None` there is exactly one owner and every route that survives masking is resident, so the whole batch can take the packing the heterogeneous prefill already uses: one expert-major graph per layer via `eval_moe_owner_expert_major_batched`. That path was written for a reduced stack in the first place, which is why it does not need the MMQ full-batch guarantee.

### Benchmark

Two Ryzen AI Max 395 (Radeon 8060S, gfx1151, ROCm 10.0.0), DeepSeek V4 Flash ROCmFP2, `--ds4-prefill sparse --chunk 2048`, same binary, same clocks, no other load. Prefill wall time:

| prompt | before | after |
|---|---|---|
| 1517 tokens | 34.1 s | **11.45 s** (FFN 25.9 s → 3.7 s) |
| 12017 tokens | 283.9 s | **96.9 s** |

### Output parity

The completion is unchanged by the second commit, and the 128-token greedy benchmark stays byte-identical to a single-node run of the same binary (sha256 `87964cbd…`), 3 runs.

Reachable only with cold-owner `None`, so no existing configuration changes behaviour. On sm_80+ the sub-batch is 8 rather than 1, so the same shape would have been 8× less bad but still 8k graph computes.

Hardware note: gfx1151, not the sm_86+ hardware in CONTRIBUTING. Happy to have the numbers re-run on your 3090 or 395 box.

Used downstream to run DeepSeek V4 Flash expert-parallel across two hosts, each holding half the routed experts.

<a href="https://cubic.dev/pr/Luce-Org/lucebox/pull/701?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/src/common/moe_hybrid_ffn_eval.cpp
+++ b/server/src/common/moe_hybrid_ffn_eval.cpp
@@ -3542,6 +3542,51 @@ bool eval_moe_hybrid_ffn_batched(
                             : storage.gate_cold    ? (int)storage.gate_cold->ne[2]
                             : 0;
     const bool cold_on_gpu = storage.cold_backend_kind == MoeHybridColdBackend::Gpu;
+    // Cold owner None (a cluster rank): every route that survived masking is
+    // resident here and there is no second owner, so the whole batch can be
+    // packed by expert into ONE graph per layer. Without this a reduced hot
+    // stack falls into the sub-batch loop far below, whose size is
+    // min(mmq_safe_sub_batch(), prefill limit) = 1 on gfx1151 - one graph per
+    // token per layer. Measured on a 1517-token prompt: 25.9 s of FFN against
+    // 7.6 s for a single node's whole prefill graph. Expert-major packing is
+    // also what keeps the reduced stack off the MMQ full-batch path that
+    // mmq_safe_full_batch=false exists to avoid.
+    const bool hot_only_expert_major =
+        !expert_compute &&
+        storage.cold_backend_kind == MoeHybridColdBackend::None &&
+        !storage.gate_cold && !storage.gate_up_cold && !storage.down_cold &&
+        n_hot_stack > 0 &&
+        moe_expert_major_prefill_enabled(n_tokens);
+    if (hot_only_expert_major) {
+        static std::once_flag logged;
+        std::call_once(logged, [n_tokens, n_hot_stack] {
+            std::fprintf(stderr,
+                         "[hybrid-ffn] hot-only expert-major batch active tokens=%d "
+                         "stack=%d (no cold owner)\n",
+                         n_tokens, n_hot_stack);
+        });
+        const auto wall_t0 = HybridClock::now();
+        std::string owner_err;
+        const bool ok = eval_moe_owner_expert_major_batched(
+            gpu_backend, cfg, desc,
+            storage.gate_hot, storage.up_hot, storage.down_hot,
+            storage.gate_up_hot, storage.hot_local_by_global,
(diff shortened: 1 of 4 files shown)


davide221:
Pushed onto your branch: merged main (the conflict was just the `dflash::common` → `luce::common` rename) and fixed the cubic findings:
- `None` now implies no cold materialization (`MoeHybridConfig::materializes_cold_experts()`), so storage construction no longer fails when the flag keeps its default of true.
- A `None` owner's routed partial never includes the shared expert. The hot-only prefill passes `include_shared=false`, and every evaluator refuses a desc that still has shexp tensors instead of silently counting the shared term once per owner.
- `eval_moe_shared_expert_batched` now checks `n_tokens` before sizing its output.
- The `server/src/cluster/` comment now points at `ggml_cluster_allreduce`.

Unit tests are in `test_moe_hybrid_storage`. The full `test_server_unit` suite passes (597/597) on lucebox4 (HIP).

davide221:
Addressed cubic's second pass in d4052bbf5:
- **P1:** `MoeHybridStorage::streams_cold_experts()` excludes `None`, so DS4 no longer sets up the cold stream engine for it, and the paged-serving check now keys on streaming.
- **P2:** the log now reports `cold_backend=none`.
- **P3:** `eval_moe_shared_expert_batched` rejects a null backend, and the test no longer has the dead `-1` route.

`test_server_unit` passes 598/598 on lucebox4.

davide221:
@cubic-dev-ai review

PR #703, opened 2026-09-04
feat(moe): register mixed-qtype decode tables for hot-only storage

maikzz32:
> **Stacked on #701**, which adds `MoeHybridColdBackend::None`. Please review that one first — the diff against `main` shown here contains both. The change belonging to this PR is the last commit, 16 lines in `deepseek4_loader.cpp`.

Adaptive (mixed ROCmFPX) experts keep their codebooks out of band and need a decode table registered per resident tensor. `register_deepseek4_moe_hybrid_mix_tables` demanded a materialized GPU cold owner, so an owner using cold-owner `None` — whose resident set is exactly its hot experts — was refused.

That refusal was unnecessary. The registrar already handles this shape: `ds4_register_compact_mix_tensor` returns success for a null tensor whose expert-id list is empty, which is precisely how an absent cold owner presents itself. Only the guard was too narrow. It now accepts either a materialized GPU cold owner, as before, or hot-only storage with no cold owner; a `Cpu` cold owner is still refused, and nothing else changes.

This also resolves the `TODO(cluster-verify)` I left beside `ds4_cluster_has_mix_experts` in my own tree.

## Why it matters

It is what makes an adaptive artifact usable by an owner whose cold set lives outside the process. Measured downstream on two Radeon 8060S nodes under ROCm 10 with DeepSeek V4 Flash, each node holding half the routed experts with cold owner `None`: the adaptive artifact loads where it previously failed with `mixed expert qtypes require materialized GPU owners`.

Against the uniform artifact on the same two nodes, verify width 4, output byte-identical to a single node:

| artifact | benchmark prompt | free prompt | exact-copy fidelity |
|---|---|---|---|
| `ROCMFP2-STRIX` (uniform) | 38.4 tok/s | 17.2 tok/s | 12/60 |
| `ROCMFPX-MIX-STRIX` (adaptive) | **42.3 tok/s** | **20.6 tok/s** | **60/60** |

Fidelity is the exact-copy protocol from `server/docs/DS4.md` — 20 identifiers × 3 repeats at temperature 0. The adaptive artifact is not merely faster here; it is the one that answers correctly.

## Verification

Built against this PR's base with `-DDFLASH27B_GPU_BACKEND=hip -DDFLASH27B_HIP_ARCHITECTURES=gfx1151 -DDFLASH27B_ROCMFP2_AFFINE=ON -DGGML_HIP_GRAPHS=ON` on a gfx1151 host: 379/379, no new warnings.

<a href="https://cubic.dev/pr/Luce-Org/lucebox/pull/703?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/src/common/moe_hybrid_ffn_eval.cpp
+++ b/server/src/common/moe_hybrid_ffn_eval.cpp
@@ -3542,6 +3542,51 @@ bool eval_moe_hybrid_ffn_batched(
                             : storage.gate_cold    ? (int)storage.gate_cold->ne[2]
                             : 0;
     const bool cold_on_gpu = storage.cold_backend_kind == MoeHybridColdBackend::Gpu;
+    // Cold owner None (a cluster rank): every route that survived masking is
+    // resident here and there is no second owner, so the whole batch can be
+    // packed by expert into ONE graph per layer. Without this a reduced hot
+    // stack falls into the sub-batch loop far below, whose size is
+    // min(mmq_safe_sub_batch(), prefill limit) = 1 on gfx1151 - one graph per
+    // token per layer. Measured on a 1517-token prompt: 25.9 s of FFN against
+    // 7.6 s for a single node's whole prefill graph. Expert-major packing is
+    // also what keeps the reduced stack off the MMQ full-batch path that
+    // mmq_safe_full_batch=false exists to avoid.
+    const bool hot_only_expert_major =
+        !expert_compute &&
+        storage.cold_backend_kind == MoeHybridColdBackend::None &&
+        !storage.gate_cold && !storage.gate_up_cold && !storage.down_cold &&
+        n_hot_stack > 0 &&
+        moe_expert_major_prefill_enabled(n_tokens);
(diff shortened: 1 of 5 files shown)


davide221:
Restacked on the updated #701 (merged main, with the review fixes there). Your loader change is unchanged, and `None` storage now always records `materialized_cold_experts=false`, which is what the widened guard expects. Builds and tests pass on lucebox4.

davide221:
Restacked on #701 d4052bbf5. On the cubic threads here: the `n_tokens` ordering is fixed in #701. The 'unreachable' and 'no call site' findings are expected until the code that runs the two boxes (rank bootstrap, masking, shared-expert add after the reduction) is upstreamed.

PR #705, opened 2026-09-04
perf(dspark): keep the Markov chain graph instead of rebuilding it per call

maikzz32:
`dspark_markov_correct_greedy_chain_fused` builds its graph and frees it again on every call. The graph is a pure function of the candidate count, whether the confidence head is wanted, and the weights it closes over — all fixed for the life of a loaded model. The arena and the allocator beside it were already cached (`static thread_local`); only the graph was not.

This keeps the last one.

## Correctness

The key compares the weights by identity (`markov_w1`, `markov_w2`, `confidence_w`, `lm_head`, backend), so a second model in the same process rebuilds rather than silently reusing the first one's graph. Both failure paths — allocator and compute — invalidate, so a retry rebuilds. `DFLASH_DSPARK_NO_CHAIN_GRAPH_CACHE=1` restores the previous behaviour for isolating a suspected stale graph.

The context is freed before `build_markov_chain_graph` may resize the arena it builds into, which is why `invalidate()` drops the graph but keeps the arena and the allocator.

## Measurement

Radeon 8060S (gfx1151), ROCm 10, DeepSeek V4 Flash, DSpark q=4, 128 tokens at temperature 0, median of 5. Output byte-identical either way.

| | head phase |
|---|---|
| off | 4.5 ms/step |
| on | 4.3 ms/step |

Modest, and I want to be straight about why I expected more. A candidate-count sweep (head = 2.2 ms at 1 candidate, 4.4 at 3) fits to 1.1 ms fixed plus 1.1 ms per candidate, which looked like graph construction. It is not: most of the constant part is the `lm_head` matmul, which reads all 129280 × 4096 of its weights whatever the candidate count. This commit removes the construction, which is the part that was genuinely wasted.

## Verification

Built against `main` with `-DDFLASH27B_GPU_BACKEND=hip -DDFLASH27B_HIP_ARCHITECTURES=gfx1151 -DDFLASH27B_ROCMFP2_AFFINE=ON -DGGML_HIP_GRAPHS=ON` on a gfx1151 host: 379/379, no new warnings. `dspark_markov_project_topk` shares the builder and is untouched — it keeps its own arena and allocator.

<a href="https://cubic.dev/pr/Luce-Org/lucebox/pull/705?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/src/common/dspark_head.cpp
+++ b/server/src/common/dspark_head.cpp
@@ -186,6 +186,56 @@ struct MarkovChainGraph {
     std::vector<ggml_tensor *> confidence;        // optional sigmoid score per depth
 };
 
+// The chain graph is a pure function of its shape: the candidate count,
+// whether the confidence head is wanted, and the weights it closes over --
+// all fixed for the life of a loaded model. It was nevertheless rebuilt on
+// every call and freed again, which on DeepSeek V4 cost 1.1 ms of a 4.5 ms
+// head phase constructing tensors rather than computing anything. Keep the
+// last graph and reuse it; the shape key rebuilds when anything it depends
+// on changes, and the weights are compared by identity so a second model in
+// the same process cannot silently reuse the first one's graph.
+struct MarkovChainGraphCache {
+    MarkovChainGraph g;
+    std::vector<uint8_t> arena;
+    ggml_gallocr_t galloc = nullptr;
+    bool           built = false;
+    const void *   lm_head = nullptr;
+    const void *   markov_w1 = nullptr;
+    const void *   markov_w2 = nullptr;
+    const void *   confidence_w = nullptr;
+    ggml_backend_t backend = nullptr;
+    int            n_cand = -1;
+    bool           want_confidence = false;
+
+    bool matches(const DraftWeights & dw, ggml_tensor * head, ggml_backend_t be,
+                 int cand, bool conf) const {
+        return built && lm_head == head && backend == be && n_cand == cand &&
+               want_confidence == conf &&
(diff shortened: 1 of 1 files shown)


Graffioh:
maybe you can also double-check and in case address:

----

The cached graph contains **pointers to the model’s weights**. Those weights must remain alive whenever the graph is used.

Currently, the cache belongs to the thread, which can outlive the loaded model:

```text
Load model A → cache graph pointing to A’s weights
Unload A     → free A’s weights, but keep cached graph
Load model B → cached graph still refers to A’s old memory
```

The PR checks pointer addresses to detect a different model, but the memory allocator can reuse the same addresses, making that check unreliable.

**The lifetime fix means making the model own the cache:** destroy the cached graph before freeing that model’s weights and backend. A newly loaded model then starts with a fresh cache.

It still reuses the graph throughout generation; it just stops keeping it after the resources it depends on are gone.

----

i think it would also simplify a bit the cache handling if that's actually the case

davide221:
Thanks, this already landed: main absorbed the chain-graph cache in eee1f6c0f, keyed on the confidence and logit-margin outputs as well, plus a drafter lifecycle generation (`dspark_note_drafter_lifecycle`). That generation fixes the reload-lifetime issue @Graffioh raised. I think this PR can be closed as superseded.

PR #718, opened 2026-09-08
feat(server): support batched qwen with single-request deepseek4

Graffioh:
Allow a batched Qwen worker and a normal, single-request DeepSeek4 worker behind the same listener. For example, Qwen can generate four candidate answers on R9700 while DS4 uses its existing fused-decode/DSpark path on Strix Halo to synthesize them. The client orchestrates those stages with explicit model names.

The router derives capacity from the sequence engine when present and reserves one request for the normal worker otherwise. Existing worker selection, request parsing, model-specific defaults, and output formatting are reused. Shutdown now propagates cancellation to normal generation; startup and JSON status expose each worker's execution mode. Includes the Qwen C4 / DS4 C1 launch recipe and shared-environment limitations.

Depends on #715. That PR remains unchanged and covers routing between batched models. This draft adds the hybrid behavior in commit `2e2aeef5`; the base remains `main` until the prerequisite is merged.

Validation: built for both HIP gfx1151 and gfx1201; 459 host tests and six CLI checks passed. New integration coverage exercises mixed-worker admission, independent progress, disconnect/capacity reuse, and active shutdown.

GPU qualification is pending. After stopping another benchmark, a guarded 0731 load was stopped before readiness under host memory pressure (about 4.8 GiB minimum available). With no GPU processes remaining, the machine reported about 97 GiB available, below the target and draft's combined weight size. This does not establish whether the hybrid fits on a clean system. No hybrid throughput, model-residency, DSpark-activation, or output-equivalence claim is made. The downloaded target's checksum matches the publisher; the real 4+1 and synthesis checks remain to be run before this draft is ready.

<a href="https://cubic.dev/pr/Luce-Org/lucebox/pull/718?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Graffioh:
Closing this runtime draft as research-only work, as requested. Draft research PR https://github.com/Luce-Org/luce_box/pull/103 records the completed 4+1 GPU qualification, standalone DS4 output comparisons, and negative three-task synthesis result, alongside code structure and failed attempts. The implementation branch and local hybrid-serving worktree remain intact.

PR #730, opened 2026-09-14
feat(server): carry drafter feature ring in concurrent prefix checkpoints

Graffioh:
Speculative continuous batching rebuilds the DFlash draft K/V window from each sequence slot's target feature ring. Restoring only attention K/V and recurrent state leaves that ring stale after slot reuse, reducing draft acceptance.

This change adds an optional feature-ring slab to paged prefix snapshots and includes it in the resident-byte estimate. `Qwen35SeqEngine` owns feature eligibility and resets draft state on admission, stale-restore fallback, recompute, and retirement. Featureless restores mask out unavailable history until the current sequence rebuilds it; captures can include features again after the entire ring has been overwritten. The scheduler and model-neutral `PrefixStore` contract remain unchanged.

Snapshot validation checks the feature tensor's dimensions, layout, capacity, and width before transferring data. `DFLASH_PREFIX_NO_FEAT=1` remains a documented debug control for exercising featureless restores; `=0` leaves feature payloads enabled.

Validation on the rebased branch:

- HIP build: `dflash_server` and all four targeted test binaries succeeded.
- `test_recurrent_snapshot`: 6/6 passed, including short-prefix copying, byte accounting, slot/dead-row isolation, malformed feature dimensions, and atomic replacement failure.
- `test_seq_engine_contract`: 33 checks; `test_seq_slot_manager`: 306 checks; `test_parallel_prefix_txn`: 29 checks.
- Qwen3.8-27B + GGUF DFlash2 smoke test on gfx1151, with `--max-concurrency 2`: cold request, unrelated request reusing the slot, then prefix restore. Repeated with `DFLASH_PREFIX_NO_FEAT=0` and `=1`; logs confirmed a 982-token restore with `feat=1` and `feat=0`, respectively. Greedy responses matched the cold request and each other.

Rebased onto `origin/main` at `f6c54171`. The new recompute path uses the same draft-state reset as admission and retirement. No new performance benchmark is claimed.


Graffioh:
Addressed the featureless-restore finding (squashed into `1fd73fa0`).

**Reachability**: today a spec engine can never *produce* a featureless checkpoint — `fixed_chain_ready_` requires `cache.target_feat` non-null, so every `capture_prefix` under speculation carries the slab. The path only exists for restore-side compatibility, but the defensive fix is now explicit rather than implied.

**Fix** (instead of full AR fallback): a per-slot `slot_ring_valid_from_` floor records the restore cut when a featureless checkpoint lands. `prepare_chain_drafts` seeds `next_pos` past it, so `draft_kv_begin_step` bulk-appends only rows the occupying sequence wrote itself — foreign rows keep `slot_pos = -1` and stay masked out of the draft context. Speculation keeps running on an honest (if short) context rather than burning draft compute on poisoned proposals, which is strictly better than AR fallback for prompts where recent context suffices.

Two extra pieces:
- `capture_prefix` on such a slot now omits the feature slab too — foreign rows must not be blessed into a `feat=1` checkpoint that would propagate staleness to downstream restores.
- `DFLASH_PREFIX_NO_FEAT=1` is kept as a permanent bench/debug knob (drops the payload from captures). It exercises this cold-ring path and measures the slab's contribution to acceptance.

E2E verified on Qwen3.8-27B + DFlash2: with the flag set, restores log `feat=0`, output stays byte-identical to the feat run, and decode runs on the self-written suffix (~17 tok/s on the prefix-dependent echo workload vs 61 tok/s warm — expected since the needed context is deliberately in the untrusted region).

Graffioh:
Addressed the capture-suppression expiry in `7648998e`.

Valid point — the floor is now self-expiring rather than lifetime. Ring rows are keyed by `position % target_feat_cap`, so the last foreign row (`floor - 1`) is overwritten once the sequence writes `cap` positions past the restore cut. `capture_prefix` now carries the slab when `checkpoint.tokens >= ring_floor + target_feat_cap`.

The floor itself stays in place after expiry — `next_pos` has already advanced past it, so the `prepare_chain_drafts` clamp is a no-op, and the record remains accurate as "this slot once restored without features".

Verified: rebuild clean, all tests pass, normal path e2e unchanged (`feat=1` captures at 218.7MB, byte-identical output at ~61 tok/s). The expiry boundary itself can't be reached e2e without a 4k+ token generation on a floored slot — and floored slots only arise under `DFLASH_PREFIX_NO_FEAT`, which suppresses captures unconditionally — so it's verified by inspection against the ring indexing (`position % cap` overwrite semantics).

Graffioh:
@cubic-dev-ai review this PR

